PR #63, opened 2020-04-22
Hanabi Integration

dissendahl:
- Hanabi, integrated from official repo as git submodule. 
- See hanabi/README.md for how to pull and setup git submodule.
- See documentation within class for full information.
- Debugged and E2E tested with unittest, including api test as part of the unit tests.

- CHANGE REQUEST: Add typing annotation in utils/env.py, see last commit.


dissendahl:
@weepingwillowben and @justinkterry : Thanks for reviewing! 

1. I will update the env to work with a default game configuration. @weepingwillowben: So I understoud now that envs should work without config. But is it okay to pass them parameters or is this not wanted all?
2. I will remove the git submodule thing and the belonging readme. 
3. Take a look at the merge conflict.

weepingwillowben:
The goal is something like this:

```
class env(AECEnv):

    metadata = {'render.modes': ['human']}

    def __init__(self, seed=None, spawn_rate=20, num_archers=2, num_knights=2, killable_knights=True, killable_archers=True, pad_observation=True, max_frames=900):
```

weepingwillowben:
So the environment will have parameters, but they will all have a default value.

justinkterry:
@clemens4321 Please move the type hints from the AECEnv class specifically to hanabi

dissendahl:
Removed the branch, therefore have to reopen pull request.

PR #77, opened 2020-05-08
Removed obsolete error handling code

lssr:


weepingwillowben:
This all looks good, but the idea is to remove even more. In particular, 

* the code that handles illegal moves should now be in the wrapper, so the illegal move if statement can be simplified to an `assert` that the move is legal. 
* the code which handles if an agent is already done, but is stepped anyways should be handled by the wrapper.

PR #83, opened 2020-05-12
Removed section that are handle by wrappers

lssr:
- Removed illegal move section. Already part of wrappers.
- Removed section that handled agent that is already done but is stepped anyways.


justinkterry:
@weepingwillowben we happy?

weepingwillowben:
Looks good

PR #97, opened 2020-05-31
fixed crashes on headless server

weepingwillowben:
pistonball, KAZ and pursuit were crashing on my headless server due to pygame errors. 

This PR fixes it.

### DIFF
--- a/pettingzoo/atari/base_atari_env.py
+++ b/pettingzoo/atari/base_atari_env.py
@@ -152,5 +152,6 @@ class BaseAtariEnv(AECEnv):
         return image
 
     def close(self):
-        pygame.display.quit()
-        self._screen = None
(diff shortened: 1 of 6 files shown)


MarioJayakumar:
Hey could you clarify what crash you were getting? The travis build is also a headless server but the only crash that was happening was that the bombardment test would take too long.

weepingwillowben:
Yes it was a:

pygame.error: No available video device

error

weepingwillowben:
Sorry, that was the wrong error. here is the error I am getting: 
pygame.error: video system not initialized

MarioJayakumar:
I'll check this out locally, this weird because i only ever got the "video system not initialized" error when using pygame.pump on a closed environment or trying to render a closed environment

PR #98, opened 2020-05-31
added agent order fix to wrapper

weepingwillowben:
Made agent_order empty when all agents are done


weepingwillowben:
So it turns out this doesn't work don't merge this yet.

PR #100, opened 2020-06-01
added obs dtype test

weepingwillowben:
The observation for a Box observation space should by a numpy array of the same shape and dtype as the Box.

Up to now, we have only checked for the shape. This PR adds a test for the dtype as well and a large number of environments fails this test. 

This PR is not yet complete, should be finished when all environments pass the tests.


weepingwillowben:
Ok, this should be ready to merge now, assuming I didn't screw things up. All envs are passing the tests.

PR #101, opened 2020-06-02
added vector env wrapper

weepingwillowben:
This adds a vector environment. The API is super weird, so before merging, 

vector_env.py

should be reviewed for sanity and for bugs.

Also it is not 100% clear that pettingzoo is the appropriate place for this wrapper.


weepingwillowben:
I think this code might actually belongs in the multi-agent stable baselines code. Thoughts?

weepingwillowben:
Just moved this over to supersuit

PR #105, opened 2020-06-07
added agent iter to env and readme

weepingwillowben:
This is the proposed replacement for agent_order.


weepingwillowben:
Here is the code I am testing it with
```python
from pettingzoo.classic import mahjong_v0
import time
import random
#from supersuit import frame_stack
import numpy as np


env = mahjong_v0.env()
orig_obs = env.reset()
for x in range(20000):
    for agent in env.agent_iter:
        assert agent == env.agent_selection
        reward, done, info = env.last()
        action = random.choice(env.infos[agent]['legal_moves'])#.sample()
        next_obs = env.step(action)

    print("env reset")
    print(env.rewards)
    orig_obs = env.reset()
```

weepingwillowben:
Added error messages for correctly wrapped environments:

```
env = mahjong_v0.env()
env.reset()
# this crashes with the following error message: AssertionError: need to call step() or reset() in a loop over `agent_iter`!
list(env.agent_iter)

env = mahjong_v0.raw_env()
env.reset()
# this gives an infinite loop
list(env.agent_iter)
```

weepingwillowben:
Added an error message for trying to use `agent_order`.

```
AttributeError: agent_order has been removed from the API. Please consider using agent_iter instead.
```

weepingwillowben:
Changed agent_iter interface to a function call `agent_iter()` with an optional argument to specify the maximum number of steps.

PR #111, opened 2020-06-25
Atari warns

weepingwillowben:


weepingwillowben:
Feels a little weird to have warnings for default behavior, but here it is.

PR #114, opened 2020-06-30
fixed bad info in prison and added check to tests

weepingwillowben:


weepingwillowben:
A small API change. Infos for each agent are required to be dicts

PR #117, opened 2020-07-05
Remove auto imports

weepingwillowben:
Removed all the imports in the `__init__.py` files. Replaced them with actual files. This was done by renaming files in atari, magent, and mpe, and creating new files which imported from the old files in gamma, sisl, and classic


justinkterry:
I thought we were renaming gamma to butterfly with this PR

weepingwillowben:
I was going to do that in another PR.

PR #118, opened 2020-07-06
renamed gamma to butterfly

weepingwillowben:
Not sure if this will pass tests. Also did a quick find and replace, so there may be issues here.


rallen10:
This breaks the PettingZoo example given in ray/rllib. See:

- https://docs.ray.io/en/master/rllib-env.html#pettingzoo-multi-agent-environments
- https://github.com/ray-project/ray/blob/master/rllib/examples/pettingzoo_env.py

Someone should submit a pull request to ray to update `from pettingzoo.gamma import prison_v0` to `from pettingzoo.butterfly import prison_v0`. Sorry for not doing it myself; I'm just getting started with PettingZoo, not sure that I want to go through the effort of a pull request on a separate library to improve examples and docs.
I could post this as in issue in the ray repo, but it actually seems more likely to be addressed if posted here.

Cheers!

PR #133, opened 2020-07-15
fixed waterworld bug where only pursuer_4 gets any reward

weepingwillowben:
I think this fixes the waterworld bug.


weepingwillowben:
Yep, seems to work now

weepingwillowben:
Make sure I'm not missing something

PR #135, opened 2020-07-17
fixed pursuit observation issues

weepingwillowben:
Some major overhauls for pursuit

* Reshaped observation space to a (7,7,3) observation space. (first is walls, second is allies, third is opponents)
* Removed layer norm. now all values in observation are integers (the type is still float) 
* Observation space is now bounded by the max number of agents of one type (default [0,30])
* Appropriate document fixes


weepingwillowben:
Note that this is passing tests.

PR #136, opened 2020-07-17
fixed hanabi rewards

weepingwillowben:
Made hanabi rewards accumulate properly and also fixed the documentation to reflect how the reward is actually allocated.


weepingwillowben:
Note that this is also passing tests

PR #143, opened 2020-07-20
improved pursuit performance

weepingwillowben:
Looked through sisl for low hanging performance fruit, found one small thing in pursuit, no where else unfortunately. All I discovered was that python is stupid slow.


weepingwillowben:
Just ran this through average reward test and it doesn't affect average reward for pursuit.

PR #149, opened 2020-07-25
moved title above infobox

weepingwillowben:
Moved title above info box so that it looks OK on narrower screens. May not be optimal on larger screens, but it isn't completely awful like small screens are currently. 

title_above

title_above2


justinkterry:
I don't understand what problem this fixes on small screens? Also I was talking about how thin the body column is when the website is half screened.

Screen Shot 2020-07-25 at 11 57 20 PM

weepingwillowben:
closing this due to conflicts with refactoring on master

Issue #158, opened 2020-08-10
Simultaneous actions


RedTachyon:
I'm not sure if I'm missing something, but the way the multiagent property is used seems to be by an agent iterator, as in during each step you iterate over all agents and choose their actions.

Is it possible to instead pass the actions for all agents in a given timestep at the same time? I feel like this could be much better for homogeneous agent cases, where it'd be beneficial to just compute the actions for all agents at once.

I'm thinking something like they did in rllib, where you pass the actions as a dictionary {"Agent0": 1, "Agent1": 3, ...}

weepingwillowben:
There has been considerable internal discussion on this for a long time, so I am glad for your input.

As we continue to discuss a permanent solution, a temporary solution is to simply use the rllib wrapper at https://github.com/ray-project/ray/blob/master/rllib/env/pettingzoo_env.py that is in the nightly build and will be part of the ray release 0.9.0, but you should be able to copy the file, change the first line to `from ray.rllib.env.multi_agent_env import MultiAgentEnv` and use it internally in your code. Note that this wrapper assumes that the environment is in fact a parallel environment and will soft-fail for classic games like chess or even hanabi, but should work fine for all other games.

RedTachyon:
Wouldn't it be viable to have a mixed/unified approach? I'm thinking about something like this: the environment generally accepts dictionary-based inputs, but specific environments where the AEC formulation is necessary can just ignore actions that are irrelevant at any given time step.

By the way, do you have any reference for the AEC game model? I haven't been able to find any specifics on that.

justinkterry:
The plan is to have a secondary API for the environments that support simultaneous actions.

The AEC game model is from a paper under review at NeurIPS but it looks like it won't get in. Right now we're in the middle of a rewrite and porting it to AAMAS, it should be on on arxiv within the month.  PettingZoo blew up much faster than we were expecting and the papers going much more slowly than we hoped.

The sections of it relevant to you are 1-3, and are in decent form (though I'd definitely appreciate feedback): https://www.overleaf.com/read/xjnyvvppdrxh. The sections after are screwed up at the moment.

justinkterry:
@RedTachyon A legitimate version of the paper on AEC games is now out: https://arxiv.org/abs/2009.13051

PR #164, opened 2020-08-12
Parallel env

weepingwillowben:
Addresses #158 

Adds a module.parallel_env() method to every environment where there are simultaneous actions. 

Needs documentation before merge


weepingwillowben:
Also eventually needs supersuit wrapper support

Issue #170, opened 2020-08-16
Better performance test


ndyashas:
Hi!. I've noticed that the current performance testing at ```pettingzoo/tests/performance_benchmark.py``` runs for only 5 seconds and is run only once. Wouldn't it be better if the test is run multiple times with a longer runtime? That way we can report the results as ```value ± error```.

I'm happy to make a PR if this is valid.

weepingwillowben:
Thanks for bringing this up. After you mentioned it, I found a bug in the performance test which is fixed by #171

weepingwillowben:
But in general, while 5 seconds doesn't seem like that long of a time, for classic environments, this is long enough to play dozens of games, and for most other environments, the complexity of the calculations seems to remain fairly constant throughout play, so the number of samples doesn't matter that much. 

We also want to keep the travis tests running quickly. So any change should not affect the ci_test's runtime.

ndyashas:
Understood!. Also, I think it would be better to use the ```time.process_time()``` rather than ```time.time()```. As mentioned [here](https://docs.python.org/3/library/time.html#time.process_time), ```time.process_time()``` does not include time elapsed during sleep. Thus, a (very slightly) better measure compared ```time.ime()```.

I will make some local changes and let you know if I get consistent reports on the ```value ± error``` reporting method I mentioned earlier. If they have similar inconsistencies over multiple runs, then it would be better to leave out that complexity.

weepingwillowben:
I merged the fix, make sure to use it in any local tests. I will await the results with interest.

ndyashas:
I have made the modifications while keeping the overall test time same, However, while testing the inconsistency seems to still be as they were in the previous version. I have also added some benchmarking reliability tests as given [here](https://pyperf.readthedocs.io/en/latest/cli.html#pyperf-check).

I have also used ```time.process_time()``` instead of ```time.time()```. PR #172

PR #172, opened 2020-08-16
Updated the pettingzoo.tests.performance_benchmark function to display output as mean ± stdev

ndyashas:
to make multiple smaller runs and report results in the
mean ± stdev format.

Commit related to #170.

### DIFF
--- a/pettingzoo/tests/performance_benchmark.py
+++ b/pettingzoo/tests/performance_benchmark.py
@@ -37,7 +37,7 @@ def _performance_benchmark(env, test_sample_time):
     return cycles_per_time, turns_per_time
 
 
-def check_stability(name, arr):
+def _check_benchmark_stability(name, arr):
     # Stability check based on the pyperf standard at
     # https://pyperf.readthedocs.io/en/latest/cli.html#pyperf-check
     mean = arr.mean()
@@ -56,12 +56,8 @@ def check_stability(name, arr):
         print(f"[WARNING] for {name} maximum is 50% greater than the mean", file=sys.stderr)
 
 
-def performance_benchmark(env):
+def performance_benchmark(env, test_time=5, test_sample_time=1, check_benchmark_stability=True):
     print("Starting performance benchmark")
-
-    test_time = 5
-    test_sample_time = 1
-    check_benchmark_stability = True
 
     assert test_time % test_sample_time == 0, "'test_time' should be a multiple of 'test_sample_time'"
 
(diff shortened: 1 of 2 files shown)


pettingzoo/tests/performance_benchmark.py
```diff
@@ -24,14 +25,59 @@ def performance_benchmark(env):
...
+        # Warn if the maximum is 50% greater than the mean
+        print(f"[WARNING] for {name} maximum is 50% greater than the mean", file=sys.stderr)
+
+
+def performance_benchmark(env):
+    print("Starting performance benchmark")
+
+    test_time = 5
+    test_sample_time = 1
+    check_benchmark_stability = True
```
ndyashas, line 64: I have left these parameters internally and not added them as arguments list so as to not break the current API. However, if added to the argument list, they still will be backward compatible.

weepingwillowben: It should be fine to add these to the arguments list

justinkterry: make sure you update the documentation accordingly

justinkterry:
I don't understand why this is functionality that we need or would want to add

ndyashas:
With the existing implementation, multiple runs report different benchmarking values without an estimation of how much error is present in them. The primary idea was to report the result along with an estimated error which would give a better idea than just the absolute value.

This modification does not significantly improve functionality. I made these with the following in mind.

- The existing implementation uses ```time.time()``` which is similar to wall-clock time and varies depending on other processes running on the system at the time, instead ```time.process_time()``` ignores the time for which the current process was put to sleep by the OS, thus a better measure compared to ```time.time()```.
- Splitting the runs allows us to estimate if the benchmarking was more-or-less consistent as it is done in the ```pyperf``` tool [here](https://pyperf.readthedocs.io/en/latest/cli.html#pyperf-check).
- Splitting the runs also lets us report the result along with estimated error. eg: ```mean ± stdev```.

weepingwillowben:
So I looked at the standard deviations reported by travis's run, and it was incredibly small. Less than 10% in all cases, if your code was right. So I guess I was right about it being small. 

We feel that users are more concerned about factors of 2 or more, as the systems which use these environments never run in real time and are never very sensitive to execution time. 

Also, we feel that time.time() is the better and more useful measure because it can handle all weirdness that may go on behind the scenes of an environment such as:

1. Environments spawning off sub-processes which do the real work (none of the current pettingzoo environments do this, but the starcraft micromanagement environments do).
1. Multiple threads (magent used to do that, I disabled it)
1. Any other weirdness that will put the process clock out of alignment with what people actually care about: wall time. 


And when you are talking about multiple seconds, the amount of sleeping that happens should be relatively small unless the system is completely overloaded.  So process time solves one problem but introduces potentially larger problems.

ndyashas:
Ah, yes. The points make sense. Thank you very much for your time and the detailed explanation!. I will close this PR and the related Issue for now, as it makes more sense to continue with the current functionality as you mentioned.

PR #174, opened 2020-08-17
Update sass to resize circle buttons and image on 4K screens

RyanNavillus:
Made improvements to reduce the size of the header on 4K screens, and fix responsive scaling.


docs/_sass/layout/_masthead.scss
```diff
@@ -51,12 +51,12 @@ header.masthead {
...
     text-align: center;
     img {
-      height: 150px;
+      height: 100%;
     }
   }
   .btn-circle {
     border-radius: 350px;
-    padding: 10%;
+    padding: 20px;
```
weepingwillowben, line 59: Did you check that these looked OK at smaller screen sizes? Right now the site looks OK until 680px; smaller than that and it looks bad. I guess I am wondering if it still look OK at 680px in width or so.

RyanNavillus: I didn't check that, I'll look at it when I work on this later tonight. Is this supposed to look good at mobile resolutions too?

justinkterry: It's supposed to be vaguely useable on mobile

PR #201, opened 2020-09-18
fix simple_speaker_listener agents' name

51616:
The names are swapped. This is a quick fix.


weepingwillowben:
Great catch. Looks like documentation already lists the speaker first, so there is no need to update that.

weepingwillowben:
I'll fix the gif on the website.

PR #208, opened 2020-09-24
Checkers

RyanNavillus:
Changed action space and observation space, and passed all test cases.


weepingwillowben:
I hooked up the checkers environment to the main CI tests and it isn't passing. Can you get the tests to pass?

docs/classic/checkers.md
```diff
@@ -0,0 +1,60 @@
...
+
+{% include info_box.md %}
+
+
+
+Checkers (also claled Draughts) is a 2-player turn based game. Our implementation is based on the OpenAI gym checkers implementation, with changes to the observation and action spaces.
+
+#### Observation Space
+
+There are 32 habitable spaces in the game of checkers, numbered from left to right and top to bottom, such that every other square in a row is numbered. On even numbered rows (starting with 0) the second square is the first habitable position. On odd numbered rows, the first square is habitable.
```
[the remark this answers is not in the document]
RyanNavillus: Yeah, that term was used either in the gym docs or the Wikipedia page for checkers so I went with that. We could change it, occupiable might be better

docs/classic/checkers.md
```diff
@@ -0,0 +1,60 @@
...
+
+---
+
+{% include info_box.md %}
+
+
+
+Checkers (also claled Draughts) is a 2-player turn based game. Our implementation is based on the OpenAI gym checkers implementation, with changes to the observation and action spaces.
+
+#### Observation Space
```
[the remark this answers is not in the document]
RyanNavillus: It says that in the table at the top

docs/classic/checkers.md
```diff
@@ -0,0 +1,60 @@
...
+The action space is also divided into 4 planes, one for each direction that a piece can move. Each plane has 32 values, one for each habitable square on the board. The action directions listed below:
+
+| Action | Direction |
+| ------ | --------- |
+| 0      | Northwest |
+| 1      | Northeast |
+| 2      | Southwest |
+| 3      | Southeast |
+
+The environment automatically decides whether the provided action is a simple move or a jump. Note that jumps are required when available. This is reflected in `infos[agent]['legal moves']` by listing jump moves and no simple moves if jumps are available.
```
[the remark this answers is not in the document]
RyanNavillus: You pick a direction, if there is an enemy piece in that spot, it's a jump. if it's empty, it's a simple move to that spot. anything else is an illegal move and will be rejected. If you just mean its confusing to read, I agree the wording kind of sucks

RyanNavillus:
The tests passed before, it looks like the checkers branch wasn't updated for a while so the tests were really different from the master branch. Now that I merged, there are some issues

RyanNavillus:
They pass the tests with the last 2 wrappers now. I didn't include that action out of bounds or illegal move wrappers. I'm kind of confused by the wrappers, are they expected to be on every env? The env already handles illegal actions internally.

weepingwillowben:
The illegal moves wrapper is so that there is a reasonable amount of consistency between environments.

The actions out of bounds wrapper is for nan values, which typically crash environments, but show up in rllib all the time.

docs/classic/checkers.md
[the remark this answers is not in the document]
justinkterry: Switch it

docs/classic/checkers.md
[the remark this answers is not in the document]
justinkterry: Fix it to say what you said there

RyanNavillus:
It looks like it's passing all the tests now. I updated the docs slightly to reflect the changes I made. I accidentally had the action space as a matrix rather than a discrete value, but that's fixed now.

weepingwillowben:
Some things that need to be fixed:

* in step() the self.rewards dictionary is not updated by the agent name, but the agent index. This effectively makes the rewards always zero.
* The observation space should be a 3d array, Height x Width x 4. The action space should similarly be documented as Height x Width x 4, even though it will just be a 128 discrete value.
* running `bash check_style.sh` should not print anything
* last() should be removed, as it is handled by the wrappers. "Why do we have mandatory wrappers" you might ask. That is a very good question that we should talk about when we go over the pettingzoo API changes in a couple weeks.

Otherwise, the code looks fine.

RyanNavillus:
The action space can't be height x width x 4 and also 128. The action space is the way that it is because only half of the board is occupiable. Increasing the size of the action or observation space would just be inefficient since half the states would be unreachable

weepingwillowben:
Sorry, I guess I mean you can choose a dimension to compress, either width or height, and then have a 4x8x4 or a 8x4x4 observation and action space. 
Even though this space is different than the board we see, a 4x8x4  or a 8x4x4  observation space will still preserve locality, so it is still suitable for processing with a CNN.

weepingwillowben:
Actually, CNNs may have a bit of an issue processing this upon further reflection

RyanNavillus:
I don't know if that's still suitable for CNN's. If you compress rows to 4 spaces so that the square is 4x8, the square (1,0) below (0,0) is southwest of (0,0 on a real board, but the square (2,0) below (1,0) is southeast of (1,0) on a real board. I might be wrong, but I think that would make training more difficult with a CNN. We either need to use an inefficient action space, or let the user's transform the spaces for CNN usage on their own.

justinkterry:
If the proper full encoding is unclear, should we just go with the most efficient representation and discuss why we chose it, the issues with this and potential solutions in the documentation? That's what we've done before for things like this

weepingwillowben:
@RyanNavillus Yeah, that was the issue that crossed my mind. Thanks for articulating it.

On the question of efficiency, note that many observation and action spaces for alphazero environments are inefficient--the reason being that CNNs just do such an incredibly better job learning these environments that it is worth having to search over a larger space during tree search (in the context of alphazero). On the other hand, a 128 dimensional action space is actually getting quite small, so maybe it isn't necessary?

RyanNavillus:
I mean, if a feed forward network can perform well on MNIST at 28x28=784 inputs, doubling the action and observation space to 256 would definitely not be too large. I think either choice is valid, but if a majority of people doing research on these environments are using CNN's, maybe we should expand it to a proper spacial representation.

weepingwillowben:
Personally, I find this sort of fascinating. It may need more investigation, but it seems likely that this is a good example of where intuitions about deep learning don't transfer over well to RL. It turns out that something about the priors implicit CNNs are just too useful for learning on board games. 

You can see what network the tournament winning, heavily tuned Lc0 chess engine uses: https://blog.lczero.org/dev/backend/nn/
Or the network archtecture described in the alphago zero paper: https://www.nature.com/articles/nature24270.epdf?author_access_token=VJXbVjaSHxFoctQQ4p2k4tRgN0jAjWel9jnR3ZoTv0PVW4gB86EEpGqTRDtpIz-2rmo8-KG06gqVobU5NSCFeHILHcVFUeMsbvwS-lxjqQGg98faovwjxeTUgZAUMnRQ

weepingwillowben:
Of course, it is worth noting that both observation spaces are more than an order of magnitude larger than checkers.

weepingwillowben:
But the interesting thing is that there is hardly any fully connected layers at all, they are just big, weird, resnets.

weepingwillowben:
And the filter sizes are huge for chess. 8x8, covering the entire board. So there is no saving of parameters.

RyanNavillus:
I updated the action space, removed last(), fixed the rewards (and tested them), fixed the styling, made sure the tests still pass, and updated the docs. I think this is ready to merge.

pettingzoo/classic/checkers/checkers.py
```diff
@@ -0,0 +1,608 @@
...
+        # Use self.ch.flatboard to update self.observation
+        board = self.ch.flat_board()
+        obs = np.zeros((8, 8, 4))
+        for i, row in enumerate(board):
+            for j, sq in enumerate(row):
+                if sq > 0:
+                    obs[i, j, sq - 1] = 1
+        return np.array(obs)
+
+    def observe(self, agent):
```
weepingwillowben, line 58: I recommend not precomputing the observation, but computing it when this function is called. Leave it up to the user whether or not to compute the observation by calling this function.

pettingzoo/classic/checkers/checkers.py
```diff
@@ -0,0 +1,608 @@
...
+            # self.ch.print_board()
+            # print(self.agent_selection)
+            # print(self.observe(self.agent_selection))
+            # print(self.observation)
+            # self.infos[self.agent_selection]['legal_moves']
+
+        self.infos[self.agent_selection]["legal_moves"] = self.legal_moves()
+
+        if winner is None and self.num_moves > self.num_moves_max:
+            print("Draw")
```
weepingwillowben, line 207: please remove the print statement. There is actually a test to catch print statement in the CI tests, but those aren't running for some reason.

pettingzoo/classic/checkers/checkers.py
```diff
@@ -0,0 +1,608 @@
...
+        self.dones[self.agent_order[1]] = winner is not None
+
+        if observe:
+            next_observation = self.observe(self.agent_selection)
+        else:
+            next_observation = None
+        return next_observation
+
+    def render(self, mode="human"):
+        print(self.ch.flat_board())
```
weepingwillowben, line 233: Perhaps a clearer render would just to print it like this?

```
  C  _  _  _
_  K  _  _
  _  K  C  _
_  _  _  _
  _  _  _  _
_  _  c  _
  k  _  _  _
c  _  _  _
```
where c is a checker and k is a king, and capital letters represent white? This is modeled after how python-chess does it, but there may be some other reasonable way.

docs/classic/checkers.md
```diff
@@ -0,0 +1,58 @@
...
+layout: "docu"
+title: "Checkers"
+actions: "Discrete"
+agents: "2"
+manual-control: "No"
+action-shape: "Discrete(8*8*4)"
+action-values: "Discrete(8*8*4)"
+observation-shape: "Box(8, 8, 4)"
+observation-values: "[0, 1]"
+num-states: "10^21"
```
weepingwillowben, line 11: I thought we weren't documenting num_states.

docs/classic/checkers.md
```diff
@@ -0,0 +1,58 @@
...
+| 192...255 | 0...63          | Southeast |
+
+When an action is chosen, the environment automatically decides whether the provided action is a simple move or a jump. Given an action with a starting location and direction, if the square immediately adjacent to the starting location in that direction is unoccupied, then the move is a simple move to that square. If the square is occupied by an enemy man, and the next square in that direction is unoccupied, then the move is a jump. In any other situation, the move is illegal. Note that each player is required to make jumps when available. This is reflected in `infos[agent]['legal moves']` by only listing jump moves if at least one is available.
+
+#### Rewards
+
+| Winner | Loser |
+| :----: | :---: |
+|   +1   |  -1   |
+
```
weepingwillowben, line 54: What about draws?

pettingzoo/classic/checkers/checkers.py
```diff
@@ -0,0 +1,608 @@
...
+        self.last_turn = "black"
+        self.rewards = {name: 0 for name in self.agents}
+        self.dones = {name: False for name in self.agents}
+        self.infos = {name: {"legal_moves": []} for name in self.agents}
+        self.infos[self.agent_selection]["legal_moves"] = self.legal_moves()
+        self.winner = -1
+        if observe:
+            return np.array(self.observation)
+
+    def _obs_rel_to_abs(self, pos):
```
weepingwillowben, line 80: is this function used anywhere?

pettingzoo/classic/checkers/checkers.py
```diff
@@ -0,0 +1,608 @@
...
+        pos64 = 0
+        if row % 2 == 0:
+            pos64 = 2 * pos + 1
+        else:
+            pos64 = 2 * pos
+        row = int(pos64 / 8)
+        col = int(pos64 % 8)
+        return (row, col)
+
+    def _obs_abs_to_rel(self, pos):
```
weepingwillowben, line 92: is this function used anywhere?

pettingzoo/classic/checkers/checkers.py
```diff
@@ -0,0 +1,608 @@
...
+        self.action_spaces = {name: spaces.Discrete(64 * 4) for name in self.agents}
+        self.observation_spaces = {
+            name: spaces.Box(low=0, high=1, shape=(8, 8, 4), dtype=np.int32)
+            for name in self.agents
+        }
+        self.observation = np.zeros((8, 8, 4))
+
+        self.reset()
+
+    def _read_observation(self):
```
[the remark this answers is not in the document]
weepingwillowben: If you think there is some better way of organizing observations, we can discuss.

weepingwillowben:
Note that now that CI tests are running properly, it is failing those tests.

pettingzoo/classic/checkers/checkers.py
```diff
@@ -0,0 +1,608 @@
...
+    def observe(self, agent):
+        return np.array(self.observation)
+
+    def seed(self, seed=None):
+        pass
+
+    def reset(self, observe=True):
+        self.ch = CheckersRules()
+        self.observation = self._read_observation()
+        self.num_moves_max = 300
```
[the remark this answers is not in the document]
RyanNavillus: How does chess handle the game clock? I think a similar clock is used for checkers

pettingzoo/classic/checkers/checkers.py
```diff
@@ -0,0 +1,608 @@
...
+            )
+            winner = "white" if self.last_turn == "black" else "black"
+        else:
+            self.num_moves += 1
+            action = self._parse_action(action)
+            self.board, turn, last_moved_piece, moves, winner = self.ch.move(
+                action[0], action[1]
+            )
+
+            self.agent_selection = self._agent_selector.next()
```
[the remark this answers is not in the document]
RyanNavillus: I just used it because the existing code had it. It seemed pointlessly confusing, I'll replace it

pettingzoo/classic/checkers/checkers.py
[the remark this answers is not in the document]
RyanNavillus: That makes sense, I can do that.

pettingzoo/classic/checkers/checkers.py
RyanNavillus: Good point

pettingzoo/classic/checkers/checkers.py
RyanNavillus: Sure

pettingzoo/classic/checkers/checkers.py
RyanNavillus: I can change it to something like this

docs/classic/checkers.md
RyanNavillus: No clue, it was in some of the other classic documentation so I looked it up. I feel like if we have the information, it's a useful thing to know.

docs/classic/checkers.md
RyanNavillus: Both agents get 0 reward, that should be documented

pettingzoo/classic/checkers/checkers.py
RyanNavillus: Nope, I'll remove it

pettingzoo/classic/checkers/checkers.py
RyanNavillus: Also no, I guess I just wrote these out of convenience and ended up not needing them

pettingzoo/classic/checkers/checkers.py
RyanNavillus: Screenshot from 2020-10-07 02-22-28

weepingwillowben: So you don't have to print out the reward with the observation, as the user can easily print the reward themselves. 

Otherwise looks good.

RyanNavillus: That was just my testing script, it's not part of the render function

weepingwillowben:
Ok, this should be ready to merge

Issue #209, opened 2020-09-26
Error with Tiger-Deer Env: Selecting Invalid Agent.


jdpena:
Hi,

I've been running the MAgent Tiger-Deer environment with 2 different algorithms: a [RandomLearner](https://chuacheowhuan.github.io/RLlib_rand_policy/) and `rllib`'s PPO. I'm also currently using `rllib`'s [PettingZooEnv](https://github.com/ray-project/ray/blob/master/rllib/env/pettingzoo_env.py). It seems both of the algorithms work for some number of iterations, but then error-out in this line https://github.com/ray-project/ray/blob/master/rllib/env/pettingzoo_env.py#L161. 

The issue is that the agent being selected, `deer_92`, is not in the `action_dict`. I checked the `self.aec_env.dones` dict, however, and the agent is there. I added a snippet of the output, below. I printed out relevant info (shown after each `== start step ==`) when entering the [step()](https://github.com/ray-project/ray/blob/master/rllib/env/pettingzoo_env.py#L141) function.  Furthermore, it also appears all steps prior to this error only select `deer_0` as the agent. I've re-ran the experiment several times and it always has the same result (e.g., `deer_0` is always chosen and then it errors-out once any other agent is chosen).

I'm not sure if this is an issue with rllib, the Tiger-Deer env, or my own config.

```
[2m[36m(pid=34152)[0m =============== start step =====================
[2m[36m(pid=34152)[0m self.aec_env.agent_selection --> deer_0
[2m[36m(pid=34152)[0m stepped_agents --> set()
[...]

PR #214, opened 2020-09-30
fixed agent iter wrapper

weepingwillowben:
Addresses #209


weepingwillowben:
Unfortunately, while this fixes the issue, I feel as if there is some complexity here I am missing at the moment.

Issue #215, opened 2020-09-30
Prospector error


weepingwillowben:
I am getting this error in prospector:

```
   return self.env.observe(agent)                                                                                        
File "/home/ben/.virtualenvs/main_env/lib/python3.6/site-packages/pettingzoo/butterfly/prospector/prospector.py", line 739, in observe                                                                                                            
sub_screen, pad_width=((pad_x, 0), (0, 0), (0, 0)), mode="constant"                                                   
File "<__array_function__ internals>", line 6, in pad                                                                   
File "/home/ben/.virtualenvs/main_env/lib/python3.6/site-packages/numpy/lib/arraypad.py", line 746, in pad                
pad_width = _as_pairs(pad_width, array.ndim, as_index=True)                                                           
File "/home/ben/.virtualenvs/main_env/lib/python3.6/site-packages/numpy/lib/arraypad.py", line 517, in _as_pairs          
raise ValueError("index can't contain negative values")                                                            
 ValueError: index can't contain negative values      
```

weepingwillowben:
Getting another error now:

```
    return super().step(action, observe)
  File "/home/weepingwillowben/.local/lib/python3.7/site-packages/pettingzoo/utils/wrappers.py", line 63, in step
    next_obs = self.env.step(action, observe=observe)
  File "/home/weepingwillowben/.local/lib/python3.7/site-packages/pettingzoo/butterfly/prospector/prospector.py", line 773, in step
    agent_bg_rects = self.background.update(agent.rect, self.dirty_fences)
  File "/home/weepingwillowben/.local/lib/python3.7/site-packages/pettingzoo/butterfly/prospector/prospector.py", line 421, in update
    self.dirty_rects.append(self.rects[top_y][right_x])
IndexError: list index out of range
```

I suspect the true problem is that the agent gets through the fence and is wandering way outside the screen.

weepingwillowben:
The second error can be reproduced with this code:

```
from pettingzoo.butterfly import prospector_v1
import time
import random
import numpy as np
import supersuit
import gym
import random
env = prospector_v1.env()
env.seed(10)
np.random.seed(10)
random.seed(10)
orig_obs = env.reset()
game_len = 0
for x in range(20000):
    for agent in env.agent_iter(1000000):
        reward, done, info = env.last()
        game_len += 1
        if (game_len // 20) % 100 == 0:
[...]

PR #217, opened 2020-10-03
Add boundary collision correction code; Fix gold bug on reset

yashaslokesh:
Gold bug was that gold was still attached to prospectors/bankers when environment reset. Fixed that.
Also, changed some integers to use the constant values defined in `constants.py`.


yashaslokesh:
For fixing #215

yashaslokesh:
Tests and flake8 passed

weepingwillowben:
Looks good to me

PR #220, opened 2020-10-04
example env constructor uses the wrong arg

yutaizhou:


justinkterry:
thanks a ton!

justinkterry:
Just out of curiosity- what makes you so uniquely interested in Knights Archers Zombies?

yutaizhou:
Its notion of agent death is what is interesting. Similar to MAgent's notion of health. We are exploring counterfactual actions and states in our algorithm, and the concept of health/death is a common case of that!

Counter question: why have the Butterfly environments output pixel observations specifically? And are there plans to output state vector for rest of the Butterfly envs like how Prisoner does it?

justinkterry:
I mean historically in reinforcement learning, graphical observations are generally viewed as a more natural/more challenging /more real world task. Like how DeepMind learning Atari games with only pixel outputs was such a big deal.

We added continuous actions to prison just because it's intended as a debugging environment and we wanted all types of neural networks to naturally be supported. It also seems like pistonball is the only environment where nongraphical observations would be super fitting.

yutaizhou:
Right. I was coming more from the perspective of decoupling pixel representation learning from RL. For something like MAgent where it's a grid environment, it makes complete sense to represent it as an "image-like" env. For more image-like env like KAZ, state vector may achieve that decoupling well

PR #225, opened 2020-10-13
Fix physics issues, adjust collisions, drawing, and rewards issues

yashaslokesh:
Work on fixing the various physics issues found during training. Originally, I was manually adjusting the position and velocity of physics objects which broke the simulation. This PR uses forces and smaller step sizes for the pymunk space for stabler simulations.


weepingwillowben:
addresses #215

weepingwillowben:
What are the changes to rewards? Just implementation fixes or something that needs to be documented?

yashaslokesh:
This is an implementation fix. The `ind_reward` parameter was not being used when an individual agent completed a task.

yashaslokesh:
This latest commit fixes a bug where rewards would be given out if the gold collided with the banker's body without checking if the gold collided with the front of the banker. This change was already made in the prospector script used for training

PR #231, opened 2020-10-16
Fix prospector render() issues with 'rgb_array' mode.

yashaslokesh:
The screen was not being updated when 'rgb_array' mode was turned on. This commit fixes that and also makes sure the pygame.display.flip() function is not called when 'rgb_array' mode is turned on


yashaslokesh:
Tests passed locally

weepingwillowben:
Thanks

PR #233, opened 2020-10-19
Removed opponent's hands from RLcard observation

lssr:
Updated Uno, Dou Dizhu, and Gin Rummy environments to remove the opponent's hand from the observation space. I added an environment argument, `full_observation_space`, to specify if the full observation space should be returned every time that `observe()` is called. 

Also, I updated the docs accordingly.

### DIFF
--- a/pettingzoo/classic/rlcard_envs/rlcard_base.py
+++ b/pettingzoo/classic/rlcard_envs/rlcard_base.py
@@ -51,7 +51,7 @@ class RLCardBase(AECEnv):
     def step(self, action, observe=True):
         obs, next_player_id = self.env.step(action)
         next_player = self._int_to_name(next_player_id)
-        self._last_obs = obs['obs']
+        self._last_obs = self.observe(self.agent_selection)
         if self.env.is_over():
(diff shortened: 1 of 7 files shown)


pettingzoo/classic/rlcard_envs/rlcard_base.py
```diff
@@ -71,7 +71,7 @@ def reset(self, observe=True):
         self.infos[self._int_to_name(player_id)]['legal_moves'] = obs['legal_actions']
         self._last_obs = obs['obs']
         if observe:
-            return obs['obs'].astype(self._dtype)
+            return self.observe(self.agent_selection)
```
weepingwillowben, line 74: Don't you also need to do something like this in the step() method?

weepingwillowben:
This is also failing the flake8 tests. You can run the `check_style.sh` script to run these locally.

lssr:
- Fixed flake8 issues.
- Updated `step()` for `rlcard_base.py`.

justinkterry:
Luis, can you please also change the argument name per the discussion in slack!?

justinkterry:
@weepingwillowben are we happy now?

Issue #222, opened 2020-10-06
Invalid Observation Values with MAgent Envs


jdpena:
The default config in `battle` and `battlefield` causes the observation values to be outside the `[0,2]` allowed range, leading to a ValueError. Note the `-0.105` in the error below. The obs shapes are correct. 

```
ValueError: ('Observation outside expected value range', Box(6929,), array([ 0.    ,  0.    ,  0.    , ..., -0.105 ,  0.3875,  0.45  ],
```

I set `step_reward`, `dead_penalty`, and `attack_penalty` to be 0 (or any value between `[0,2]`) and it fixed the issue.

Edit: it appears all MAgent envs suffer from this issue except `tiger-deer`.

yutaizhou:
To piggy back off this question, can someone help me understand the observation of battle or battlefield? The obs_space is (13,13,41), which represents:
empty 1,
obstacle 1,
red 1,
blue 1,
minimap_red 1,
minimap_blue 1,
binary_agent_id 10,
one_hot_action 21,
last_reward 1,
agent_position ??

Which leave us with 38 of the 41 filters of axis=2 explained. I am not sure what agent_position means, so i don't know if that description would take up 3 filters to represent.

justinkterry:
We'll reply to this soon, but let me first mention that we're planning to overhaul magent observations and documentation in the next few weeks (after the death thing is fixed)

yutaizhou:
Here is my temporary workaround for OP's issue:

```
env = battlefield_v1.env()
default_space = env.observation_spaces['red_0']
desired_space = GymBox(low = -default_space.high,
                 high = default_space.high,
                 shape = default_space.shape,
                 dtype = default_space.dtype)
for agent_id in env.observation_spaces.keys():
      env.observation_spaces[agent_id] = desired_space
```

yutaizhou:
actually, The bounds are probably better off being inf and -inf, since Battle can have +5 rewards or neg reward. That is if no further reward modification is done by users, e.g. sum up all rewards from agents within a cycle and use it as team reward.

https://www.pettingzoo.ml/magent/battle

weepingwillowben:
Thanks for pointing out the incorrect observation bounds. It seems as though the culprit is the last_reward, and as the minimap observations are bounded between 0 and 2, and all other observations seem to be bounded between 0 and 1. We are not big belivers of infinite bounds, so perhaps reward observations can just be bounded properly. 

[...]

PR #234, opened 2020-10-19
Magent doc updates

weepingwillowben:
Fixed magent documentation and also updated the bounds on the reward observation as mentioned in #222 .

Note that this does not detail the observation for combined_arms, because the exact order of the map channels in that environment (or any magent environment with more than 2 "groups" is absolutely mindboggling. In particular, the code that handles this logic is in https://github.com/PettingZoo-Team/MAgent/blob/master/src/gridworld/GridWorld.cc#L897, and what it does seems very strange in the case where there are more than 2 groups.


weepingwillowben:
This is all in the big PR #238

PR #246, opened 2020-11-06
updated average total rewards

weepingwillowben:


weepingwillowben:
Things that need to be reviewed:
* Why did pistonball's reward change so much? It probably has to do with the max_cycles change.
* Why did pursuit's reward change so much? Not clear.

Expected changes:
* Prison is much longer running environment
* Prospector and entombed were affected by bug fixes
* MPE was effected by reducing the default of max_cycles by a factor of 4
* And waterworld was effected by bug fixes

PR #259, opened 2020-12-01
added aec waterworld rewards

weepingwillowben:
Waterworld "action magnitude penalty" is now allocated between cycles in the .rewards dictionary. 

Manually checked that accumulated rewards are exactly the same. Inclined not to bump version for this reason.


justinkterry:
We most definitely should bump version for something like this. Also Ryan's working on all the other waterworld fixes and was going to do this. I swear I told you that. Anyhow we should wait to release this until Ryan does all his things.

weepingwillowben:
Yeah, I made this change because Ryan asked about the waterworld fixes

PR #262, opened 2020-12-18
fixed bug that always returns first agent's observation

ChaceAshcraft:
I noticed that I was getting the same agent's observations for each agent, and I think the bug is in the couple lines I changed. I tried to hold to the intent of the function and change as little as possible. If this does not reflect the intent, or is not a bug, please let me know.


justinkterry:
Hey thanks a ton, I'll look at this soon. Let me first say though that our CI currently isn't working--we're in the middle of trying to transition from travis to github actions.

weepingwillowben:
Can confirm the bug and the fix. Thank you for the contribution.

ChaceAshcraft:
Absolutely! Thank you for your work!

PR #269, opened 2020-12-27
Hanabi illegal actions

weepingwillowben:
* Fixed hanabi illegal termination to have the correct reward (-score, so that it terminates with 0 reward, just as it would if it lost).
* Also fixed illegal terminate wrapper


weepingwillowben:
This PR is based off @paarasbhandari 's classic observation PR, so tests won't run

justinkterry:
The documentation also needs to be fixed

weepingwillowben:
Ok, I added a line to the documentation.

PR #273, opened 2020-12-31
Allow for external test calling

justinkterry:
This makes our tests cleaner to call outside of pettingzoo (i found this out working on bird flocking)


justinkterry:
This still doesn't quite have the desired behavior. It requires:

```
from pettingzoo.tests import api_test
api_test.api_test(env, num_cycles=1, render=False, verbose_progress=True)
```

where as ideally it would be 

```
from pettingzoo.tests import api_test
api_test(env, num_cycles=1, render=False, verbose_progress=True)
```

weepingwillowben:
Looks fine to me.

PR #274, opened 2020-12-31
Rename tests to test

justinkterry:
Ideally people will be using our tests for other environments. A plural name like we had is very strange in an import


justinkterry:
My other PR (or a fixed version of it) has to be merged first, then I'll fix this

PR #280, opened 2021-01-04
Remove black death argument from KAZ

justinkterry:
The logic has been moved to a supersuit wrapper


pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -475,18 +470,6 @@ def step(self, action):
         done = not self.run or self.frames >= self.max_cycles
         self.dones = {a: done for a in self.agents}
 
-        if self._agent_selector.is_last() and not self.black_death:
```
weepingwillowben, line 478: This part still needs to be here for regular agent death to occur.

PR #298, opened 2021-01-16
Bump pymunk version and fix setup.py

justinkterry:


requirements.txt
```diff
@@ -1,7 +1,11 @@
 gym>=0.18.0
 pygame==2.0.0
 numpy>=1.18.0
+<<<<<<< HEAD
```
weepingwillowben, line 4: The reason tests are failing is because the git merge conflict guards got inserted into in the code.

justinkterry: The magical errors are the first test, not this. That wouldn't have fixed it, but hopefully PyMunk 5.7.1 will prevent me from having to fix i.t

Issue #272, opened 2020-12-29
ALSA lib pcm.c:7963:(snd_pcm_recover) underrun occurred


ewanlee:
Hello, when I parallelized the pursuit environment in SISL, the bug shown in the title appeared. This bug will appear when the number of parallels is large, but it will not affect the running of the program.

Specifically, I first wrap the environment through the following wrapper:

```python
class parallel_wrapper():
    def __init__(self, parallel_env):
        self.env = parallel_env
        self.agents = sorted(parallel_env.action_spaces.keys())
        self.action_space = [parallel_env.action_spaces[agent] for agent in self.agents]
        self.observation_space = [parallel_env.observation_spaces[agent] for agent in self.agents]

    def reset(self):
        obs = self.env.reset()
        obs_n = [obs[agent].flatten() for agent in self.agents]
        return obs_n

    def seed(self, seed):
        return self.env.seed(seed)

    def step(self, action_n):
        action_n = np.argmax(action_n, axis=1)
        actions = {agent:action_n[i] for i, agent in enumerate(self.agents)}
        obs, rews, dones, infos = self.env.step(actions)
        obs_n = [obs[agent].flatten() for agent in self.agents]
        reward_n = [rews[agent] for agent in self.agents]
        done_n = [dones[agent] for agent in self.agents]
        info_n = {'n': [infos[agent] for agent in self.agents]}
        
        return obs_n, reward_n, done_n, info_n

if scenario_name == 'pursuit':
            from pettingzoo.sisl import pursuit_v2
            env = parallel_wrapper(pursuit_v2.parallel_env())
            return env
```

Next, I use `SubprocVecEnv` in the `baselines` library to parallelize the wrapped environment:
```python
def worker(remote, parent_remote, env_fn_wrapper):
    parent_remote.close()
    env = env_fn_wrapper.x()
    while True:
        cmd, data = remote.recv()
        if cmd == 'step':
            ob, reward, done, info = env.step(data)
            if all(done):
                ob = env.reset()
            remote.send((ob, reward, done, info))
        elif cmd == 'reset':
            ob = env.reset()
            remote.send(ob)
        elif cmd == 'reset_task':
            ob = env.reset_task()
            remote.send(ob)
        elif cmd == 'close':
            remote.close()
            break
        elif cmd == 'get_spaces':
            remote.send((env.observation_space, env.action_space))
        elif cmd == 'get_agent_types':
[...]

PR #299, opened 2021-01-16
Suppress alsa errors on Linux

justinkterry:
So it turns out that issue #272 was our fault, sort of. Initializing pygame initializes an audio connect through SDL, which defaults to alsa on linux. Doing this is prone to absolutely insane error messages. This should fix that.


justinkterry:
I just tested this and it works.

PR #303, opened 2021-01-29
Fixes to RPS and RPSLS

lssr:
- changed constant names to uppercase convention (`rock` to `ROCK`).
- Added the sentence "The observation is the last opponent action and its space ..." to `rps` and `rpsls` documentation to explicitly state what is included in each observation.


weepingwillowben:
looks good to me

PR #313, opened 2021-02-07
Bug fix in simple_adversary_v2

nishantkr18:
The number of "good" agents and landmarks can now be changed via the api call.


RyanNavillus:
Nice catch, thanks!

RyanNavillus (approved):

Issue #314, opened 2021-02-07
ZeroDivisionError: division by zero


SachinKonan:
Traceback (most recent call last):
  File "test_piston_ball.py", line 64, in <module>
    observations, rewards, dones, infos = parallel_env.step(actions)
  File "/opt/conda/envs/marl/lib/python3.6/site-packages/pettingzoo/utils/to_parallel.py", line 36, in step
    self.aec_env.step(actions[agent])
  File "/opt/conda/envs/marl/lib/python3.6/site-packages/pettingzoo/utils/wrappers.py", line 238, in step
    super().step(action)
  File "/opt/conda/envs/marl/lib/python3.6/site-packages/pettingzoo/utils/wrappers.py", line 63, in step
    self.env.step(action)
  File "/opt/conda/envs/marl/lib/python3.6/site-packages/pettingzoo/utils/wrappers.py", line 156, in step
    super().step(action)
  File "/opt/conda/envs/marl/lib/python3.6/site-packages/pettingzoo/utils/wrappers.py", line 63, in step
    self.env.step(action)
  File "/opt/conda/envs/marl/lib/python3.6/site-packages/pettingzoo/butterfly/pistonball/pistonball.py", line 388, in step
    global_reward = (100 / self.distance) * (self.lastX - ball_min_x)
ZeroDivisionError: division by zero

When the distance is zero, or self.lastX - self.wall_width = 0, then this error pops up.

RyanNavillus:
Could you share the code you used to produce this error? `self.distance` is a very large number by default and does not change while acting in the environment. We haven't run into this issue in practice, but I'll look into it.

SachinKonan:
Here is the code: `from pettingzoo.butterfly import pistonball_v3
from alex_net import Encoder
from policy import Policy, Experience
from torch.distributions import Categorical
import torch
import torch.optim as optim
from torch.autograd import Variable

#torch.autograd.set_detect_anomaly(True)

encoding_size = 100
action_space = 3
eps = 1e-3
lr = 0.0001
epochs = 100
n_agents = 5
max_cycles = 200

parallel_env = pistonball_v3.parallel_env(n_pistons=n_agents, local_ratio=0.9, time_penalty=-0.1, continuous=False,
                                          random_drop=True, random_rotate=True, ball_mass=0.75, ball_friction=0.3,
                                          ball_elasticity=1.5, max_cycles=max_cycles)


alex_encoder = Encoder()
[...]

PR #315, opened 2021-02-08
Pistonball

RyanNavillus:
Addresses #314


weepingwillowben:
Given that there were other changes to pistonball since the last release, should the initial ball placement be rethought?

For example, should the ball placement be uniformly distributed between the left and right walls? Or uniformly distributed between the halfway point and the right wall?

RyanNavillus:
I don't know what benefit that would provide? I think the current ball placement very close to the right wall makes sense since the goal is to move the ball all the way left. We could consider rewriting the local vs global reward, which is programmed in a sort of weird way.

PR #317, opened 2021-02-10
Fix crash caused by modulo zero - when all agents die, black death

himelbrand:
Added an if statement checking if the agents_order of the agent selector is empty before calling next on the agent_selector.

I came across this bug, which caused division (modulo) by zero.

This happens when using parallel env + black death wrapper combined.

Specifically, it happens when the last agent is killed by a zombie, as far as I can tell.

This extra line prevents the run to crash.

### DIFF
--- a/pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
+++ b/pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
@@ -483,8 +483,8 @@ class raw_env(AECEnv, EzPickle):
             self.kill_list = []
 
             self._agent_selector.reinit(_live_agents)
-
-        self.agent_selection = self._agent_selector.next()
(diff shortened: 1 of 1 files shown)


justinkterry:
@weepingwillowben

weepingwillowben:
Thanks for bringing this up. I'll look at this shortly.

weepingwillowben:
Are you using the black_death_v0 wrapper?

weepingwillowben:
The black_death_v0 wrapper has a known bug. Please use the black_death_v1 wrapper available in the latest version of supersuit.

weepingwillowben:
While I can't reproduce the error experimentally, I can see why the div by 0 error would occur. 

The fix itself, it looks good. Thanks for the contribution, it will be included in the next release.

PR #320, opened 2021-02-14
State method

Rodridelazcano:
- Added .state() method and .state_space attribute for AEC and parallel env
- Implemented in butterfly env
- Tested with pytest


pettingzoo/test/state_test.py
```diff
@@ -0,0 +1,73 @@
...
+        if env.state_space.high.shape != env.state_space.shape:
+            assert False, "Environment's state_space.high and state_space have different shapes"
+
+
+def test_state(env, num_cycles):
+    env.reset()
+    state_0 = env.state()
+    for agent in env.agent_iter(env.num_agents * num_cycles):
+        observation, reward, done, info = env.last(observe=False)
+        print('ok')
```
weepingwillowben, line 34: This print should be removed.

pettingzoo/test/state_test.py
```diff
@@ -0,0 +1,73 @@
...
+        if done:
+            action = None
+        else:
+            action = env.action_spaces[agent].sample()
+
+        env.step(action)
+        new_state = env.state()
+        print(env.state_space)
```
weepingwillowben, line 42: This print as well.

pettingzoo/butterfly/cooperative_pong/cooperative_pong.py
```diff
@@ -217,6 +217,8 @@ def __init__(self, randomizer, ball_speed=9, left_paddle_speed=12, right_paddle_
         original_color_shape = (original_shape[0], original_shape[1], 3)
         # self.observation_space = [gym.spaces.Box(low=0.0, high=1.0, shape=(original_shape), dtype=np.float32) for _ in range(self.num_agents)]
         self.observation_space = [gym.spaces.Box(low=0, high=255, shape=(original_color_shape), dtype=np.uint8) for _ in range(self.num_agents)]
+        # define the global space of the environment or state
+        self.state_space = gym.spaces.Box(low=0, high=255, shape=((original_shape[0], original_shape[1] * 2, 3)), dtype=np.uint8)
```
weepingwillowben, line 221: The dimensions of this space can be cleaner. 
shape=(self.s_height, self.s_width, 3)

pettingzoo/utils/env.py
```diff
@@ -24,6 +24,9 @@ def observe(self, agent):
     def render(self, mode='human'):
         raise NotImplementedError
 
+    def state(self):
+        raise NotImplementedError
```
weepingwillowben, line 28: You can give the NotImplementedError a message to make the error more clear.

weepingwillowben (commented):
So again, this contribution is really good. Like one of the best we have ever gotten from someone who is new. I have a few comments, but they are all really subtle things which took me months to figure out. 

general comments. 

* The error message for accessing the state_space in an environment which does not define it can be improved slightly. If you look here: https://github.com/Rodridelazcano/PettingZoo/blob/state-method/pettingzoo/utils/wrappers.py#L215 you can see a specific error message for a undefined attribute. You can make one for state space.
* A state() method should be added to the ParallelEnv and _parallel_env_wrapper. You can write a very short test for testing the state for a parallel env (just call the method and make sure the result is in the state space, no other checks necessary). You can put this method in the `state_test.py` file.
* Documentation should be added to the "Additional Environment API" section of the docs/api.md file. This documentation should tell the user: 1. What the state method is for. 2. What environments have the state right now and 3. What happens if an environment has not implemented the state method.

docs/api.md
```diff
@@ -52,6 +52,10 @@ PettingZoo models games as *Agent Environment Cycle* (AEC) games, and thus can s
 
 `action_spaces`: A dict of the action spaces of every agent, keyed by name. This cannot be changed through play or reseting.
 
+`state_space`: The space of a global observation of the environment. Currently, only `butterfly` environments have this attribute implemented. If `state_space` is called in an unsupported environment the following `AttributeError` will raise:
```
justinkterry, line 55: "Currently, only `butterfly` environments have this attribute implemented." is unnescesary

docs/api.md
```diff
@@ -52,6 +52,10 @@ PettingZoo models games as *Agent Environment Cycle* (AEC) games, and thus can s
...
+
+`Environment {} has no state_space attribute implemented yet.`
```
justinkterry, line 57: this should be in the same line as the rest

docs/api.md
```diff
@@ -62,6 +66,10 @@ PettingZoo models games as *Agent Environment Cycle* (AEC) games, and thus can s
 
 `observe(agent)`: Returns the observation an agent currently can make. `last()` calls this function.
 
+`state()`: Returns a global observation of the current state of the environment. Currently, only `butterfly` environments have this method implemented. If `state()` is called in an unsupported environment the following `NotImplementedError` will raise:
```
justinkterry, line 69: same comments as for state space

justinkterry (changes requested):
hey this looks really good. i made a few small comments on documentation

docs/api.md
justinkterry: Actually, including the specific error message isn't really needed. in the documentation at all, just mention that not all environments will support it (in many cases it doesn't even make sense to support this, like with Atari). Relatedly, "yet" should be removed from the error message.

docs/api.md
weepingwillowben: I would mention that not all environments have to implement the method

PR #321, opened 2021-02-14
fixed pistonball bug caused by pymunk

himelbrand:
When using pistonball.parallel_env()

I came across the following error:

Aborting due to Chipmunk error: The shape's body must be added to the space before the shape.
	Failed condition: shape->body->space == space
	Source:Chipmunk2D\src\cpSpace.c:423

I fixed this bug by adding the body (i.e. piston) before adding the shape (segment) to the space.

Hope this helps.

### DIFF
--- a/pettingzoo/butterfly/pistonball/pistonball.py
+++ b/pettingzoo/butterfly/pistonball/pistonball.py
@@ -195,7 +195,7 @@ class raw_env(AECEnv, EzPickle):
         segment = pymunk.Segment(piston, (0, 0), (self.piston_width - (2 * self.piston_radius), 0), self.piston_radius)
         segment.friction = .64
         segment.color = pygame.color.THECOLORS["blue"]
-        space.add(segment)
+        space.add(piston, segment)
         return piston
 
     def move_piston(self, piston, v):


justinkterry:
Me or Ryan will look into this later today, but I have two questions:

-What did you do to create this error? I've played pistonball about a million times now and haven't seen this.
-Can you please make sure you're using the version of pymunk specified in the library (5.7.x)?

justinkterry:
-Also, how on Earth did you figure out how to fix that error? If I understand correctly this is a problem we've been getting when we wanted to upgraded to pymunk 6.0 but could never figure out the origins of.

himelbrand:
What I did to get this error was (as far as I can tell) using the parallel API:

env = pistonball.parallel_env(n_pistons=15, local_ratio=0.2, time_penalty=-0.1, continuous=False, random_drop=True, random_rotate=True, ball_mass=0.75, ball_friction=0.3, ball_elasticity=1.5, max_cycles=900)

when I used this setting I got the error I mentioned, right when starting the run.

However, I did notice that I do have the wrong version of pymunk, I have version 6.0 installed.

And I found the fix for this error, by searching the web for the error string, and traversing the environment code to find the origin of the problem, with a bit of debugging I was able to find the issue and fix it.

So I'm not sure if this PR is even relevant since it solves an issue for a different version of pymunk and not the one used originally.

But I guess this fix is helpful for the time you decide to upgrade to version 6.0

justinkterry:
I merged it since we do want to upgrade to 6.0 in the near future (it gets rid of the stupid error message amongst other things). The one other environment that has this error with pymunk 6.0 is butterfly/prospector.

I assume the problem in prospector is here: https://github.com/PettingZoo-Team/PettingZoo/blob/2dc95d8b18e2d68cce5dcf68db0c0216737e10be/pettingzoo/butterfly/prospector/prospector.py#L220

Do you have any idea what the fix would be there? It's still not obvious to me.

justinkterry:
We figured it out, nevermind. Thanks again!

PR #330, opened 2021-02-20
bump butterfly versions

weepingwillowben:


justinkterry:
It's my understanding that cooperative pong doesn't need it's version bumped since it doesn't use pymunk and didn't have any default arguments changed.

PR #332, opened 2021-02-22
Fix the default max_cycles according to MPE documentation.

uhlajs:
Documentation of MPE environments:
The game terminates after the number of cycles specified by the max_cycles environment argument is executed.
The default for all environments is **100 cycles**. Note that in the original OpenAI source code, the default value was 25, not 100.

### DIFF
--- a/pettingzoo/mpe/simple_adversary_v2.py
+++ b/pettingzoo/mpe/simple_adversary_v2.py
@@ -4,7 +4,7 @@ from pettingzoo.utils.conversions import parallel_wrapper_fn
 
 
 class raw_env(SimpleEnv):
-    def __init__(self, N=2, max_cycles=25):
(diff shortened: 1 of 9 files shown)


justinkterry:
@weepingwillowben did we forget to fix the documentation when we decreased the number of cycles? What do you want to do here?

weepingwillowben:
Yeah, the documentation should be fixed, not the code.

justinkterry:
@weepingwillowben please do a PR fixing the documentation.

weepingwillowben:
Erm. I am really confused now. The documentation is up to date: https://www.pettingzoo.ml/mpe/simple_adversary and hasn't changed since november.

What documentation had it at 100?

weepingwillowben:
Ok, I found the problematic documentation. Fixed in #336

PR #334, opened 2021-02-22
Added support for num_forests!=2 in mpe/simple_world_comm

vstark21:
Environment just supports num_forests=2 but not any other value you can see at this [line](https://github.com/PettingZoo-Team/PettingZoo/blob/master/pettingzoo/mpe/simple_world_comm_v2.py#L9). So I have added the support for other values as well.

### DIFF
--- a/docs/mpe/simple_world_comm.md
+++ b/docs/mpe/simple_world_comm.md
@@ -55,5 +55,5 @@ simple_world_comm.env(num_good=2, num_adversaries=4, num_obstacles=1,
 
 `max_cycles`:  number of frames (a step for each agent) until game terminates
 
-`num_forests`: number of forests that hide agents inside from being seen
+`num_forests`: number of forests that can hide agents inside from being seen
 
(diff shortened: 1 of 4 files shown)


justinkterry:
@weepingwillowben

weepingwillowben:
I'll take a look at this soon

weepingwillowben:
So I gave it a look over, and it generally looks good, it can handle edge cases and the logic all checks out. 

It would be preferable if this change also changed the associated documentation here: https://github.com/PettingZoo-Team/PettingZoo/blob/master/docs/mpe/simple_world_comm.md#arguments

vstark21:
Changed the associated documentation as well.

justinkterry:
Why is there a documentation change for pong in this?

vstark21:
Change is in atari/pong_v1.py file, added a comment for assert statement for num_players==2or4. I missed mentioning it in the first comment.

justinkterry:
Thanks a ton!

PR #346, opened 2021-03-08
fixed from parallel wrapper

weepingwillowben:
@rodrigodelazcano I think this is the correct fix to your problem. The parallel agent API does not take in a done, it takes in a smaller dictionary when you don't want to step the agent.


rodrigodelazcano:
This solves the issue. Thanks!

PR #349, opened 2021-03-09
added back magent test, added asserts that agent is in world

weepingwillowben:
Ok, this PR should be OK now. This PR should fix the old segfault issue.


weepingwillowben:
Ok, now the observation docs for the MAgent PR have the minmap mode specified: https://github.com/PettingZoo-Team/PettingZoo/blob/magent_test/docs/magent/battlefield.md#observation-space

PR #388, opened 2021-05-27
Mpe state

rodrigodelazcano:
state attribute and state method for mpe environments. The state for each env is a a concatenation of the observation array from each agent.


benblack769:
Same problem as #389  with too many commits.

PR #390, opened 2021-05-27
correct minor magent doc typo

rodrigodelazcano:


justinkterry:
Wait what's the change?

rodrigodelazcano:
In the Magent docs it says: If an agent "in" on a coordinate...
I though that the "in" is a small typo that needs to be changed for "is"

PR #391, opened 2021-05-28
Mpe state

rodrigodelazcano:
Added state_space and state method to mpe environments. The state is is defined as a 1D array formed by the concatenation of the observation arrays of each agent.

### DIFF
--- a/pettingzoo/mpe/_mpe_utils/simple_env.py
+++ b/pettingzoo/mpe/_mpe_utils/simple_env.py
@@ -66,6 +66,10 @@ class SimpleEnv(AECEnv):
     def observe(self, agent):
         return self.scenario.observation(self.world.agents[self._index_map[agent]], self.world).astype(np.float32)
 
+    def state(self):
+        states = tuple([self.scenario.observation(self.world.agents[self._index_map[agent]], self.world).astype(np.float32) for agent in self.agents])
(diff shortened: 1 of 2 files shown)


test/pytest_runner.py
```diff
@@ -23,7 +23,7 @@ def test_module(name, env_module):
     if "classic/" not in name:
         max_cycles_test(env_module)
 
-    if "butterfly/" in name:
+    if ("butterfly/" or "mpe/") in name:
```
benblack769, line 26: Python does not work like this. It will evaluate `bool("butterfly/" or "mpe/")` -> `True`
and then ask if `True in name`.

pettingzoo/mpe/_mpe_utils/simple_env.py
```diff
@@ -62,6 +66,10 @@ def seed(self, seed=None):
     def observe(self, agent):
         return self.scenario.observation(self.world.agents[self._index_map[agent]], self.world).astype(np.float32)
 
+    def state(self):
+        states = tuple([self.scenario.observation(self.world.agents[self._index_map[agent]], self.world).astype(np.float32) for agent in self.agents])
```
benblack769, line 70: To be on the safe side, use `self.possible_agents`

PR #397, opened 2021-05-31
Connect 4 PyGame graphics

Kyle-Sang:
Added code for rendering the Connect4 environment via Pygame. Correct placement of pieces is based on the images themselves, so screen width and height can be modified as long as it matches the 99*86 resolution of the board image.


justinkterry:
Please include a gif that I can see

Kyle-Sang:
Connect4_Test

justinkterry:
and make it pass tests

benblack769:
So unlike in butterfly environments like pistonball, in connect 4, rendering is not needed for generating observations. 

So there is no need for any drawing or any pygame calls at all in the __init__, reset, step or observe functions. 

You can see how this is accomplished in atari: https://github.com/PettingZoo-Team/PettingZoo/blob/master/pettingzoo/atari/base_atari_env.py#L172 or MPE: https://github.com/PettingZoo-Team/PettingZoo/blob/master/pettingzoo/mpe/_mpe_utils/simple_env.py#L173

benblack769:
Eliminating unnecessary drawing is important because:

1. Drawing is very slow and
2. Initializing the display when not needed causes all sorts of problems (this is the reason tests are failing)

Kyle-Sang:
Connect4_Updated_Test

pettingzoo/classic/connect_four/connect_four.py
```diff
@@ -117,11 +129,51 @@ def reset(self):
...
+                if self.board[i] == 1:
+                    self.screen.blit(red_chip, ((i % 7) * (tile_size) + (tile_size * (6 / 13)), int((i / 7)) * (tile_size) + (tile_size * (6 / 13))))
+                elif self.board[i] == 2:
+                    self.screen.blit(black_chip, ((i % 7) * (tile_size) + (tile_size * (6 / 13)), int((i / 7)) * (tile_size) + (tile_size * (6 / 13))))
+
+            pygame.display.update()
+
+        observation = np.array(pygame.surfarray.pixels3d(self.screen))
+
         print("{}'s turn'".format(self.agent_selection))
```
benblack769, line 166: This should probably be removed completely.

pettingzoo/classic/connect_four/connect_four.py
```diff
@@ -2,11 +2,21 @@
...
+    import pygame
+    from os import path as os_path
+    cwd = os_path.dirname(__file__)
+    image = pygame.image.load(cwd + '/' + path)
+    return image
+
+
 def env():
     env = raw_env()
     env = wrappers.CaptureStdoutWrapper(env)
```
benblack769, line 22: This should also be removed. This implements a `ansi` render mode, that will no longer work once the prints are removed.

benblack769:
The `rgb_array` update looks good. Just a couple of comments about removing printing.

pettingzoo/classic/connect_four/connect_four.py
benblack769: By this I mean the CaptureStdoutWrapper specifically.

Issue #286, opened 2021-01-05
Prospector stateful environment issue


benblack769:
Reseeding an environment and resetting should give an identical environment state every time, whether it is a newly created environment, or an environment which just completed an previous episode. This is what it means for a gym environment to be properly episodic. Every environment is tested to make sure this is the case, but one environment fails: Prospector.

Prospector is currently [excluded](https://github.com/PettingZoo-Team/PettingZoo/blob/master/test/pytest_runner.py#L20) from the relevant test, to allow the tests to pass in CI, however it fails the seed_test.

Visual inspection of observations and rewards did not detect a difference between the tested environments, so the test failure is likely not meaningful. However, it would be good to make sure. My current hypothesis is that the physics collision cache holds state in between resets, making minor changes to the collision dynamics in the environment.

PR #398, opened 2021-05-31
issue 286

rodrigodelazcano:
#286 already solved. I tested the seed test for the prospector env with more than 10000 cycles without any errors. 
I proceeded to include prospector env in the seed test.


justinkterry:
Could you try it for 1 million to be safe?

justinkterry:
If that's too much for your laptop message me

benblack769:
I suspect this was solved by the pymunk update awhile back. It definitely failed consistently with only 10000 cycles back then.

justinkterry:
And that's why I demanded we bump versions for new pymunk. LGTM.

PR #399, opened 2021-06-06
fix agent binary_id

rodrigodelazcano:
In the magent adversarial pursuit environment the binary_id agent feature was missing. The documentation for this environment states that this observation is implemented. Also, it is good to add it so that the observations are consistent between the environments


benblack769:
So the reason this was left out was because it was set this way in the original paper. Once the extra_features was made optional, I guess it no longer made a lot of sense to exclude it though.

PR #400, opened 2021-06-07
rllib run and rendering files

kaanozdogru:
Adding to new files to PettingZoo/tutorials. These files are modified versions of the ones in rl_scratch. run_rllib.py is almost the the same as the original code except that the commented out code is deleted. There are some modifications in render_game.py. First, I changed the checkpoint path where the trained policy is saved in the VM because the previous path didn't existed. Secondly, I made some changes in the preprocessing of the environment so that the shape of the observation from the environment and input of the policy model are the same.


justinkterry:
rename files to render_rllib_pistonball and rllib_pistonball and I'll merge

benblack769:
Since this is a tutorial, perhaps comments in the code describing what is going on would be helpful. In particular, a new user would likely be baffled by the weird RLlib parts of the code.

PR #402, opened 2021-06-11
Go pygame graphics

Kyle-Sang:


Kyle-Sang:
GO_Test

justinkterry:
(ben if you're wondering the graphics assets still need to be updated)

Kyle-Sang:
Go_test_updated

justinkterry:
@benblack769 you happy with this?

benblack769:
So it looks like CI tests aren't testing rendering.

In order to test rendering in rgb_array mode,  you should make sure that passing in rgb_array returns an image of the screen.

The following script will test for this. You can run the script, and then check that the saved image (`connect4.png` in this case) is correct. It doesn't appear to work right now. If you can go back and fix this for all the games which you are adding rendering support for, that would be great. 

```python
from pettingzoo.classic import connect_four_v3
from pettingzoo.butterfly import cooperative_pong_v2
from PIL import Image

env = connect_four_v3.env()
env.reset()
array = env.render(mode='rgb_array')

im = Image.fromarray(array)
im.save("connect4.png")
```

benblack769:
So if it renders the same when you run the script as when it renders on the screen (you can check the file that the script saves), then I am good with all 4 PRs.

PR #410, opened 2021-06-20
Number of players argument for allowed games

kaanozdogru:


kaanozdogru:
rlcard changed depracted the function get_action_num for get_num_actions in the new version which gives the shape of the action space.

PR #417, opened 2021-07-03
Fixed go encoding error

SonSang:


jkterry1:
@benblack769

benblack769:
This seems to be a valid fix. Good catch!

PR #421, opened 2021-07-10
observation space is changed from 3d to 1d vector

kaanozdogru:


benblack769:
Can you get this to pass tests? For the error I see now, the shape of a Box needs to be a tuple, not an int. There needs to be a comma in the parenthasis. Look at other environments for syntax there.

kaanozdogru:
@benblack769 so I think I don't have any issues left with the observation space. I would really appreciate if you can review it. However there is an out of bounds error related to the `action_mask`. Is this the error @rodrigodelazcano was talking about?

pettingzoo/classic/rlcard_envs/dou_dizhu.py
```diff
@@ -34,9 +34,10 @@ def _scale_rewards(self, reward):
...
         if self._opponents_hand_visible:
-            observation = obs['obs'].astype(self._dtype)
+            observation = np.zeros(901, dtype=self._dtype)
+            observation[:len(obs['obs'])] = obs['obs'].astype(self._dtype)
         else:
-            observation = obs['obs'][[0, 2, 3, 4], :, :].astype(self._dtype)
+            observation = np.concatenate((obs['obs'][:54].astype(self._dtype), obs['obs'][108:648].astype(self._dtype)))
 
         legal_moves = self.next_legal_moves
         action_mask = np.zeros(309, int)
```
rodrigodelazcano, line 43: The new action mask should be of size 27472.

pettingzoo/classic/rlcard_envs/dou_dizhu.py
```diff
@@ -34,12 +34,13 @@ def _scale_rewards(self, reward):
     def observe(self, agent):
         obs = self.env.get_state(self._name_to_int(agent))
         if self._opponents_hand_visible:
-            observation = obs['obs'].astype(self._dtype)
+            observation = np.zeros(901, dtype=self._dtype)
+            observation[:len(obs['obs'])] = obs['obs'].astype(self._dtype)
         else:
-            observation = obs['obs'][[0, 2, 3, 4], :, :].astype(self._dtype)
+            observation = np.concatenate((obs['obs'][:54].astype(self._dtype), obs['obs'][108:648].astype(self._dtype)))
```
rodrigodelazcano, line 40: What observation features are being extracted here? I think that the new version of dou dizhu has different observation sizes for the landlord and for the peasants.

kaanozdogru: So here I'm extracting `current_hand`, `last_action` and `last_9_actions` whether it's a landlord or peasant. They have the same indices for those states.

jkterry1:
Closed by #429

PR #430, opened 2021-07-21
Go and chess 8 step history

Kyle-Sang:


jkterry1:
@benblack769

benblack769:
So there was no need to change np.bool to bool. In the backend, numpy treats them exactly the same way.

benblack769:
There is no need to track the observation_history, only the board_history.

pettingzoo/classic/go/go_env.py
```diff
@@ -98,7 +101,11 @@ def _encode_rewards(self, result):
     def observe(self, agent):
         current_agent_plane, opponent_agent_plane = self._encode_board_planes(agent)
         player_plane = self._encode_player_plane(agent)
+
+        self.board_history = np.dstack((current_agent_plane, opponent_agent_plane, self.board_history[:, :, :-2]))
```
benblack769, line 105: This logic should go in the step() function rather than the observe function. The reason is that observe can be called many times, or no times, after a step(). In other words, observe() should never change the env state, it should only be a getter. step() should change the env state.

pettingzoo/classic/go/go_env.py
```diff
@@ -44,14 +44,17 @@ def __init__(self, board_size: int = 19, komi: float = 7.5):
...
-            [spaces.Dict({'observation': spaces.Box(low=0, high=1, shape=(self._N, self._N, 3), dtype=np.bool),
+            [spaces.Dict({'observation': spaces.Box(low=0, high=1, shape=(self._N, self._N, 3), dtype=bool),
                           'action_mask': spaces.Box(low=0, high=1, shape=((self._N * self._N) + 1,), dtype=np.int8)})
              for _ in range(self.num_agents)])
 
         self.action_spaces = self._convert_to_dict([spaces.Discrete(self._N * self._N + 1) for _ in range(self.num_agents)])
 
         self._agent_selector = agent_selector(self.agents)
 
+        self.board_history = np.zeros((board_size, board_size, 8))
```
benblack769, line 55: This should be an np.bool type, and of size 16, not 8.

benblack769:
Once #432 gets merged, the problems with the observation should be a bit more clear, as the observation will not be in the observation space.

jkterry1:
The np.bool change was my request. Numpy is depricating it.

I merged your PR, Kyle can merge in master now and get the error.

benblack769:
If numpy deprecates np.bool, they will also deprecate bool. They are literally the same thing.

benblack769:
@jkterry1

benblack769:
Or, I mean, numpy literally converts it to mean the same thing.

jkterry1:
Np depricated np.bool in favor of bool exactly because they mean the same thing. Check old CI test warnings or Google it to confirm.

benblack769:
@jkterry1 I see. Very strange.  This issue claims the correct replacement is np.bool8: https://github.com/inducer/compyte/issues/32, which makes more sense to me but I guess official docs here https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations aggress with you.

benblack769:
Looks good. I am happy with the code changes. Just need to bump versions.

PR #437, opened 2021-07-31
Pygame Load Image Fixes

jkterry1:
So the way we got the PyGame image loading logic used throughout pettingzoo is something I copied from a stack overflow tutorial while making pistonball around 2 years ago before pettingzoo even existed. Apparently, that article was sort of wrong? You're supposed to convert the image to a surface on loading in order to get a 2x-10x performance improvement (though pygame has no warnings for this and the documentation is unclear). This PR needs to be legitimately reviewed by someone else.

### DIFF
--- a/docs/_scripts/generate_gif_image.py
+++ b/docs/_scripts/generate_gif_image.py
@@ -3,7 +3,7 @@ import time
 import numpy as np
 import json
 
-from pettingzoo.test.all_modules import all_environments
+from test.all_modules import all_environments
 from pettingzoo.classic import gin_rummy_v0
 from PIL import Image
 import os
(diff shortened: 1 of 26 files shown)


benblack769:
I'll generate gifs of all the environments and make sure they look OK.

benblack769:
So `.convert()` is the wrong tool because it requires the display to be initialized.

benblack769:
I fixed prospectors performance issues with respect to this, so I'll see what solution I used there.

benblack769:
Significantly increase stepping performance in 

* prison (4x)
* KAZ (2x)

Increases rendering performance in all envs except texas holdem, due to when the images are loaded in the environment.

benblack769:
To validate correct display:

classic_texas_holdem_v3
classic_texas_holdem_no_limit_v3
butterfly_pistonball_v4
butterfly_cooperative_pong_v3
butterfly_knights_archers_zombies_v7
butterfly_prison_v3
classic_go_v4
classic_rps_v2
classic_connect_four_v3

Prospector is too big to upload to github, but it looks good as well.

PR #459, opened 2021-08-23
reverted black and isort

benblack769:
I found that this commit dramatically lowers the quality of the code, and makes certain sections very hard to read. Computers don't always write better code than humans, sometimes they make really poor choices as far as readability goes.


pettingzoo/test/api_test.py
```diff
@@ -26,75 +23,39 @@ def test_observation(observation, observation_0):
...
-        ):
-            warnings.warn(
-                "The class of observation spaces is different between two agents"
-            )
-        if not isinstance(
-            env.action_spaces[agent], env.action_spaces[agent_0].__class__
-        ):
+        assert isinstance(env.observation_spaces[agent], gym.spaces.Space), "Observation space for each agent must extend gym.spaces.Space"
+        assert isinstance(env.action_spaces[agent], gym.spaces.Space), "Agent space for each agent must extend gym.spaces.Space"
+        if not (isinstance(env.observation_spaces[agent], gym.spaces.Box) or isinstance(env.observation_spaces[agent], gym.spaces.Discrete)):
```
cclauss, line 48: The second parameter of `isinstance()` can be a tuple.
* https://docs.python.org/3/library/functions.html#isinstance
```suggestion
        if not isinstance(env.observation_spaces[agent], (gym.spaces.Box, gym.spaces.Discrete)):
```

pettingzoo/test/api_test.py
```diff
@@ -26,75 +23,39 @@ def test_observation(observation, observation_0):
...
+            warnings.warn("Observation space for each agent probably should be gym.spaces.box or gym.spaces.discrete")
```
cclauss, line 49:
```suggestion
            warnings.warn(
                "Observation space for each agent probably should be gym.spaces.box
                or gym.spaces.discrete"
            )
```

pettingzoo/test/api_test.py
```diff
@@ -26,75 +23,39 @@ def test_observation(observation, observation_0):
...
+        if not (isinstance(env.action_spaces[agent], gym.spaces.Box) or isinstance(env.action_spaces[agent], gym.spaces.Discrete)):
```
cclauss, line 50: `isinstance()` accepts a tuple.

pettingzoo/test/api_test.py
```diff
@@ -26,75 +23,39 @@ def test_observation(observation, observation_0):
...
+            warnings.warn("Action space for each agent probably should be gym.spaces.box or gym.spaces.discrete")
+        if (not isinstance(agent, str)) and agent != 'env':
```
cclauss, line 52: For what value of `agent` will this compound conditional ever be `True`?

If `agent` is not a string, then how could it possibly be equal to a particular string?!?

Also, this compound conditional and the one that follows both start with `if not isinstance(agent, str)` so that question should be asked just once and the other conditionals should be embedded in a block underneath it.

cclauss (commented):
Perhaps [`black --line-length 120 .`](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html?highlight=line-length#line-length) would be more to your style.  A few notes on the code block...

benblack769:
@cclauss So all of these are valid points. And if this was a regular PR to add some feature, I would totally make all these changes. However, there is probably 1000 lines of low quality, overly long lines throughout the codebase that is sort of like this. None of which would have made it through real code review, but we didn't really have resources to do proper code review at the time. And we certainly don't have time to rewrite it all by hand now. So the question currently at hand is if the code can be improved by using tools to rewrite the code. Some of the tools you have applied to this codebase have certainly improved the state of things, I am just thinking that `black` did not really increase the quality of the code.

cclauss:
Does `black --line-lenght 120 .` #460 improve things?

benblack769:
I looked at it and I also played around with a few other line lengths, including 100, 140, and 300, and while there are places where the code gets improved due to these changes, there are always places where the code gets butchered, one way or another.

RyanNavillus:
I agree that the way black formats parentheses is pretty awful in a lot of cases, but it also fixes a lot of random code inconsistencies that I'm a fan of (spacing out operations and parameters, and comment formatting), and there are a few cases where I think the extra spacing for the dictionary/parentheses is actually a good thing. From briefly looking through the example that @cclauss shared, I think that 120 is a good middle ground. I think it's better than completely scrapping linting altogether.

cclauss:
Perhaps the way to go is to put perfectly formatted code between [`# fmt: off` and `# fmt: on`](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html) so that psf/black will ignore it.

benblack769:
I'm certainly biased because I have a PR right now that got kind of screwed by these changes, and if this isn't removed from the git history, I'll basically have to rewrite the whole thing, which kind of sucks.

@cclauss 
Can I have a list of commands which you ran to generate this commit? If so, then I think I can do some fancy git tricks locally to get my changes merged in without having to rewrite all that code. That would make my life a lot easier.

cclauss:
```
python3 -m pip install black
black .
```

PR #463, opened 2021-08-23
Website update

Kyle-Sang:


benblack769:
I noticed this new dependency in the gemfile:  `webrick 1.7.0`

Does github pages support this? If it does not, then we can't use it.

cclauss:
Lots of conflicts.

docs/_data/img.yml
```diff
@@ -0,0 +1,3 @@
+nav:
+  arrow: /assets/img/expand_arrow.svg
+  menu: /assets/img/menu_icon.svg
```
cclauss, line 3: The ⛔ above is because GitHub expects that each file ends with one and only one carriage return.

docs/_data/navigation.yml
```diff
@@ -1,12 +1,300 @@
...
+      - title: "Wrappers"
+        url: "/environment_creation#wrappers"
+
+      - title: "Developer Utils"
+        url: "/environment_creation#developer-utils"
 
-  - title: "Dev Docs"
-    url: "/dev_docs"
+      - title: "Tests"
+        url: "/environment_creation#tests"
```
cclauss, line 300: ⛔

cclauss (approved):

PR #465, opened 2021-08-25
Migrate to new AutoROM release

JesseFarebro:
AutoROM now installs without the game name as the path. Furthermore, when installing to a user-specified directory `ROM/` isn't automatically appended.


JesseFarebro:
CI passes. AutoROM is backwards compatible upon release.

PR #472, opened 2021-08-28
auto deprication enabled

benblack769:


benblack769:
Ok, I think this should be good to merge now.

Issue #488, opened 2021-09-22
MPE change color during step


nicofirst1:
I'm using the MPE env for a navigation task.
I have a set of custom Landmarks class that need to be visited every N steps. At each step I increase the amount of time the landmark has not been visited with a custom attribute.
On top of that i would like to modify the colors of the landmark in order to reflect the time passed without visit.

My custom class is the following:
```

class TimerLandmark(Entity):
    """
    Timer landmark class.
    Each landmark increases its timer by the 'increase' param per step.
    This timer is proportional (timer* penalty) to the penalty agents get at each turn.
    The timer resets when an agent lands upon it and the timer starts from zero.

    So the longer a landmark stays untouched the worse the penalty gets.
    """

    colors = list(Color("green").range_to(Color("red"), 100))

    def __init__(self, increase=0.1):
        super().__init__()
        self.timer = 0
        self.increase = increase
        self.counter = 0

    def reset(self, world, np_random):
        self.timer = 0
        self.color = np.array([0, 1, 0])
        self.state.p_pos = np_random.uniform(-0.9, +0.9, world.dim_p)
        self.state.p_vel = np.zeros(world.dim_p)

    def step(self):
    """ Where the color changes"""
        self.counter += 1
        self.timer = self.increase * self.counter
        c = self.colors[self.counter]
        c = c.get_hsl()
        self.color = np.array(c)


    def reset_timer(self):
        self.timer = 0
```

And my step function in a `SimpleEnv` class is
```

    def step(self, action):
        super(raw_env, self).step(action)

        for landmark in self.world.landmarks:

            visited = False

            # check if one agent has visited the landmark
            for agent in self.world.agents:
                if self.is_collision(agent, landmark):
                    visited = True
                    break

            # if visited reset else increase
            if visited:
                landmark.reset_timer()
            else:
                landmark.step()
```

Since the colors dont change i looked around in the render and saw that the actual code where the color is instantiated is called just once at the start of the episode. Is there something i can do to dynamically change the colors?

benblack769:
I don't really know of a particular way to do this. The only advice I can gives is to try to figure out how to change the code to get the color every time render is being called, then submit a PR

benblack769:
Here is where I would start:

1. Don't add your landmark's geom to the viewer here: https://github.com/PettingZoo-Team/PettingZoo/blob/master/pettingzoo/mpe/_mpe_utils/simple_env.py#L222
[...]

PR #495, opened 2021-10-04
Fixed issue #488

nicofirst1:
Add class type for continuous color rendering as mentioned in issue [488](https://github.com/PettingZoo-Team/PettingZoo/issues/488)


jkterry1:
So after some discussion with Ben, I think it makes more sense for your use cases to remain in a fork and not in PettingZoo proper, but thank you :)

PR #504, opened 2021-10-12
updated no limit holdem to v6 (rlcard 1.0.5)

SonSang:
Updated no limit texas holdem to v6.


benblack769:
@jkterry1 Look good to me

PR #517, opened 2021-10-18
added render tests back to CI

benblack769:


benblack769:
@jkterry1 @elbamos Tests are passing here. I think #578 can be closed, and this merge.

PR #539, opened 2021-11-11
Parallel API Test Improvements

jsuarez5341:
Replaces several of the Parallel API test assertions with warnings. Ideally, API-conformant envs should pass with no warnings, but this is a decent first step. In particular:

- Removes a breaking assumption that possible_agents exists in all environments
- Assert -> warning: Environment is not done but contains no agents
- Assert -> warning: Environment contains no agents and not all possible_agents are done
- Assert -> warning: Agents will never respawn after marked done

Also fixed a typo in the contributor guidelines and added some details on setup. Passes linter and all tests except those dependent upon Atari ROM -- I could not figure out the correct install path for AutoROM

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,11 +17,11 @@ If you're thinking about creating submitting a new environment, please contact u
 Contributing code is done through standard github methods:
 
 1. Fork this repo
-1. Commit your code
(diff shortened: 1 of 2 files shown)


jsuarez5341:
@benblack769 From a conversation with @jkterry1 earlier today on limitations in the parallel API. Warnings are messy but a decent first-step. Both the main API and parallel API tests currently produce warnings for conformant environments

PR #540, opened 2021-11-13
Added comments to env files

Kyle-Sang:
first draft


update_docs.py
```diff
@@ -0,0 +1,229 @@
...
+    'leduc_holdem' : { 'markdown': 'leduc_holdem.md', 'code' : 'pettingzoo/classic/rlcard_envs/leduc_holdem.py', 'environment' : (leduc_holdem_v4.env())},
+    'mahjong' : { 'markdown': 'mahjong.md', 'code' : 'pettingzoo/classic/rlcard_envs/mahjong.py', 'environment' : (mahjong_v4.env())},
+    'rps' : { 'markdown': 'rps.md', 'code' : 'pettingzoo/classic/rps/rps.py', 'environment' : (rps_v2.env())},
+    'texas_holdem' : { 'markdown': 'texas_holdem.md', 'code' : 'pettingzoo/classic/rlcard_envs/texas_holdem.py', 'environment' : (texas_holdem_v4.env())},
+    'texas_holdem_no_limit' : { 'markdown': 'texas_holdem_no_limit.md', 'code' : 'pettingzoo/classic/rlcard_envs/texas_holdem_no_limit.py', 'environment' : (texas_holdem_no_limit_v6.env())},
+    'tictactoe' : { 'markdown': 'tictactoe.md', 'code' : 'pettingzoo/classic/tictactoe/tictactoe.py', 'environment' : (tictactoe_v3.env())},
+    'uno' : { 'markdown': 'uno.md', 'code' : 'pettingzoo/classic/rlcard_envs/uno.py', 'environment' : (uno_v4.env())},
+    'adversarial_pursuit' : { 'markdown': 'adversarial_pursuit.md', 'code' : 'pettingzoo/magent/adversarial_pursuit_v3.py', 'environment' : (adversarial_pursuit_v3.make_env())},
```
benblack769, line 141: This call to make_env is giving me an error when I try to run it. I assume this should just be `.env()`?

update_docs.py
```diff
@@ -0,0 +1,229 @@
...
+    'simple_tag' : { 'markdown': 'simple_tag.md', 'code' : 'pettingzoo/mpe/simple_tag_v2.py', 'environment' : (simple_tag_v2.make_env())},
+    'simple' : { 'markdown': 'simple.md', 'code' : 'pettingzoo/mpe/simple_v2.py', 'environment' : (simple_v2.make_env())},
+    'simple_world_comm' : { 'markdown': 'simple_world_comm.md', 'code' : 'pettingzoo/mpe/simple_world_comm_v2.py', 'environment' : (simple_world_comm_v2.make_env())},
+    'multiwalker' : { 'markdown': 'multiwalker.md', 'code' : 'pettingzoo/sisl/multiwalker/multiwalker.py', 'environment' : (multiwalker_v7.env())},
+    'pursuit' : { 'markdown': 'pursuit.md', 'code' : 'pettingzoo/sisl/pursuit/pursuit.py', 'environment' : (pursuit_v4.env())},
+    'waterworld' : { 'markdown': 'waterworld.md', 'code' : 'pettingzoo/sisl/waterworld/waterworld.py', 'environment' : (waterworld_v3.env())}
+}
+for file in files:
+
+    file_content = [ line for line in open(files[file]['markdown']) ]
```
benblack769, line 162: Why are we reading the markdown file? Shouldn't we be trying to write the markdown file with only information provided by the environment values and comments?

benblack769:
In general, I think this is a good direction to go and this change is definitely heading there. I could not get it to locally, getting errors about "markdown file not found" and "make_env() missing 1 required positional argument: 'raw_env'", so it is hard to review in detail, but some high level design comments are:

1. My biggest issue from a design perspective is that the markdown files should only be written to, not read. 
2. In addition, the files which are not maintained by hand, but are instead generated, should be clearly marked as such, with clear comments at the top of the file, or a different file naming scheme.

jkterry1:
Closing in favor of sander's work

PR #557, opened 2021-11-27
Update rllib_pistonball.py

nubonics:
Added ray's shutdown to prevent leftover ray instances
commented out local_ratio, as it errors otherwise
formatted params of the parallel env to be more readable

### DIFF
--- a/tutorials/rllib_pistonball.py
+++ b/tutorials/rllib_pistonball.py
@@ -71,7 +71,7 @@ def env_creator(args):
 
 
 if __name__ == "__main__":
-    shutdown()
+    # shutdown()
     
     env_name = "pistonball_v4"
 
--- a/tutorials/rllib_pistonball.py
+++ b/tutorials/rllib_pistonball.py
@@ -1,7 +1,6 @@
 from ray import tune
 from ray.rllib.models import ModelCatalog
 from ray.tune.registry import register_env
-# from ray.rllib.utils import try_import_tf
 from ray.rllib.env.wrappers.pettingzoo_env import ParallelPettingZooEnv
 from pettingzoo.butterfly import pistonball_v4
 import supersuit as ss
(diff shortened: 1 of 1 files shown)


tutorials/rllib_pistonball.py
```diff
@@ -60,6 +71,8 @@ def env_creator(args):
 
 
 if __name__ == "__main__":
+    shutdown()
```
benblack769, line 74: It seems odd to put this shutdown at the beginning of the script. Shouldn't it come at the end?

benblack769:
Also, if you could combine this change with your other change to rllib_pistonball.py into the same PR, and close the other PR, it would make it easier to review/test these fixes.

Thanks so much for fixing this very broken example.

tutorials/rllib_pistonball.py
nubonics: indeed, but this shuts any instances of ray that were not closed properly prior to running this script

nubonics: probably isnt required though, however, i commented it out with my last commit, just in-case someone else happens to run into the same problem i was

nubonics:
> Also, if you could combine this change with your other change to rllib_pistonball.py into the same PR, and close the other PR, it would make it easier to review/test these fixes.
> 
> Thanks so much for fixing this very broken example.

The two pull requests have been merged.

No! Thank you for contributing to such an amazing project!

tutorials/rllib_pistonball.py
benblack769: Interesting. Makes sense that this would be helpful, we have had problems with zombie ray processes before.

benblack769:
Thanks for combining these PRs. I am testing on an RTX 2060, and I am getting out of memory errors, which isn't ideal. I'm trying to find a set of hyperparameters which can be run on a consumer GPU like this.

benblack769:
Made a few minor changes.

PR #559, opened 2021-11-30
Turn based wrapper

RyanNavillus:
This probably needs a thorough review by @benblack769 but it passes the (modified) parallel API test. 

The reason for creating the turn_based_to_parallel wrapper is because some applications require the parallel API for simultaneous games, and would preferably also be able to use the same API for classic games. However, in a two-player game, the current API gives the same observation to player1 and player2 when accepting actions, which is unfair to player2 who should receive an updated observation after player1 takes an action.

The turn-based version of the parallel API still accepts actions for each player, but only plays the action for the active player. It also adds the "active_agent" to each agent's info dict.

I also made modifications to the parallel API to allow it to safely sample legal actions from classic games, and added a more descriptive warning for an issue I ran into.

### DIFF
--- a/pettingzoo/utils/conversions.py
+++ b/pettingzoo/utils/conversions.py
@@ -258,6 +258,8 @@ class turn_based_to_parallel_wrapper(to_parallel_wrapper):
             rewards = defaultdict(int)
             dones = dict(**self.aec_env.dones)
             infos = dict(**self.aec_env.infos)
+            for agent in self.aec_env.agents:
+                infos[agent]["active_agent"] = self.aec_env.agent_selection
             observations = {agent: self.aec_env.observe(agent) for agent in self.aec_env.agents}
             return observations, rewards, dones, infos
         self.aec_env.step(actions[self.aec_env.agent_selection])
@@ -282,5 +284,7 @@ class turn_based_to_parallel_wrapper(to_parallel_wrapper):
             else:
                 break
 
+        for agent in self.aec_env.agents:
+            infos[agent]["active_agent"] = self.aec_env.agent_selection
         self.agents = self.aec_env.agents
         return observations, rewards, dones, infos
--- a/pettingzoo/utils/conversions.py
+++ b/pettingzoo/utils/conversions.py
@@ -32,6 +32,14 @@ def from_parallel(par_env):
         return ordered_env
 
 
+def turn_based_to_parallel(aec_env):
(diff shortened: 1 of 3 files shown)


jkterry1:
Once ben reviews this I'd like joseph saurez to take a look too before we merge

RyanNavillus:
Also, I'm pretty confident that the observations and info are correct based on manually inspecting those values, but I haven't tested it with environments that return rewards at every timestep, or weird ones like uno.

pettingzoo/utils/conversions.py
```diff
@@ -236,3 +244,43 @@ def close(self):
...
+
+class turn_based_to_parallel_wrapper(to_parallel_wrapper):
+    def reset(self):
+        self.aec_env.reset()
+        self.agents = self.aec_env.agents[:]
+        observations = {agent: self.aec_env.observe(agent) for agent in self.aec_env.agents if not self.aec_env.dones[agent]}
+        return observations
+
+    def step(self, actions):
+        if actions == {}:
```
benblack769, line 257: Why is this case needed? It shouldn't be?

pettingzoo/utils/conversions.py
```diff
@@ -236,3 +244,43 @@ def close(self):
...
+        rewards = dict(**self.aec_env.rewards)
+        dones = dict(**self.aec_env.dones)
+        infos = dict(**self.aec_env.infos)
+        observations = {agent: self.aec_env.observe(agent) for agent in self.aec_env.agents}
+
+        while self.aec_env.agents:
+            obs, rew, done, info = self.aec_env.last()
+            agent = self.aec_env.agent_selection
+            observations[agent] = obs
+            rewards[agent] = rew
```
benblack769, line 275: Very subtle, but this is actually incorrect. This will set the cumulative reward, but the agent already got its reward from other player's turns. The reward scheme needs to be consistent.

benblack769:
Ok, I made a few fixes. Should be somewhat easier to understand now.

benblack769:
@RyanNavillus Can you take a look at this and check that it is doing what you want it to do?

benblack769:
And that the fixes I made make sense?

RyanNavillus:
Thanks for making these changes. Yes, they make sense. I'm trying to use this now, that should be a good way of testing it. Let's leave this open for another few days in case I need to make some changes.

RyanNavillus:
This should be ready to merge now, it seems to work in my use case.

PR #569, opened 2021-12-04
Added rgb_array to capture_stdout

Rohan138:
This was causing the MCMES SSD errors.

### DIFF
--- a/pettingzoo/utils/wrappers/capture_stdout.py
+++ b/pettingzoo/utils/wrappers/capture_stdout.py
@@ -17,6 +17,8 @@ class CaptureStdoutWrapper(BaseWrapper):
 
                 val = stdout.getvalue()
             return val
+        elif mode == "rgb_array":
+            return super().render()
 
     def __str__(self):
         return str(self.env)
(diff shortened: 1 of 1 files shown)


benblack769:
So generally speaking this wrapper is only meant to wrap environments which print text to the terminal, like the pettingzoo chess environment does. What environment did it cause an issue with in this case?

Rohan138:
Ah, that makes sense. I was getting the error with the SSD environments. However, the docs on the (PettingZoo environment creation)[https://www.pettingzoo.ml/environment_creation#example-custom-environment] page seem a little misleading in recommending them for all environments, or maybe that's just me misreading them.

Rohan138:
@benblack769 do you want to merge this or close it? I'm fine either way. It might help if there's a weird environment that both renders and prints to console.

benblack769:
That makes sense. That documentation is certainly misleading.

benblack769:
I think this can be merged anyways, but its good to know where things went wrong.

benblack769:
So a simpler, probably better logic for this change would simply be:

```
        if mode == "ansi":
            with capture_stdout() as stdout:

                super().render("human")

                val = stdout.getvalue()
            return val
        else:
            return super().render(mode)
```

PR #575, opened 2021-12-09
Force updated supersuit version

andrewtanJS:
Update supersuit version to support `pettingzoo_env_to_vec_env_v1` in supersuit. With the conflicting `gym` requirements, pip will only install a compatible version of supersuit (which is missing the mentioned updated method).


jkterry1:
(In case you didn't see, tests are failing)

PR #577, opened 2021-12-10
Supress pygame import message

andrewtanJS:
This suppresses the pygame support prompt which occurs anytime we import it.

### DIFF
--- a/pettingzoo/__init__.py
+++ b/pettingzoo/__init__.py
@@ -1,3 +1,4 @@
+import os
 import sys
 
 from pettingzoo.utils import AECEnv, ParallelEnv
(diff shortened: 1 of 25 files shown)


benblack769:
As far as I can tell, simply putting:

```
import os

os.environ['PYGAME_HIDE_SUPPORT_PROMPT'] = 'hide'
```

At the top of the `pettingzoo/__init__.py` file fixes all the errors with 2 lines of code. If this works, then this will be greatly preferable.

andrewtanJS:
Ah yes, this PR does add the lines you mentioned to `pettingzoo/__init__.py`. The other changes simply remove all instances of the same line in other files to prevent redundancy.

benblack769:
Ah. I didn't see that. Cool.

PR #582, opened 2021-12-12
added is_paralleizable metadata

benblack769:


jkterry1:
Shouldn't you list the false states for this so that people can write reasonable if statements...?

benblack769:
```
if env.metadata.get("is_parallelizable", False):
```
Isn't a good enough condition? I don't feel super great forcing every single user to include crazy amounts of metadata.

PR #584, opened 2021-12-14
Added Source Code Checking with Pyright

Ryan-Amaral:
Added pyright to the CI/Actions testing process just like in OpenAI Gym. Initially I just set it to use the same configuration as in OpenAI Gym. This includes type checking, unused import checking, and a few other things. 

In the future, files should be added to the list of files to check (`include` in `pyproject.toml`) once updated, it currently only checks the file `base_atari_env.py`. Though maybe it should just check all python files in the project (except for a few exceptions)?

For now, the pyright tests are set so that on failure the testing process will continue, this is because there are a lot of checks currently failing (lack of type declarations, potential unbound variables, etc.). Perhaps in the future once issues are fixed, the `continue-on-error` flag could be removed, but for now it is needed to not fail the testing pipeline.

In VS code with the pylance extension the potential issues are highlighted, so that's another nice part of the change.


jkterry1:
why does it say "uses: jakebailey/pyright-action@v1"

benblack769:
I think the plan is to only check for type hints in externally accessible code. This would include:


1. The base AECEnv and ParallelEnv environment definition
2. The api tests
3. The utilities

So base_atari_env would not be tested.

Ryan-Amaral:
@jkterry1 That's the same way it works in the OpenAI Gym repo, its just a github action for pyright checking.

@benblack769 OK, I was unaware of that. I will eventually update this PR to remove the changes/checking on `base_atari_env` and add a couple of example changes into AECEnv.

jkterry1:
@Ryan-Amaral hey, just following up on this

Ryan-Amaral:
> @Ryan-Amaral hey, just following up on this

Hi, sorry haven't really touched it the last week or so. I'm thinking I'll get it done by early next week.

Ryan-Amaral:
Sorry it took so long, I changed it so that it checks the AEC and Parallel env base classes. I've just added simple changes to that file, namely just stating that the str functions return str. 

There is still much more to be done in that file though to have proper type checking for everything. So as mentioned before I set it so that the CI still passes even if the pyright checks fail.

jjshoots:
Hi @Ryan-Amaral, sorry for bringing up something from so long ago, but what's the status of this PR right now?

Issue #581, opened 2021-12-12
Move seed into an argument in reset like Gym will in the next release


jkterry1:

benblack769:
@jkterry1 Do you have a plan to move gym-ale to this new reset API? I don't see it here: https://github.com/mgbellemare/Arcade-Learning-Environment/blob/master/src/gym/envs/atari/environment.py#L231

jkterry1:
Jesse has to update Atari environments to this himself and hasn't. I'm currently trying to resolve the situation.

PR #588, opened 2021-12-17
Add seed parameter to reset() [DO NOT MERGE YET]

elbamos:
And call self.seed(seed) in non-deterministic environments if the seed parameter is set. 

Addresses #581 .

### DIFF
--- a/.github/workflows/linux-test.yml
+++ b/.github/workflows/linux-test.yml
@@ -43,10 +43,3 @@ jobs:
         pytest ./test/all_parameter_combs.py
         pytest ./test/variable_env_test.py
         pytest ./test/doc_examples_test.py
-    - name: 13 lines test
-      run: |
-        if [[ $(python -c 'import sys; print(sys.version_info[1])') -lt 10 ]]
-        then
-            pip install stable_baselines3 supersuit==3.3.2
-            sed s/total_timesteps=2000000/total_timesteps=2000/g tutorials/13_lines.py | sed s/n_steps=256/n_steps=16/g | xvfb-run -s "-screen 0 1024x768x24" python -
-        fi
--- a/docs/_includes/code/aec_rps.py
+++ b/docs/_includes/code/aec_rps.py
@@ -107,7 +107,7 @@ class raw_env(AECEnv):
         '''
         pass
 
-    def reset(self):
+    def reset(self, seed=None):
         '''
         Reset needs to initialize the following attributes
         - agents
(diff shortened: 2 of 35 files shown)


benblack769:
The documentation and api tests also need to be changed to completely resolve #581.

benblack769:
Let me know if you need any help with this.

elbamos:
I can address the issue with the wrappers and the base method; can you take a look at what’s required in terms of updated documentation and api tests?

benblack769:
I can take a look, sure. Note that once you change the wrappers to not forward the seed() method, some of the tests will fail.

elbamos:
My most recent push propogates the seed logic to the wrappers and fixes some of the seed api tests.  Most tests pass -- however the `atari` and `magent` tests fail on a type error, that `reset()` doesn't have a seed parameter. 

I'm not sure why that is - the environments for atari and magent are appropriately catching the call to `reset()` and calling `seed()` on themselves rather than passing it to the underlying external environment. 

Any hints?

elbamos:
I've fixed the wrappers, updated the seed tests, and added deprecation warnings to the base seed() method. All tests should pass.  I'm not sure what needs to be done with documentation?

elbamos:
It looks that this PR is now clearing all tests internal to PettingZoo. It's getting stuck on the 13-lines test, because that depends on wrappers imported from supersuit that haven't been updated for the changed seed api. 

How do we move forward?

jkterry1:
@elbamos could you please create a PR to supersuit to add support for this?

elbamos:
I made an attempt: https://github.com/Farama-Foundation/SuperSuit/pull/123 but trying to update both repos at the same time really complicates testing and validation.  

Considering that SuperSuit depends on PettingZoo, I suggest taking out of PettingZoo any dependencies on SuperSuit, which I think are limited to the use of SuperSuit wrappers in the 13-lines test.

benblack769:
@elbamos Yeah, on reflection, I think adding the 13 lines test was a mistake (my mistake, to be specific). Feel free to remove it.

elbamos:
This should be good to merge now.

jkterry1:
I just want to note here that this should not be merged until the version of Gym with this fix is released. This release will most likely take another month (or longer if more goes wrong). Given the large number of environment fixes coming in right now, we likely will want to cut a PettingZoo release before then.

benblack769:
@elbamos Just made a few changes to the tests and base environment stuff here, generally looked good, all the environments were updated and all that.

benblack769:
Actually, @elbamos I found one issue with the implementation that I fixed. 

the logic:

```
if seed:
     self.seed(seed)
```

doesn't work if the `seed=0`. So I had to change them to 

```
if seed is not None:
     self.seed(seed)
```

jjshoots:
Closing because #669 handles this

PR #595, opened 2021-12-22
Multiwalker fixes

Rohan138:
Several documentation fixes, plus fixed the condition for walkers falling off the cliff.


benblack769:
The changes to the code look good to me. 

I'm not sure where the 130 number comes from. I would encourage making this `self.terrain_length` an argument https://github.com/Farama-Foundation/PettingZoo/blob/1977fc6021421e5473eba44f4acf5aea1dfeb828/pettingzoo/sisl/multiwalker/multiwalker_base.py#L336 because right now it feels 

1. A bit short.
2. Varying depending on the number of walkers is very weird. 
3. Then the maximum reward can depend on the value of this argument.

Once that is encoded as an argument, you could even add to the documentation that the maximum reward each agent can get is based on the terrain length.

Rohan138:
Yeah, that makes sense, and the variable terrain length and proportionate reward scaling should probably be upstreamed into BipedalWalker in Gym as well.

carlosluis:
@Rohan138 regarding the upstream changes to gym, I have a couple of questions:

- When you say "variable terrain length", do you mean to have it as an argument as @benblack769 suggested for the multi walker? If so, is that something that makes sense in the single-agent case? I'd think that it would be make sense to have a fixed length instead.
- What do you mean by proportionate reward scaling? I don't see any particular line of code that scales rewards based on `terrain_length`. The total reward you can get in the environment is _automatically_ scaled by `terrain_length`, so there's nothing to be added to gym on that front, right?
- Aside from the above points, is there something else from your changes that need to be upstreamed into gym?

Issue #589, opened 2021-12-20
Seemingly buggy gifs in pistonball


jkterry1:
Both of these seem to raise meaningful concerns about accurate collision detection with the left wall:

![image](https://user-images.githubusercontent.com/19422700/146797834-861717f8-4a0e-43e7-88e6-e34751fb91a3.png)

![image](https://user-images.githubusercontent.com/19422700/146797851-e948a1c1-fee0-44b0-bed2-040b2c6c7df4.png)

jkterry1:
Note that these were all created with continuous=False

jkterry1:
![0_3c_97 40](https://user-images.githubusercontent.com/19422700/146884399-8af181a9-c022-45de-b66e-e1464eec4a61.gif)
![0_4a_95 79](https://user-images.githubusercontent.com/19422700/146884421-8bd22c70-a14e-41d9-8f2b-80182f2adc61.gif)

jkterry1:
These gifs more clearly show the bouncing behavior

PR #596, opened 2021-12-22
Fix for #589 where the ball was bouncing off the left wall

jjshoots:
Previous environment done check was done by checking the ball position in comparison with the left wall for every frame. This caused an aliasing issue where if at time t the ball didn't collide with the wall, at time t+0.5 the ball collides with the wall, and at time t+1 the ball is already away from the wall, the done was not triggered.

This PR fixes that issue by instead checking the position of the ball using a first order approximation using the ball velocity information. If next position is less than the wall position, done is triggered.


jkterry1:
@jjshoots could you please message me on discord? I was going to ask you to look at a few more things, but seem to have lost my mind and can't find you on discord

PR #597, opened 2021-12-22
Fix ball clipping through nose of cake paddle; rework ball motion and collision handling; bump cooperative_pong to v5

eczy:
The cake paddle collision detection function encloses horizontal and vertical collisions in the same if-else statement such that some vertical collisions went undetected if the ball had nonzero horizontal velocity. This PR fixes this issue and bumps cooperative_pong to v5.


eczy:
I've added another commit that reworks the ball movement mechanic and collision handling on both types of paddles. This fixes an issue where the ball appears to bounce despite not coming into contact with the cake paddle.

eczy:
Note: this last commit fixes `cake_paddle=False`, which was broken due to the cake paddle and normal paddle having different `update(...)` function signatures. I used backticks in the commit message on accident, so the message did not display properly 😬.

pettingzoo/butterfly/cooperative_pong/cake_paddle.py
```diff
@@ -55,6 +48,24 @@ def update(self, area, action, b_rect):
...
+        # handle collision from top
+        if b_rect.bottom > rect.top and b_rect.top - dy < rect.top and b_speed[1] > 0:
+            b_rect.bottom = rect.top
+            if b_speed[1] > 0:
+                b_speed[1] *= -1
+        # handle collision from bottom
+        elif b_rect.top < rect.bottom and b_rect.bottom - dy > rect.bottom and b_speed[1] < 0:
+            b_rect.top = rect.bottom
+            if b_speed[1] < 0:
+                b_speed[1] *= -1
```
jjshoots, line 61: Is there any reason why `b_speed[1] > 0` is needed? Since `dx` and `dy` come from `b_speed` within `ball.py` itself, I don't see why this check is needed. I think assertations would make more sense if `np.sign(b_speed[1]) == np.sign(dy)` is needed to be asserted.

It's also rather curious that there are redundant values representing speeds, ie: `dx`, `dy`, `b_speed[0]`, `b_speed[1]`. Looking through the code, I don't see why they can't be removed and just `b_speed` used for everything. It seems that at some point or another, an effort was made to let this env be run with variable frame times, but that effort has ended. Eitherways, I think the use of `dx` and `dy` can be removed.

pettingzoo/butterfly/cooperative_pong/paddle.py
```diff
@@ -0,0 +1,67 @@
...
+        if b_rect.bottom > self.rect.top and b_rect.top - dy < self.rect.top and b_speed[1] > 0:
+            b_rect.bottom = self.rect.top
+            if b_speed[1] > 0:
+                b_speed[1] *= -1
+        # handle collision from bottom
+        elif b_rect.top < self.rect.bottom and b_rect.bottom - dy > self.rect.bottom and b_speed[1] < 0:
+            b_rect.top = self.rect.bottom - 1
+            if b_speed[1] < 0:
+                b_speed[1] *= -1
+        return True, b_rect, b_speed
```
jjshoots, line 67: At the moment, I don't see any difference between the collision logic for `paddle.py` with that of `cake_paddle.py`'s individual cake tiers. For this reason, I think it would make more sense if:
1.  `_process_collision_with_rect` was moved into another file, maybe something like `collision_handler.py`. Consequently, this function should then also be able to handle collisions from the right (at the moment it only handles collisions from the left).
2. Cake paddle is updated to be a list of surfs instead of many individual surfs, and a foreach loop used to handle the collisions with the above function.
3. Make `paddle.py`'s `process_collision` use the aformentioned `_process_colllision_with_rect`.

pettingzoo/butterfly/cooperative_pong/ball.py
```diff
@@ -0,0 +1,65 @@
...
+        self.surf = pygame.Surface(dims)
+        self.rect = self.surf.get_rect()
+        self.speed_val = speed
+        self.speed = [int(self.speed_val * np.cos(np.pi / 4)), int(self.speed_val * np.sin(np.pi / 4))]
+        self.bounce_randomness = bounce_randomness
+        self.done = False
+        self.hit = False
+        self.randomizer = randomizer
+
+    def update2(self, area, p0, p1):
```
jjshoots, line 21: Is there any reason why this is named `update2` instead of `update`?

jjshoots (commented):
I like the direction that this PR is taking - modularizing each component of the env. But at the moment, the solution looks somewhat half-baked, kind of like a v4.5. This is by no means wrong, it's just that it should probably be polished a bit more before a merge. I've added some comments that I think will help take this to the next step, hope that these are helpful!

pettingzoo/butterfly/cooperative_pong/cooperative_pong.py
```diff
@@ -10,8 +10,10 @@
 from pettingzoo.utils.agent_selector import agent_selector
 from pettingzoo.utils.conversions import parallel_wrapper_fn
 
+from .ball import Ball
 from .cake_paddle import RENDER_RATIO, CakePaddle
```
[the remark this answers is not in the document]
eczy: Moving `RENDER_RATIO` to `cooperative_pong.py` causes a circular dependency issue, but I agree that this should be a parameter of the environment as a whole rather than being contained in the cake paddle class. I've moved it to `cooperative_pong.py`, and I've also moved the global variables to keyword arguments of the environment.

pettingzoo/butterfly/cooperative_pong/cake_paddle.py
eczy: If you are referring to the `b_speed[1] > 0` on line 53, this is needed to avoid incorrectly repositioning the ball in the rare case where the ball is moving perfectly horizontally. If you are referring to the `b_speed[1] > 0` on line 55, then this is indeed not needed due to the aforementioned check and can be removed.

Additionally, you are correct that `dx` and `dy` are not needed since they come directly from `b_speed`. I did not originally remove them to avoid modifying function signatures, but since this PR has grown to be more disruptive than originally intended, I will address this as well.

pettingzoo/butterfly/cooperative_pong/ball.py
eczy: I believe this is named `update2` instead of `update` since `Ball` inherits `pygame.sprite.Sprite` which already has an `update` function. It could probably use a more descriptive function name.

pettingzoo/butterfly/cooperative_pong/paddle.py
eczy: I agree that collision handling should probably be in its own module and that `Paddle` and `CakePaddle` should probably inherit from a common parent class. I'll address this in a future commit.

jkterry1:
@eczy could you please add FPS metadata to cooperative pong like this PR does? https://github.com/Farama-Foundation/PettingZoo/pull/601

There's so many changes in this PR that jat can't easily do it in his

eczy:
@jkterry1 I certainly can. Apologies for the scope creep on this PR. It's gotten much larger than I originally intended.

jjshoots:
@eczy with an addition of the FPS parameter, it's probably needed that all entity speeds depend on this FPS parameter, otherwise someone will inadvertently change the FPS and then wonder why isn't it doing anything. Would it be possible to implement such a change to this environment? Something like the below snipper ai stole from KAZ

```py
        self.knight_speed = KNIGHT_SPEED * 15. / FPS
        self.angle_rate = ANGLE_RATE * 15. / FPS

        if self.knight_speed % 1. != 0.:
            raise ValueError(f'FPS of {FPS} leads to decimal place value of {self.knight_speed} for knight_speed.')
        if self.angle_rate % 1. != 0.:
            raise ValueError(f'FPS of {FPS} leads to decimal place value of {self.angle_rate} for angle_rate.')

        self.knight_speed = int(self.knight_speed)
        self.angle_rate = int(self.angle_rate)

```

Essentially, we assume the environment has a base speed of 15 FPS, and then scale all speeds by the FPS parameter. Then assert that they all remain round numbers before casting them to ints so things move around consistently.

I believe this needs to be done for both paddles and the ball.

eczy:
@jjshoots As I understand it from my last conversation with @jkterry1 on the topic, I was asked to add the FPS parameter in Cooperative Pong solely for API compatibility purposes. There are already parameters for scaling the speed of both the ball and paddles, so for an environment this simple perhaps this is unnecessary? If this is something that we feel is necessary, however, I would be happy to implement it.

jjshoots:
@eczy Ok, in this case it's probably safe to merge and this FPS / speed coupling dealt with in another PR (possibly in #601).

@jkterry1 safe to merge after CI approval

Issue #590, opened 2021-12-20
FPS metadata


jkterry1:
Gym environments have the `video.frames_per_second` item in metadata. It would likely be a fairly good idea to add this to PettingZoo as well, remembering the rendering FPS for each PettingZoo environment is fairly messy and not precisely documented.

PR #601, opened 2022-01-03
Added FPS metadata to environments

jjshoots:
This PR addresses #590. It adds the ability to control the FPS of of each environment, as well as have it in the metadata. Note that for the most part, the FPS must be multiples of 15 for the environments to work without glitches. This change has not been added to cooperative pong yet because #597 is not yet merged.


pettingzoo/butterfly/knights_archers_zombies/src/weapons.py
```diff
@@ -8,17 +8,16 @@
 BLACK = (0, 0, 0)
 GRAY = (200, 200, 200)
 ARROW_SPEED = 45
-SPEED_VAL = 3
 ANGLE_RATE = 10
-KNIGHT_SPEED = 5
+SWORD_SPEED = 20
```
jkterry1, line 12: why is this change needed?

jjshoots: For some reason it wasn't used at all (actually a lot of defined variables aren't used at all, seems to be forgotten between updates), so I removed it. And then in line 63 of the old file, the speed was hard defined to be 20, which I changed to be a global variable to be consistent with the rest of the file.

jjshoots:
Accidentally commited to the wrong branch ^, fixed it now.

benblack769:
LGTM.

jkterry1:
@benblack769 there's been more changes; please review it now

pettingzoo/butterfly/knights_archers_zombies/src/players.py
```diff
@@ -88,31 +99,42 @@ def __init__(self, agent_name):
...
         self.action = -1
         self.score = 0
         self.is_archer = False
         self.is_knight = True
         self.agent_name = agent_name
 
+        self.knight_speed = KNIGHT_SPEED * 15. / FPS
+        self.angle_rate = ANGLE_RATE * 15. / FPS
+
+        if self.knight_speed % 1. != 0.:
```
benblack769, line 113: Just curious, why is it  problem if the knight speed is fractional?

pettingzoo/classic/backgammon/backgammon_env.py
```diff
@@ -20,9 +20,10 @@ def env(**kwargs):
 
 class raw_env(AECEnv):
     metadata = {
-        'render.modes': ['human'],
+        "render.modes": ["human"],
         "name": "backgammon_v3",
         "is_parallelizable": False,
+        "video.frames_per_second": 2,
```
benblack769, line 26: Why is it 2 frames per second? I don't recall any FPS style logic anywhere in these classic environments, unless I am forgetting something.

pettingzoo/magent/adversarial_pursuit_v3.py
```diff
@@ -71,7 +71,11 @@ def get_config(map_size, minimap_mode, tag_penalty):
 
 
 class _parallel_env(magent_parallel_env, EzPickle):
-    metadata = {'render.modes': ['human', 'rgb_array'], 'name': "adversarial_pursuit_v3"}
+    metadata = {
+        "render.modes": ["human", "rgb_array"],
+        'name': "adversarial_pursuit_v3",
+        "video.frames_per_second": 5,
```
benblack769, line 77: Same here. Is this a suggestion for how fast it should be rendered?

benblack769:
So I have a few small questions in comments above. 

The only big question I have is how to verify that these FPS changes are complete, that is, if someone changes the FPS, the environment's relative velocities, etc will mostly stay the same. As far as I can tell, the options are

1. Don't assume that the FPS variable is complete, that is, changing it could be a very risky thing to do. Put a comment saying as such by the declaration of these variables. 
2. Do some sort of testing with various FPS, perhaps producing videos to manually verify that the general behavior is the same.

benblack769:
Actually, if you have manually carefully verified that changing the FPS doesn't significantly alter the environments by simply running them and studying their behavior, then thats probably OK as well.

pettingzoo/butterfly/knights_archers_zombies/src/players.py
jjshoots: At first glance I do not think this will be an issue, but I kept the same convention to keep in line with the other butterfly envs where speeds are usually in terms of grid spacing per step. It's mostly to assert that I don't break anything else down the line.

pettingzoo/classic/backgammon/backgammon_env.py
jjshoots: Indeed, there is actually no need for an FPS in these environments (they are turn based afterall). But the FPS metadata is here just for the end user to use in case they intend to view things as a clip with the .render() function. I tested a few different FPSes and have also looked at the gifs in the Pettingzoo website, and eyeballed that 2 FPS seems to look decent.

pettingzoo/magent/adversarial_pursuit_v3.py
jjshoots: Yeap, this metadata thing is mostly a suggestion for how fast it should be rendered at.

jjshoots:
@benblack769 With the exception of MPE, Atari, and Butterfly, the FPS metadata is just a suggestion for how fast the environment should be rendered.

In the case of Butterfly, since we have explicit control over the environments, I've made sure that changing the FPS does not change the behaviour of the environment.

In the case of Atari, the environment actually depends on the Stella emulator, which by default runs things with an FPS of 60. I believe this is what OpenAI and DM uses for their implementations as well. In any case, changing the FPS of this environment cannot be easily done within the Pettingzoo environment.

In the case of MPE, this is actually [hardcoded](https://github.com/openai/multiagent-particle-envs/blob/master/multiagent/core.py#L94) to be 10 in the environment dependency, so the suggestion in the FPS metadata is for it to be rendered at 10 FPS (I just noticed I put 15 in there, the following commit fixes this).

benblack769:
Sounds good to me.

PR #602, opened 2022-01-05
Cleanup and streamlining of KAZ

jjshoots:
This PR is a rather comprehensive cleanup and streamlining of the KAZ repo, several existing hidden bugs are also cleaned up. Due to the heavy nature of refactoring that this has gone through, it is uncertain if new bugs have popped up, though every effort was made to assert that this is not the case.

Note: This PR depends on #601


jkterry1:
@jjshoots this bumps version right?

jjshoots:
@jkterry1 No it doesn't. It should right?

jkterry1:
Yes, please make it bump the environment version

benblack769:
@jjshoots  I'm trying to check out this repository locally so I can play around with it, but the repository seems to be broken. I can't clone it or fetch it, and Github can't even display the base link to the fork https://github.com/jjshoots/PettingZoo/ 

While this branch seems to be OK, it is making it hard to review. Can you figure out how to fix this so that review is easier?

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -153,209 +144,108 @@ def action_space(self, agent):
...
+                self.archer_list.remove(archer)
                 if archer.agent_name not in self.kill_list:
                     self.kill_list.append(archer.agent_name)
-        return zombie_list, archer_list, all_sprites, archer_killed
 
     # Zombie Kills the Sword
-    def zombie_sword(self, zombie_list, sword_list, all_sprites, score):
-        for sword in sword_list:
-            zombie_sword_list = pygame.sprite.spritecollide(sword, zombie_list, True)
+    def sword_hit(self):
```
benblack769, line 220: This doesn't make a lot of sense to me. After watching for awhile I think I might see what this does. The zombie hits the sword, then the sword disappears? Is that what is intended?

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -385,129 +275,124 @@ def observe(self, agent):
...
             self.kill_list = []
 
-            self._agent_selector.reinit(_live_agents)
+            self._agent_selector.reinit(self.agents)
+
         if len(self._agent_selector.agent_order):
             self.agent_selection = self._agent_selector.next()
-        self._cumulative_rewards[agent] = 0
+
+        self._cumulative_rewards[self.agent_selection] = 0
```
benblack769, line 390: This change breaks the reward. The previous agent's reward should get set to 0, the self.agent_selection is the next agent at this point in the function.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jjshoots: Yes, this is the intended behaviour, at least this was the behaviour in the previous version. 2 lines down, the zombie that is hit gets deleted too.

jjshoots:
@benblack769 Thanks for pointing out the broken fork. Tbh I have no idea why it's happening, 5/6 times when I load https://github.com/jjshoots/PettingZoo/ it fails, but every odd time it loads 'properly'.

It's a bit of a hacky solution, but for now would it suffice to checkout via the commit ID while I figure this out? Ie:
```
git fetch origin c3d649489f7128f72ff7d74dd7c0f4759bf65261
git checkout origin c3d649489f7128f72ff7d74dd7c0f4759bf65261
```

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
benblack769: Yeah, I see this. I don't think it is intended behavior. I don't see it in the documentation. Perhaps @jkterry1 can clear up what the intended behavior is?

benblack769:
@jjshoots That workaround seemed to work for me.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jkterry1: That's clearly not the intended behavior for the environment

Issue #574, opened 2021-12-08
Notes regarding from_parallel and other conversions


jkterry1:
Right now, pettingzoo has tools like:

```
from pettingzoo.utils.conversions import from_parallel
env = from_parallel(par_env)
```

from_parallel should be given a better name, e.g. parallel2aec, the names of conversions should be revisited, and these should be put into pettingzoo.wrappers when that merge happens

benblack769:
Resolved by #604

PR #604, opened 2022-01-05
Renames as per #574

jjshoots:
Rename from_parallel and to_parallel to parallel2aec and aec2parallel respectively. This addressed #574.


pettingzoo/utils/conversions.py
```diff
@@ -32,6 +32,22 @@ def from_parallel(par_env):
         return ordered_env
 
 
+def to_parallel(aec_env):
```
benblack769, line 35: Do we really want legacy support for this? Personally, I am OK with these functions raising an deprication error telling users to change the names of their functions. At least a very explicit and insistent user warning. @jkterry1 Thoughts?

jjshoots: It breaks some stuff with supersuit, the idea is to have this here first, then update the one on supersuit, and then remove it altogether.

benblack769: In the past, when we have things that break functionality in supersuit, we simply release the two packages together. However, if you do want to go with this 3 step process, these functions should at least give a deprication warning.

jjshoots: I will look into making a PR for supersuit in the near future then.

PR #611, opened 2022-01-27
Refactor Multiwalker to use Pygame

andrewtanJS:
This PR refactors the rendering function of the two environments to use pygame instead of pyglet. The refactored rendering is consistent with the previous pyglet rendering, with some changes to the `rgb_array` output of the `render` method.

Here is a side by side comparison of the two environments. Left is the old pyglet version, and right is the new pygame version
multiwalker1
multiwalker2
The current output of the render method when `mode="rgb_array"` is now of shape (400, 600, 3), which is consistent with the specified display dimensions. The old render method output had a shape of (800, 1200, 3). For simplicity sake, we use `np.transpose(np.array(pygame.surfarray.pixels3d(self.screen)), axes=(1, 0, 2))`, where previously the output was obtained from using the pyglet color buffer.

### DIFF
--- a/pettingzoo/sisl/multiwalker/multiwalker_base.py
+++ b/pettingzoo/sisl/multiwalker/multiwalker_base.py
@@ -3,6 +3,9 @@ import math
 
 import Box2D
 import numpy as np
+import pygame
+from pygame import gfxdraw
+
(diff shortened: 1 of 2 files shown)


jkterry1:
@andrewtanJS can you please fix setup.py?

trigaten (approved):
LGTM

Issue #499, opened 2021-10-05
Manual control bug


jkterry1:
from pettingzoo.butterfly import pistonball_v4
pistonball_v4.manual_control(ball_mass=.0001, ball_elasticity=50000)

Gives the error

```
Traceback (most recent call last):
  File "/Users/justinterry/Documents/random_demo.py", line 2, in <module>
    pistonball_v4.manual_control(ball_mass=.0001, ball_elasticity=50000)
  File "/usr/local/lib/python3.9/site-packages/pettingzoo/butterfly/pistonball/manual_control.py", line 53, in manual_control
    env.step(a)
  File "/usr/local/lib/python3.9/site-packages/pettingzoo/utils/wrappers/order_enforcing.py", line 68, in step
    super().step(action)
  File "/usr/local/lib/python3.9/site-packages/pettingzoo/utils/wrappers/base.py", line 68, in step
    self.env.step(action)
  File "/usr/local/lib/python3.9/site-packages/pettingzoo/utils/wrappers/clip_out_of_bounds.py", line 24, in step
    assert (
AssertionError: action should have shape (1,), has shape ()
```

rlew631:
Just tried recreating the error running PettingZoo 1.14.0 and got a deprecation error. Running _from pettingzoo.butterfly import pistonball_v5_ results in an import error.

I saw there was a __master_v1.11.2__ branch on the repo. Should I roll back to that and attempt to debug or focus on fixing the import on the current release and go from there?

jkterry1:
Hey,

v5 should work if you're using the newest version on pypi, I think master might be on v6 now. What import error are you getting?

rlew631:
So this is weird... I was getting
`Traceback (most recent call last):
  File "/Users/rlew/Documents/pettingzoo/demo.py", line 1, in <module>
    from pettingzoo.butterfly import pistonball_v5
  File "/usr/local/lib/python3.9/site-packages/pettingzoo/butterfly/__init__.py", line 5, in __getattr__
    return depricated_handler(env_name, __path__, __name__)
  File "/usr/local/lib/python3.9/site-packages/pettingzoo/utils/deprecated_module.py", line 42, in depricated_handler
    raise ImportError(f"cannot import name '{env_name}' from '{module_name}'")
ImportError: cannot import name 'pistonball_v5' from 'pettingzoo.butterfly'`
when trying to run a clean install of PettingZoo.

Changing my code to:
[...]

PR #614, opened 2022-01-28
fixed manual control issue

benblack769:


benblack769:
Addresses #499, specifically for pistonball

PR #618, opened 2022-01-28
Reopen Vector State KAZ

jjshoots:
This adds a vectorizable state to KAZ.

There are two options for vector states:

**Dynamic sizes**
Pass `vector_state='dynamic'` as a kwarg to the env. `env.state()` will now return an Nx4 ndarray, where:
The first 2 in N are archers.
The next 2 in N are knights.
Every vector after the first 4 are for zombies.
Each vector state is represented as `x, y, v1, v2` where `v1, v2` represent a unit vector representing the direction that the agent is facing.

**Constant sizes**
Pass `vector_state='constant'` as a kwarg to the env.
`env.state()` will now return an 120x4 ndarray, where:
The first 8 values represent the states of the archers.
The next 8 values represent the states of the knights.
Every other value represents states of the zombies, in multiples of 4.


pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -50,6 +50,8 @@ def __init__(
         pad_observation=True,
         line_death=False,
         max_cycles=900,
+        vector_state=None,
```
benblack769, line 53: Maybe this should be True/False?

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -261,41 +273,124 @@ def zombie_all_players(self, knight_list, archer_list, run):
...
+
+            # unit vector
+            unit_vec = np.array([0, -1])
+
+            # get rotated relative unit vectors
+            rel_ang = rot_mat @ unit_vec
+
+            # get relative positions scaled to screen size
+            rel_pos = all_pos - agent_pos
+            rel_pos = rel_pos / const.SCREEN_DIAG
```
benblack769, line 341: This magnitude normalization should also happen when someone calls `state()`

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -261,41 +273,124 @@ def zombie_all_players(self, knight_list, archer_list, run):
...
+
+            # give more emphasis to closer positions
+            rel_pos = np.sign(rel_pos) * (1 - abs(rel_pos))
```
benblack769, line 344: Not sure I like this strategy. In the current map, the difference between being inside a knights attack range and outside will be like 0.98 vs 0.96, which is pretty hard for an ML algorithm to split.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -261,41 +273,124 @@ def zombie_all_players(self, knight_list, archer_list, run):
...
+            # get rotation matrix
+            c, s = np.cos(rel_ang), np.sin(rel_ang)
+            rot_mat = np.array([[c, -s], [s, c]])
+            rot_mat = np.transpose(rot_mat, (2, 0, 1))
+
+            # unit vector
+            unit_vec = np.array([0, -1])
+
+            # get rotated relative unit vectors
+            rel_ang = rot_mat @ unit_vec
```
benblack769, line 337: Pretty sure that rel_ang will end up just being something vaugely equivalent to `(np.cos(rel_ang), np.sin(rel_ang))`, no need for all the rotation matrix stuff.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -261,41 +273,124 @@ def zombie_all_players(self, knight_list, archer_list, run):
...
+            all_pos = state[:, :2]
+            all_ang = state[:, 2:]
+
+            # compute the angles that everything is facing
+            all_ang = np.arctan2(all_ang[:, 0], all_ang[:, 1])
+            agent_ang = np.arctan2(agent_ang[:, 0], agent_ang[:, 1])
+
+            # get relative angles that everything is facing relative
+            # to the current agent
+            rel_ang = all_ang - agent_ang
```
benblack769, line 326: Unfortunately, I think there might be some problems if the angle is relative. In particular, its hard to determine whether you are facing the top or bottom of the screen, without comparing against the zombies, which feels a bit weird.

Its also weird to have the angle be relative and the position not be consistent with the angle.

I think it is fine to just leave the angle as absolute, and the position as relative. If you feel strongly that the angle should be relative, then probably some absolute angle should also be added somewhere to the observation.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -261,41 +273,124 @@ def zombie_all_players(self, knight_list, archer_list, run):
...
+
+            # get relative positions scaled to screen size
+            rel_pos = all_pos - agent_pos
+            rel_pos = rel_pos / const.SCREEN_DIAG
+
+            # give more emphasis to closer positions
+            rel_pos = np.sign(rel_pos) * (1 - abs(rel_pos))
+
+            # combine the positions and angles
+            state = np.concatenate([rel_pos, rel_ang], axis=-1)
```
benblack769, line 354: Just had another thought abotu the relative position. While the relative position is nice for combat, it is not so easy to determine the absolute position of the zombies. The absolute position of zombies is important because if they get too close to an absolute position at the bottom of the screen, you lose. I.e. it is critical to know their absolute position to know how close to losing you are, for purposes of value estimation.

So perhaps either the current agent's absolute position can be added to the observation so it can be summed with the relative position (typically an easy operation for a dense neural network) to get the zombies positions, or the zombie's absolute positions can be added on in addition to the relative positions.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -261,41 +273,124 @@ def zombie_all_players(self, knight_list, archer_list, run):
...
     def state(self):
         """
         Returns an observation of the global environment
         """
-        state = pygame.surfarray.pixels3d(self.WINDOW).copy()
-        state = np.rot90(state, k=3)
-        state = np.fliplr(state)
+        state = None
+        # change this to True if image space state is required
+        if False:
```
benblack769, line 365: Shouldn't this depend on the environment argument?

benblack769 (commented):
Looks like there was some good progress. 
1. You can add these new arguments here: https://github.com/Farama-Foundation/PettingZoo/blob/master/test/all_parameter_combs.py#L45 to make sure that CI tests are checking that it works.
2. The observation space also needs to change when this argument is set (tests won't pass without it).

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jjshoots: You're right! So used to dealing with 3D vectors I never considered this.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jjshoots: I've fixed this according to your suggestion

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jjshoots: Fixed.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jjshoots: Fixed.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jjshoots: For now, I've made the change for everything to be relative, position and angles. And the only absolute thing is the _current_ agent's observation. I suppose this makes sense because otherwise the current agent's state will always be 0, 0, 0, -1.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jjshoots: Fixed.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jjshoots: 'Fixed' - the new regime uses `1 / rel_distance` for per-agent observation, and uses a linear scale in [0, 1] for global and absolute states.

benblack769: I know I suggested this approach, but some more thought is needed, there are a lot of odd observations that can occur. 

1. Some threshold should be given so that this doesn't result in infinite blowup of the observation value. Basically some sense of when a position is "equal". A more continuous way of doing this is increase the epsilon in the denominator, perhaps to 0.1, i.e. `rel_pos = 1. / (rel_pos + 1e-1)`. This will bound the observation between [-10,10] with the distance function effectively being linear if the agent is really close. There are a lot of options here, just a suggestion for one of them. 
2. If dx=100 and dy=1, then this observation strategy will yield (0.01, 1) for the observation which is a bit strange, the value in the y direction is quite large, considering how far away the object is. Perhaps rather than the coordinates being in the denominator, the full euclidian distance should be in the denominator. `rel_pos = 1. / (np.linalg.norm(rel_pos) + 1e-6)`. 
3. If the relative position is calculated in a non-linear way like this, rather than in the obvious (x-x') method, then it will be hard to determine the zombie's absolute positions with just your own absolute position and a non-linear relative positions to the zombie.
4. The ` rel_pos = np.sign(rel_pos) * (1 - abs(rel_pos))` doesn't really make sense in combination with this inverse magnitude logic.

test/all_parameter_combs.py
```diff
@@ -43,6 +43,7 @@
     (knights_archers_zombies_v8.env, dict(line_death=False)),
     (knights_archers_zombies_v8.env, dict(pad_observation=False)),
     (knights_archers_zombies_v8.env, dict(max_cycles=100)),
+    (knights_archers_zombies_v8.env, dict(vector_state=True)),
```
benblack769, line 46: Can you also include the other new parameters here?

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -147,15 +159,16 @@ def seed(self, seed=None):
 
     # Spawn Zombies at Random Location at every 100 iterations
     def spawn_zombie(self):
-        self.zombie_spawn_rate += 1
-        zombie = Zombie(self.np_random)
+        if len(self.zombie_list) < self.max_zombies:
```
benblack769, line 162: Seems to work nicely.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -261,41 +274,119 @@ def zombie_all_players(self, knight_list, archer_list, run):
...
+
+            # give more emphasis to closer positions
+            rel_pos = np.sign(rel_pos) * (1 - abs(rel_pos))
+
+            # get rotation matrix of agent
+            c, s = np.cos(agent_ang), np.sin(agent_ang)
+            rot_mat = np.array([[c, -s], [s, c]])
+            rot_mat = np.squeeze(rot_mat).T
+
+            rel_pos = rel_pos @ rot_mat
```
benblack769, line 345: I'm probably not as experienced as you are with this sort of geometric transformation, but I wanted to check. Does this rotate the observation by the given angle, or the negative of the given angle? To make sure that "forward" is always the same direction, I think you need to rotate it by -agent_ang. Not 100% sure though.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -261,41 +274,119 @@ def zombie_all_players(self, knight_list, archer_list, run):
...
+            return state
 
     def state(self):
         """
         Returns an observation of the global environment
         """
-        state = pygame.surfarray.pixels3d(self.WINDOW).copy()
-        state = np.rot90(state, k=3)
-        state = np.fliplr(state)
+        state = None
```
benblack769, line 364: Declaring variables is not necessary in python

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -261,41 +274,119 @@ def zombie_all_players(self, knight_list, archer_list, run):
...
+            # set the current agents one to be absolute
+            # set the current agent's entity type to 3
+            state = np.concatenate([all_ids, rel_pos, rel_ang], axis=-1)
+            state[i] = agent_state
+            state[i, 3] = 1.
+            state = self.pad_vector_state(state)
+
+            return state
 
     def state(self):
```
benblack769, line 360: I see that the state() is changed, but the state_space is not. Tests aren't catching this (which is my bad), so I opened a PR that should test the state_space. https://github.com/Farama-Foundation/PettingZoo/pull/619

If this is merged in, that should cause the CI tests to fail, assuming I did it right.

benblack769 (commented):
Some more progress, some more concerns. Mainly in more complexities surrounding the observation space.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jjshoots: The implementation above is correct, the TLDR is that the transpose reverses the rotation of the rotation matrix.

The long version is:

Let's let `rel_pos` be a [1, 2] vector denoted as v and `rot_mat` be a [2, 2] matrix denoted as R.

The rotation of v about R would be something like:
v_rotated = R x v.T = v x R

To reverse the rotation of R, we take the inverse, but since it's orthogonal, the transpose will do, so:
v_reverse_rotated = R.T x v.T = v x R.T

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jjshoots: Fixed this, thanks! For some reason I never knew it could do that, probably cause my LSP keeps giving me unbounded errors.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jjshoots: Some opinions regarding those above:
1. Yes, I agree with this, I've fixed it in the latest commit.
2. I think this is generally ok. It's probably pretty easy for the network to learn that the true threat of losing is actually `rel_x * rel_y`, and not just `max(rel_x, rel_y)`. This is the basis for disentangled x and y axis anyway so it should probably be ok?
3. This is definitely true too, but it's a tradeoff between giving more emphasis to nearer zombies vs. being able to have nice linear space. In any case, I am open to any other function. 
4. Removed this, forgot to remove it earlier.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jjshoots: I've made the fix to append the current agents absolute position to the beginning of the vector, and also added an argument to allow typemasks.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -99,11 +107,15 @@ def __init__(
             self.agent_name_mapping[k_name] = a_count
             a_count += 1
 
+        shape = [512, 512, 3] if not self.vector_state else [self.num_tracked, 8]
+        low = 0 if not self.vector_state else -np.inf
```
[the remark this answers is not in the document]
jjshoots: Fixed, but will change when the base scale changes according to future decisions.

test/all_parameter_combs.py
jjshoots: Included in the latest commit.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jjshoots: Fixed.

RedTachyon:
Why is the distance expressed by the inverse of the actual distance? I might be missing something, but this actually makes it take values in [0, \infty], and that's not really good. Can't we just do effectively `[distance_x / width, distance_y / height]`?

(I'll start working on more of a review, but that's the first thing that stands out as very odd to me)

jjshoots:
@RedTachyon this was a suggestion by @benblack769. The bound is not exactly [-infty, infty] because we add an epsilon (see env.epsilon), where epsilon is like 0.1. The core idea of this is because if it were linear: 
> Not sure I like this strategy. In the current map, the difference between being inside a knights attack range and outside will be like 0.98 vs 0.96, which is pretty hard for an ML algorithm to split.

pettingzoo/butterfly/knights_archers_zombies/src/players.py
```diff
@@ -44,7 +55,7 @@ def update(self, action):
                 self.rect.x += math.cos(move_angle) * self.speed
                 self.rect.y -= math.sin(move_angle) * self.speed
             elif action == 2 and self.rect.y < const.SCREEN_HEIGHT - 40:
-                self.rect.x += math.cos(move_angle) * self.speed
+                self.rect.x -= math.cos(move_angle) * self.speed
```
RedTachyon, line 58: Was this an old bug? Or what's the reason for this change?

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
```diff
@@ -45,24 +46,43 @@ def __init__(
...
         )
 
+        # whether we want RGB state or vector state
+        self.vector_state = vector_state
+        # agents + zombies + weapons
+        self.num_tracked = (
+            num_archers + num_knights + max_zombies + num_knights + max_arrows
+        )
+        self.use_typemasks = use_typemasks
+        self.typemask_width = 6
```
RedTachyon, line 82: Why is it 6? The PR description states it should be 4 if I understand that right.

RedTachyon (commented):
I added a few specific comments, but I have some other more general doubts:

1. This very much requires some more comments and documentation, I'm still confused on how large the typemask should be and why
2. Some tests would be useful for the vector observations. Maybe do a spawn with a fixed seed, figure out what the observations should be and check that they're indeed correct
3. I disagree about the inverse distance representation, in my experience dealing with large numbers is more of a pain than thresholds like this. I guess I'm a bit late to the party, but that's my 5 cents, I'd keep it as a simple normalized distance.
4. I'd be a bit more careful with the relative position representation, it's incredibly easy to mess up (I worked with a few environments with a very similar structure). Intuitively I'd prefer to avoid the temptation of "realism" and give agents the absolute coordinates, perhaps adjusting the ordering of the representations. This could be the topic of a much larger discussion, but before that, can you please document how your current approach works exactly? (i.e. the observing agent is (x_1, y_1), a nearby zombie is (x_2, y_2), what exactly does the agent observe)

pettingzoo/butterfly/knights_archers_zombies/src/players.py
jjshoots: Yep, this was an old bug, the agent couldn't move backwards when it was facing 90 or 270 degrees.

pettingzoo/butterfly/knights_archers_zombies/knights_archers_zombies.py
jjshoots: The PR description is outdated, I ought to update it, I'll do that within the next few hours from now. But the reason it's 6 is because we also track the arrows and swords, so it's now [zombie, archer, knight, arrow, sword, you].

jjshoots:
@RedTachyon I've updated the PR description, hopefully this helps make the code a bit more understandable. If there's anything I'm missing or if there are any confusing bits, do let me know.

PR #620, opened 2022-01-29
fixed all parameter test

benblack769:
* Makes all parameter test also run render_test, seed test, state test.


benblack769:
Seems like the state space for MAgent is sometimes wrong, not sure why....

Issue #518, opened 2021-10-18
[Proposal] Adding type hints


jkterry1:
Per [this](https://sethmlarson.dev/blog/2021-10-18/tests-arent-enough-case-study-after-adding-types-to-urllib3) recently article, we really should work to add type hints to Gym over time. This is something that can be worried about after the major upcoming changes are planned, but none-the-less this is important for a package of Gym's scale. Incremental contributions towards this end are very welcome. I'm also going to be worrying about getting this added to Gym.

PR #627, opened 2022-02-11
Add typing to the core classes

RedTachyon:
Related to #518

This doesn't fully exhaust the extent of typing information that can be included, but is a small first step in the direction of sanity. The main thing is adding types to all methods in AECEnv and ParallelEnv, and also specifying all the fields that are expected to be present.

The only actual code change at the moment is adding the `__iter__` method to `AECIterator` to make it an actual `Iterator`. As a side note, I'm like 90% sure the presence of both `AECIterable` and `AECIterator` is redundant, but I didn't want to mess with it yet.


jjshoots:
Personally this LGTM, seems pretty straightforward and I can't see introducing any catastrophic bugs. Is this free to merge? Or do you want more eyes on it?

PR #634, opened 2022-02-17
Fix simple_spread global reward

singh-jayant:
Loop through agents to match with closest landmark instead of landmarks to match with closest agent. Noticed that some agents run away, while some cluster around the centroid of the landmarks with no incentive for the escaped agents to return back.

Rationale: If we loop through landmarks in a 3 agent env, and 2 agents escape while 1 sits in the centroid of the landmarks, there is no change in global reward, no matter how far the other two agents escape, thus previous function provided irrelevant reward signal.

### DIFF
--- a/docs/mpe/simple_spread.md
+++ b/docs/mpe/simple_spread.md
@@ -10,7 +10,7 @@ observation-values: "(-inf,inf)"
 state-shape: "(54,)"
 state-values: "(-inf,inf)"
 average-total-reward: "-115.6"
-import: "from pettingzoo.mpe import simple_spread_v2"
+import: "from pettingzoo.mpe import simple_spread_v3"
 agent-labels: "agents= [agent_0, agent_1, agent_2]"
 ---
 
@@ -38,7 +38,7 @@ Agent action space: `[no_action, move_left, move_right, move_down, move_up]`
 ### Arguments
 
 ``` python
-simple_spread_v2.env(N=3, local_ratio=0.5, max_cycles=25, continuous_actions=False)
+simple_spread_v3.env(N=3, local_ratio=0.5, max_cycles=25, continuous_actions=False)
 ```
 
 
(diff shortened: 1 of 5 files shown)


singh-jayant:
I have noticed poor adoption of this environment in research despite being analogous to useful, real world cooperation tasks. I believe this to be the main reason behind it

benblack769:
Thanks for the observation/suggestion. Its always useful to get deeper insights into these enviornments

However, after some reflection, I belive the current environment better models cooperative tasks than your suggestion.

Why? Because ultimately, scoring each agent by its closeness to a landmark is is not a cooperative task, it is an independent one. Each agent can greedily increase its reward without communication or regard for its peers.

In the current environment, agents can try to coordinate who gets to which landmark, through some sort agreed upon communication or arbitration protocol. They are rewarded for successfully coordinating because there is a global shared reward, so agents are rewarded for other agents getting closer to other landmarks. So it is a cooperative environment.

Note that like all cooperation, this is a difficult task. I am not suprised that standard RL approaches don't work particularly well, especially given the dubiously constructed observation space, and the shallowness of networks that are typically applied to this enviornment.

singh-jayant:
Hi, I realise this does provide independent rewards (and indeed the agents are very greedy) but we can fix the cooperation issue by taking a mean (or sum) over the agent rewards and then just repeat it n times so its the same reward for all agents. I was more concerned about the issue of poor reward signal.

Image

The image above kind of represents a rough scenario thats inevitable in this environment and inescapable because the reward signal simply doesn't cover any agent in the red region. Since it is the landmarks matching to the closest agent, any state action pair in the region outside would have no effect on the reward since there would almost certainly be another agent that is closer to the 3 landmarks. This makes the red region a credit assignment hell. There seems no way for an agent to learn this environment, since the reward signal does not depend on the two agents on the right.

I'll probably make the changes I suggested tonight and try again.

benblack769:
Instead of dramatically changing the environment, if the problem is a poor reward signal, I would recomend just addding some optional reward shaping. 

For example, instead of rewarding the min distance agent to the landmark, the softmin can be taken. A temperature parameter can be added to control the temperature of the softmin function, perhaps defaulting to the hardmin (temp=0) for evaluation.

jkterry1:
Hey, I'm going to close this after some discussions with Ben

PR #635, opened 2022-02-19
Fix agent labels

trigaten:


benblack769:
Hmm. Looks like the randint() issue finally caught up to us. I'll try to fix this in another PR

PR #638, opened 2022-02-22
Add human enabled control

### LINKED ISSUE #547: user input
Hi, 
Is it possible to have some user-input controlled interface like the [one](https://github.com/openai/mlsh/blob/master/gym/examples/agents/keyboard_agent.py) in OpenAI Gym?
I tried to modify the script by gym on pettingzoo but it seems there is some in compatibility in the `to_parallel_wrapper`. I got the error report `AttributeError: 'to_parallel_wrapper' object has no attribute 'viewer'
`.
The script I used is here:
```
 #!/usr/bin/env python
from __future__ import print_function

import sys, gym
import time
from env.import_env import make_env
from utils.data_struct import AttrDict


#
# Test yourself as a learning agent! Pass environment name as a command-line argument.
#
EnvArgs = {
    'name': None,
    'type': None,
    'num_envs': 1, 
    'ram': True, 
    'against_baseline': False,
    'seed': 1223,
    }
EnvArgs['env_type'] = 'pettingzoo'
EnvArgs['env_name'] = 'boxing_v1'
env_args = AttrDict(EnvArgs)

env = make_env(env_args)

print(env.action_space)
if not hasattr(env.action_space, 'n'):
    raise Exception('Keyboard agent only supports discrete action spaces')
ACTIONS = env.action_space.n
ROLLOUT_TIME = 1000
SKIP_CONTROL = 0    # Use previous control decision SKIP_CONTROL times, that's how you
                    # can test what skip is still usable.

human_agent_action = 0
human_wants_restart = False
human_sets_pause = False

def key_press(key, mod):
    global human_agent_action, human_wants_restart, human_sets_pause
    if key==0xff0d: human_wants_restart = True
    if key==32: human_sets_pause = not human_sets_pause
    a = int( key - ord('0') )
    if a <= 0 or a >= ACTIONS: return
    human_agent_action = a

def key_release(key, mod):
    global human_agent_action
    a = int( key - ord('0') )
    if a <= 0 or a >= ACTIONS: return
    if human_agent_action == a:
        human_agent_action = 0
env.reset()
env.render()
print(env, env.unwrapped, env.unwrapped.unwrapped.unwrapped)
env.unwrapped.viewer.window.on_key_press = key_press
env.unwrapped.viewer.window.on_key_release = key_release

def rollout(env):
    global human_agent_action, human_wants_restart, human_sets_pause
    human_wants_restart = False
    obser = env.reset()
    skip = 0
    for t in range(ROLLOUT_TIME):
        if not skip:
            #print("taking action {}".format(human_agent_action))
            a = human_agent_action
            skip = SKIP_CONTROL
        else:
            skip -= 1

        obser, r, done, info = env.step(a,a)
        env.render()
        time.sleep(0.05)
        if done: break
        if human_wants_restart: break
        while human_sets_pause:
            env.render()
            time.sleep(0.1)

print("ACTIONS={}".format(ACTIONS))
print("Press keys 1 2 3 ... to take actions 1 2 3 ...")
print("No keys pressed is taking action 0")

while 1:
    rollout(env)
```

jjshoots:
This PR is a response to and closes #547.

As a preliminary, it has been only added to KAZ. Once the format is satisfactory and scalable, I will add it to the other environments too.

## How do thing
Simply run `python3 pettingzoo/butterfly/knights_archers_zombies/manual_policy.py` to get a demo.

## Known Bugs
- At the moment, there is a bug in the agent movements that is fixed in [this line of code](https://github.com/Farama-Foundation/PettingZoo/pull/618/files#diff-6e39d112e7fbf8e3dca08424ae78348c4c9916ef57c1ce0d2f45271b15393890R58).
- ~~Issue from #636 preventing CI from passing~~

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -194,6 +194,43 @@ from pettingzoo.utils import random_demo
 random_demo(env, render=True, episodes=1)
 ```
 
+### Playing Alongside Trained Policies
+
+Sometimes, you may want to control a singular agent and let the other agents be controlled by trained policies.
+Some games support this via:
+
+``` python
+import time
+from pettingzoo.butterfly import knights_archers_zombies_v8
+
+env = knights_archers_zombies_v8.env()
+env.reset()
+
+manual_policy = knights_archers_zombies_v8.ManualPolicy(env)
+
+for agent in env.agent_iter():
+    observation, reward, done, info = env.last()
+
+    if agent == manual_policy.agent:
+        action = manual_policy(observation, agent)
+    else:
+        action = policy(observation, agent)
+
+    env.step(action)
+
+    env.render()
+    time.sleep(0.05)
+
+    if done:
+        env.reset()
+```
+
+`ManualPolicy` accepts several default arguments:
+
+`agent_id`: Accepts an integer for the agent in the environment that will be controlled via the keyboard. Use `manual_policy.availabla_agents` to query what agents are available and what are their indices.
+
+`show_obs`: Is a boolean which shows the observation from the currently selected agent, if available.
+
 ### Observation Saving
 
 If the agents in a game make observations that are images then the observations can be saved to an image file. This function takes in the environment, along with a specified agent. If no `agent` is specified, then the current selected agent for the environment is chosen. If `all_agents` is passed in as `True`, then the observations of all agents in the environment is saved. By default, the images are saved to the current working directory in a folder matching the environment name. The saved image will match the name of the observing agent. If `save_dir` is passed in, a new folder is created where images will be saved to. This function can be called during training/evaluation if desired, which is why environments have to be reset before it can be used.
--- a/pettingzoo/butterfly/knights_archers_zombies/manual_policy.py
+++ b/pettingzoo/butterfly/knights_archers_zombies/manual_policy.py
@@ -15,11 +15,11 @@ class ManualPolicy:
         if True:
             self.default_action = 5
             self.action_mapping = dict()
-            self.action_mapping[pygame.K_w] = 0 # front
-            self.action_mapping[pygame.K_s] = 1 # back
-            self.action_mapping[pygame.K_q] = 2 # rotate left
-            self.action_mapping[pygame.K_e] = 3 # rotate right
-            self.action_mapping[pygame.K_f] = 4 # weapon
+            self.action_mapping[pygame.K_w] = 0  # front
+            self.action_mapping[pygame.K_s] = 1  # back
+            self.action_mapping[pygame.K_q] = 2  # rotate left
+            self.action_mapping[pygame.K_e] = 3  # rotate right
+            self.action_mapping[pygame.K_f] = 4  # weapon
 
     def __call__(self, observation, agent):
         # only trigger when we are the correct agent
--- a/pettingzoo/butterfly/knights_archers_zombies/manual_policy.py
+++ b/pettingzoo/butterfly/knights_archers_zombies/manual_policy.py
@@ -17,9 +17,9 @@ class ManualPolicy:
             self.action_mapping = dict()
             self.action_mapping[pygame.K_w] = 0  # front
             self.action_mapping[pygame.K_s] = 1  # back
-            self.action_mapping[pygame.K_q] = 2  # rotate left
-            self.action_mapping[pygame.K_e] = 3  # rotate right
-            self.action_mapping[pygame.K_f] = 4  # weapon
+            self.action_mapping[pygame.K_a] = 2  # rotate left
+            self.action_mapping[pygame.K_d] = 3  # rotate right
+            self.action_mapping[pygame.K_SPACE] = 4  # weapon
 
     def __call__(self, observation, agent):
         # only trigger when we are the correct agent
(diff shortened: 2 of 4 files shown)


jkterry1:
This needs to be added to docs

docs/api.md
```diff
@@ -194,6 +194,43 @@ from pettingzoo.utils import random_demo
...
+
+    if agent == manual_policy.agent:
+        action = manual_policy(observation, agent)
+    else:
+        action = policy(observation, agent)
+
+    env.step(action)
+
+    env.render()
+    time.sleep(0.05)
```
benblack769, line 222: The manual policy already has a timer slowing it down, I don't think this should be necessary in addition.

benblack769: Sorry, ignore the above. I would recomend adding the clock.tick like in main() rather than this time in this example code.

pettingzoo/butterfly/knights_archers_zombies/manual_policy.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # TO-DO: show current agent observation if this is True
+        self.show_obs = show_obs
+
+        # action mappings for all agents are the same
+        if True:
+            self.default_action = 5
+            self.action_mapping = dict()
+            self.action_mapping[pygame.K_w] = 0  # front
+            self.action_mapping[pygame.K_s] = 1  # back
```
benblack769, line 19: Weirdly, moving front and then back does not leave me in the same place. Is this a known feature in the environment?

pettingzoo/butterfly/knights_archers_zombies/manual_policy.py
```diff
@@ -0,0 +1,77 @@
...
+            self.action_mapping[pygame.K_q] = 2  # rotate left
+            self.action_mapping[pygame.K_e] = 3  # rotate right
+            self.action_mapping[pygame.K_f] = 4  # weapon
```
benblack769, line 22: Is wsqe a known key mapping? I found it a bit unfamiliar. But I could just be out of the loop.

docs/api.md
```diff
@@ -194,6 +194,43 @@ from pettingzoo.utils import random_demo
...
+
+manual_policy = knights_archers_zombies_v8.ManualPolicy(env)
+
+for agent in env.agent_iter():
+    observation, reward, done, info = env.last()
+
+    if agent == manual_policy.agent:
+        action = manual_policy(observation, agent)
+    else:
+        action = policy(observation, agent)
```
benblack769, line 217: Might want to mention that this `policy` is a function you need to define yourself, it comes from something you trained or wrote, not from anywhere in pettingzoo. This has caused confusion elsewhere in the documentation.

benblack769:
Looks pretty good, once I figured out the controls I thought it was pretty slick. Just a few comments, and a question about the environment.

pettingzoo/butterfly/knights_archers_zombies/manual_policy.py
jjshoots: Yes, this is a bug in the KAZ environment currently. A fix for it is in the vector state PR.

pettingzoo/butterfly/knights_archers_zombies/manual_policy.py
jjshoots: Originally I used WASD and space for attack, but then manual_control.py uses WSQE and F, and uses AD to change agents. I wasn't sure what tonuse, so stuck with the one in manual_control.py. That said, this can be easily changed, though I suspect similae issues will crop up in other envs.

docs/api.md
jjshoots: Gotcha

pettingzoo/butterfly/knights_archers_zombies/manual_policy.py
jkterry1: Could you please use WASD and space for attack and then use QE to switch agents? I think the keybindings were done by Mario years ago with zero review

elliottower:
Not sure if I'm supposed to comment here or add a new issue/PR but there's a slight bug, when running the manual_policy.py it gives an error that the character is dead and doesn't render anything. The fix is to change [Line 59](https://github.com/Farama-Foundation/PettingZoo/blob/master/pettingzoo/butterfly/knights_archers_zombies/manual_policy.py#L59l) to specify render mode: `env = knights_archers_zombies_v10.env(render_mode="human")
`

As a side note is there any progress on the general API for manual policies? I implemented it a slightly different way for a custom env for a game I'm working on ([gobblet-rl](https://github.com/elliottower/gobblet-rl)) but I like the flag `has_manual_policy` and separating the manual input into a separate file, will probably restructure my code to do the same thing.

pseudo-rnd-thoughts:
@jjshoots or @WillDudley Do you understand the problem or could fix it?

elliottower:
Looking at the code I think it's just that ``knights_archers_zombies.py`` has render_mode=None by default, whereas in the documentation it says the metadata: `` "render_modes": ["human", "rgb_array"]``, so I think it must just be either a bug and it should be human by default, or that `manual_policy.py` [L59 ](https://github.com/Farama-Foundation/PettingZoo/blob/master/pettingzoo/butterfly/knights_archers_zombies/manual_policy.py#L59l) should just be changed from .env() to .env(render_mode="human"), as I said above.

PR #639, opened 2022-03-02
Update deprecated_module.py

jackyoung96:
Modify version check inequality 
from ```int(alt_version) > int(version)```
to ```int(alt_version) >= int(version)```

### DIFF
--- a/pettingzoo/utils/deprecated_module.py
+++ b/pettingzoo/utils/deprecated_module.py
@@ -36,7 +36,7 @@ def depricated_handler(env_name, module_path, module_name):
         if is_env(alt_env_name):
             alt_name, alt_version = alt_env_name.rsplit("_v")
             if alt_name == name:
-                if int(alt_version) > int(version):
+                if int(alt_version) >= int(version):
                     return DeprecatedModule(name, version, alt_version)
                 else:
                     raise ImportError(f"cannot import name '{env_name}' from '{module_name}'")


benblack769:
Thanks for the PR. I'm curious what problem this solves, because in my understanding this shouldn't change anything. If the version is equal, then the actual module should just be imported and this exception handling routine should not be called.

I would like to know where my understanding is going wrong.

jackyoung96:
You are right. The problem is that the below line is not called when the first time you install the pettingzoo.
https://github.com/Farama-Foundation/PettingZoo/blob/14c78e73fe7f866ecb7e3be4c1f4532b802a913f/pettingzoo/utils/deprecated_module.py#L27

I think 
https://github.com/Farama-Foundation/PettingZoo/blob/14c78e73fe7f866ecb7e3be4c1f4532b802a913f/pettingzoo/utils/deprecated_module.py#L40
this line should be called at least once for registering.

I realized this because I got an error when I use pettingzoo==1.15.0. But after downgrading to pettingzoo==1.11.2, the error disappear and the only difference is called the below line once or not.
https://github.com/Farama-Foundation/PettingZoo/blob/14c78e73fe7f866ecb7e3be4c1f4532b802a913f/pettingzoo/utils/deprecated_module.py#L40

benblack769:
@jackyoung96 Hmm. Yes, this might be the cause of a ghost we have been seeing for a while where `from pettingzoo.module import env` just fails for no good reason, but especially seems to happen when you change versions. We should try to figure out if this or something similar fixes the problem. If so, then you are a hero here.

benblack769:
I tried to reproduce the ghostly problem I am familiar with, and this did not fix it, it just changed the nature of the problem. We may end up having to do what Gym does and [import everything with try;except:](https://github.com/openai/gym/blob/master/gym/envs/mujoco/mujoco_env.py#L11) in order to fix this problem.

benblack769:
Closing in favor of #643. That change should make the error messages much more reasonable in the future, so if you have this problem again, the error message should be much more reasonable, and we can make a much more direct fix.

Thanks again for bringing this up.

PR #643, opened 2022-03-14
fixed depricated wrapper

benblack769:
Makes change that Jesse suggested that should allow errors to be handled in a much better way, allowing us to see what is really going on.


benblack769:
Seems to work as before, if this fixes the MacOS tests (which still appear to be failing), then I think we can tentatively call  this issue fixed.

benblack769:
And it seemed to fix the MacOS tests. Great.

PR #645, opened 2022-03-21
changed atari action space default

benblack769:
This change brings PettingZoo's action space in closer alignment with Gym's, removing duplicate actions and leaving unique ones, but has the following concerns:

1. A single unifier controller is not possible without the full action space. This could be an issue when building a human interface or when generalizing across environments.
2. A full action space is much easier to understand and document.


benblack769:
This PR needs to document the elements of the full action space, as well as the elements of each reduced action space. This can be found in the C code https://github.com/Farama-Foundation/Multi-Agent-ALE/blob/master/src/common/Constants.h#L27

And the reduced set can be found here: https://github.com/Farama-Foundation/Multi-Agent-ALE/blob/master/src/games/supported/Surround.cpp#L64

PR #651, opened 2022-04-04
Update mpe.md

nollied:
i haven't seen any environments with variable observation sizes.


benblack769:
Ah, sorry about the ambiguous language in the documentation. 

What we meant is that some of the agents in the MPE environments have different observation sizes than other agents in the same environment. For example, in speaker_listener, the speaker can only see which goal is the target, and the other agent can only see its own position and velocity, as well as whatever the speaker says.

jjshoots:
Hi @nollied, would it be possible to update the docs to reflect the thing that Ben has mentioned? Then we can close this PR after that.

benblack769:
Closed in favor of #664. Feel free to comment if you think that doc change is still unclear.

PR #653, opened 2022-04-07
filter action settings of dead agents and remove group handle swap

### LINKED ISSUE #542: Something wrong with magent/battle_v3 -- some agents doesn't do actions
```
from pettingzoo.magent import battle_v3
from pettingzoo.utils import random_demo

env = battle_v3.env(map_size=20, minimap_mode=False, step_reward=-0.005,
                    dead_penalty=-0.1, attack_penalty=-0.1, attack_opponent_reward=0.2,
                    max_cycles=100000, extra_features=False)
env.reset()
random_demo(env, render=True, episodes=1)
```

After running for some time, some agents went on strike。

https://user-images.githubusercontent.com/53855240/142435576-08d25d2f-9b95-470a-8135-7f38780f229b.mp4

dsctt:
Closes #542
The results can be confirmed by running the script in there.  

Two main issues:
1. Teams (e.g. blue, red) not matching up with the group handles in magent. For example, the second team created would have the 0'th index handle in magent but the `magent_env` code assumes they are in order. As a result, actions would be assigned to opposite teams. I'm not sure why they were switched in `battle_v3` but I removed this. If they need to be that way, let me know why and I can try to make the adjustment happen downstream.
2. In magent c++, when an agent dies it gets deleted and the `agents` array for its group gets resized but the python code was setting actions as if it was still there but being ignored which further caused incorrect action assignments. This PR adds a filter for dead agents before sending actions.


pettingzoo/test/parallel_test.py
```diff
@@ -80,3 +82,6 @@ def parallel_api_test(par_env, num_cycles=1000):
                     live_agents.remove(agent)
 
             assert set(par_env.agents) == live_agents
+
+            if len(live_agents) == 0:
```
dsctt, line 86: Honestly, not sure about this one. It fixes the test for `tiger_deer_v3` but line 24 fixes the test for the other one that had an error which was `adversarial_pursuit_v3` and that line makes sense whereas I feel like this one shouldn't be needed. For some reason tiger_deer keeps resetting at 100 and I cannot figure out why. It was doing this before but the difference is this PR no longer allows mismatching action settings to go through (see line 196 of `pettingzoo/magent/magent_env.py` above). Up to you if I should keep trying to figure tiger_deer out as part of this PR.

pettingzoo/magent/magent_env.py
```diff
@@ -118,6 +113,7 @@ def reset(self):
         self.env.reset()
         self.frames = 0
         self.all_dones = {agent: False for agent in self.possible_agents}
+        self.team_sizes = [self.env.get_num(handle) for handle in self.handles]
```
dsctt, line 116: It could be cleaner to refactor and call this reset() function in __init__. I was trying it out but it started requiring more changes than I expected, leading to other things irrelevant to this PR.

pettingzoo/test/parallel_test.py
benblack769: Tiger-deer always resets at time step 100 because all the tigers die of starvation at the same time, because they failed to eat any of the deer.

pettingzoo/magent/magent_env.py
```diff
@@ -191,10 +188,12 @@ def state(self):
         return state
 
     def step(self, all_actions):
-        action_list = [0] * self.max_num_agents
+        action_list = [-1] * self.max_num_agents
         for i, agent in enumerate(self.possible_agents):
             if agent in all_actions:
                 action_list[i] = all_actions[agent]
+
+        action_list = list(filter(lambda v: v != -1, action_list))
```
benblack769, line 196: Instead of filtering, would it work if the agents were simply iterated over?

```
action_list = []
for i, agent in self.agents:
    action_list.append(all_actions[agent])
```

pettingzoo/test/parallel_test.py
benblack769: So I think this is intended behavior.

dsctt: Ah that's right. That reminds me, I was thinking about this case and whether this then makes for a good mechanism for determining if its done. The current implementation would make sense if we assume that in every use case, all agents being dead = done. If there are other conditions for doneness, maybe using something more direct is better. What do you think?

pettingzoo/magent/magent_env.py
dsctt: Agree

benblack769:
I'm happy with this PR, I think it is ready to merge.

PR #654, opened 2022-04-08
Fixed reward termination logic

matteobettini:
The reward logic for this environment has serious bugs. 

The first reward issue is tackled in this pull request, it fixes the following problems:

- Multiple termination rewards were assigned in case multiple termination criteria were met.
- Reward for a fallen walker was assigned to all walkers
- Fallen walkers where also receiving an extra termination reward for no reason

The pull request implements the logic from the documentation:

- Fall reward is assigned only to the actually fallen workers
- Termination reward is assigned only once in case any of the termination criteria is met
- Improves the code flow 

Alongside this, also the documentation on the website is old and does not reflect the current environment. It says "A positive reward is awarded to each walker, which is the change in the package distance summed with 130 times the change in the walker’s position.", but in the code the walker position is never used in the positive reward

### DIFF
--- a/docs/sisl/multiwalker.md
+++ b/docs/sisl/multiwalker.md
@@ -24,7 +24,7 @@ agent-labels: "agents= ['walker_0', 'walker_1', 'walker_2']"
 </div>
 
 
-In this environment, bipedal robots attempt to carry a package as far right as possible. A package is placed on top of 3 (by default) bipedal robots which you control. A positive reward is awarded to each walker, which is the change in the package distance summed with 130 times the change in the walker's position. The maximum achievable total reward depends on the terrain length; as a reference, for a terrain length of 75 steps with the scaling factor of 130, the total reward under an optimal policy is around 300 . If `shared_reward` is chosen (True by default), all agents receive the mean reward across agents. The agents also receive a small shaped reward of -5 times the change in their head angle to keep the head straight.
+In this environment, bipedal robots attempt to carry a package as far right as possible. A package is placed on top of 3 (by default) bipedal robots which you control. A positive reward is awarded to each walker, which is proportional to the change in the package distance and is scaled by `forward_reward`. The maximum achievable total reward depends on the terrain length; as a reference, for a terrain length of 75 steps, the total reward under an optimal policy is around 300 . If `shared_reward` is chosen (True by default), all agents receive the mean reward across agents. The agents also receive a small shaped reward of -5 times the change in their head angle to keep the head straight.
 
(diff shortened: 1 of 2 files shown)


jkterry1:
Hey, thanks so much for this. Would you mind adding doc fixes to your PR?

benblack769:
Thanks for the contribution. We tried to make the rewards sane in the v5 update, but there are a lot of cases with all the arguments we allow, and it means that some of the logic is very confusing, but mostly correct, I think. I'll leave more specific  comments soon.

Some changes that definitely should be made are:

* Documentation for the package should change to "positive reward is awarded to each walker, which is the change in the package distance times 130."
* self.game_over does not appear to be assigned anywhere, so it should probably be removed, unless I am missing something.

matteobettini:
@benblack769 self.game_over is set in line 68 and 71.

I see no way in which the current termination logic could be correct. I broke down the problems in my initial message. There are typos, termination rewards being assigned multiple times, fall rewards assigned to all walkers, and things done in the wrong order. I agree on the doc change.

matteobettini:
I added some fixes to the docs.

I have to say, at the current state, a complete restructure of docs and code could be very useful. As a new user of the environment I found the docs diverge significantly from the code, and both very confusing.

The changes I am proposing just temporary fix the most compelling issues.

benblack769:
I went through and commented the current code to show what I think the code is doing and why I don't think it is assigning any rewards twice:

```python
# fallen_walkers is a vector of booleans. So this line 
# will distribute to become a vector of floats which 
# will assign 0 if the walker is not fallen and
# fall_reward if the walker is fallen
rewards += self.fall_reward * self.fallen_walkers

# if the environment is terminated due to the fall, it 
# assigns all walkers a terminatal reward, penalizing them
# for the environment finishing unproductively
# the agent which actually falls will have both this reward
# and the fall reward added together
if self.terminate_on_fall and np.sum(self.fallen_walkers) > 0:
	rewards += self.terminate_reward
	done = [True] * self.n_walkers
for i, (fallen, walker) in enumerate(zip(self.fallen_walkers, self.walkers)):
	if fallen:
		if not done[i]:
			# this reward is only relevant if self.terminate_on_fall is False,
			# since if terminate_on_fall is true, then done[i] will be True by the above condition
			# what this does is assign the terminate_reward to the agent 
			# which falls if terminate_on_fall is false, since for that walker, 
			# the environment did terminate.
			rewards[i] += self.terminate_reward
		if self.remove_on_fall:
			walker._destroy()
```

PR #656, opened 2022-04-13
turn_based_aec_to_parallel no longer inherits from aec_to_parallel

RyanNavillus:
Prevent the is_parallelizable warning from running when using the new wrapper, since it is effectively still an aec env that you interact with using the parallel API


benblack769:
Looks good to me.

PR #671, opened 2022-04-26
fix CI tests to use release_test.sh

jjshoots:
This fixes CI to use release_test.sh


benblack769:
Makes sense to me. The Macos tests are a subset of the linux tests because 1) the graphical setup is different and 2) macos in the cloud is extremely expensive an often has limited capacity for CI.

PR #672, opened 2022-04-27
Removed prison environment and all references to the same.

aditya-pola:
Replaced prison_v4 with prospector_v4 in docs/api.md under examples.


docs/api.md
```diff
@@ -179,8 +179,8 @@ Where `max_episodes` and `max_steps` both limit the total number of evaluations
 Often, you want to be able to play before trying to learn it to get a better feel for it. Some of our games directly support this:
 
 ``` python
-from pettingzoo.butterfly import prison_v4
-prison_v4.manual_control(<environment parameters>)
+from pettingzoo.butterfly import prospector_v4
```
jjshoots, line 182: We should change this to KAZ instead instead of prospector, prospector will be culled in the near future.

PR #696, opened 2022-05-09
added options argument to reset()

BolunDai0216:
Added `options` argument to `reset()`.


BolunDai0216:
I think SuperSuit also needs to be updated to pass the 13 lines test, I will do a PR for SuperSuit in the next few days.

PR #697, opened 2022-05-10
Draft with Classic Envs

RushivArora:
Currently only experimenting with backgammon to make sure I get the right idea.

backgammon_envs renamed to backgammon.py (all `raw_envs` and `envs` now in backgammon.py)
backgammon.py renamed to bg_game.py
Fixed the dependencies when changing file names


RushivArora:
Closing this in favour of #711

PR #708, opened 2022-05-27
Added `info` as the output of `reset()`

BolunDai0216:
Added `info` as an output of `reset()` and added `return_info` as an argument to the `reset()`


BolunDai0216:
The tests are failing due to Supersuit not yet updated.

PR #728, opened 2022-06-14
Make official tutorial more prominent

cool-RR:
As discussed on Discord.

### DIFF
--- a/docs/tutorials.md
+++ b/docs/tutorials.md
@@ -5,19 +5,22 @@ layout: "links"
 
 # Tutorials
 
-##  Multi-Agent Deep Reinforcement Learning in 13 Lines of Code Using PettingZoo
-[https://towardsdatascience.com/multi-agent-deep-reinforcement-learning-in-15-lines-of-code-using-pettingzoo-e0b963c0820b](https://towardsdatascience.com/multi-agent-deep-reinforcement-learning-in-15-lines-of-code-using-pettingzoo-e0b963c0820b)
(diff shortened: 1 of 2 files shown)


docs/tutorials.md
```diff
@@ -5,19 +5,22 @@ layout: "links"
 
 # Tutorials
 
-##  Multi-Agent Deep Reinforcement Learning in 13 Lines of Code Using PettingZoo
-[https://towardsdatascience.com/multi-agent-deep-reinforcement-learning-in-15-lines-of-code-using-pettingzoo-e0b963c0820b](https://towardsdatascience.com/multi-agent-deep-reinforcement-learning-in-15-lines-of-code-using-pettingzoo-e0b963c0820b)
+## Official tutorial
```
jjshoots, line 8: Could we rename it from Official Tutorial to something like Recommended Start or something similar? The issue is that we're not the maintainers of SB3, and SB3 has had historically very slow updates to support the latest Gym versions and by extension PZ.

By definition, all the tutorials are actually pretty much the same, just done using different RL libraries (which we are not the maintainers for) or no library at all in the case of Tic Tac Toe.

cool-RR:
(It looks like the test failure has nothing to do with my change and can be ignored.)

cool-RR:
I see the change to https://www.pettingzoo.ml/tutorials , I'm guessing it'll be deployed at some point in the future without my intervention.

Issue #719, opened 2022-06-03
[Proposal] Refactor Pytest multiple calls to a single call


kir0ul:
### Proposal

I would suggest to rewrite the 7 calls to Pytest in [`release_test.sh`](https://github.com/Farama-Foundation/PettingZoo/blob/master/release_test.sh) to a single call.

### Motivation

I find the [current testing logs](https://github.com/Farama-Foundation/PettingZoo/runs/6713136740?check_suite_focus=true#step:6:1) pretty unreadable, instead of having to navigate between 7 different testing logs, it would be nice to have one single call to Pytest that recapitulates all the warnings, errors, and passed tests at the end.

### Pitch

I would suggest to remove the `release_test.sh` Bash script completely and simply call `pytest -v` which will run all of those tests in one call.

I've been told that this Bash script is useful to run only a subset of the tests by commenting the others. To me it seems easy enough to run any of those manually by writing for example `pytest ./test/variable_env_test.py` in the terminal, but if it is useful to people to have them separated and already written, then I would suggest to create a `Makefile` with those 7 commands to be run easily without having to comment anything into a Bash script.

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/PettingZoo/issues) in the repo (**required**)

jjshoots:
I'm in agreement with removing release test in favour of something more elegant, it's probably OK for you to go ahead and get this done. I see that SuperSuit doesn't have this funky release test thing as well.

kir0ul:
@jjshoots (or anyone else), do you want me to actually add `make` commands to replace the calls in `release_test.sh`?

jjshoots:
@kir0ul If we can consolidate it such that the file is no longer a thing, that would be good yes. Though I don't think using `make` would be a good idea, we already have enough dependencies for tests as is and people are bound to have problems. That said, if `make` is deemed superior, I'm open to reviewing what it would look like and may consider using that instead.

kir0ul:
I don't know if it's superior, indeed it's a pretty old tool but it's one of the standard Linux tools. I guess I'll create a `Makefile` in a single commit so that you can test if you like it, and if not I'll just revert it.

kir0ul:
So it's in this commit: https://github.com/Farama-Foundation/PettingZoo/pull/731/commits/2ab0f6974bfb811e67fc63c215bf9897e883c741.
You can install it with:
[...]

PR #731, opened 2022-06-19
Refactor Pytest multiple calls to a single call

kir0ul:
# Description

This PR fixes issue https://github.com/Farama-Foundation/PettingZoo/issues/719. The main changes are the following:
- Removed `release_test.sh`
- Made according changes to the docs
- Bumped GH actions versions
- Updated the `tests` section in `setup.py`

## Type of change

- Refactoring

# Checklist:

- [x] I have run `./release_test.sh` and no errors are present.
- [x] I solved any possible warnings that `./release_test.sh` has generated that are related to my code to the best of my knowledge.


kir0ul:
A few comments:
- Now we can see [there's a total of 705 warnings](https://github.com/Farama-Foundation/PettingZoo/runs/6957890431?check_suite_focus=true#step:6:160) :scream: 
- There seems to be many duplicates between `requirements.txt` and `setup.py`, maybe that could be simplified?

jjshoots:
`requirements.txt` is actually pretty heavily outdated and it may be a good idea to consolidate it into just having `.`.

kir0ul:
> `requirements.txt` is actually pretty heavily outdated and it may be a good idea to consolidate it into just having `.`. Do you think you could do that, then I'll merge this PR? Thanks!

Sounds good, will give it a try!

kir0ul:
I think I start to understand the main reason why the `release_test.sh` script was needed: tests were not discovered automatically by Pytest, so you could not run all the tests at once simply by running the `pytest` command.

I fixed it by [adding a bit of configuration in the `pyproject.toml` file](https://github.com/Farama-Foundation/PettingZoo/blob/2ab0f6974bfb811e67fc63c215bf9897e883c741/pyproject.toml#L15=).
Another solution would have been to rename the test files by prepending `test_` to the file names, for example `all_parameter_combs.py` should have been called `test_all_parameter_combs.py`. I believe the latter is the [conventional way of doing things](https://docs.pytest.org/en/stable/explanation/goodpractices.html#conventions-for-python-test-discovery) as you don't need to configure anything.

Also now we can see [in the test logs](https://github.com/Farama-Foundation/PettingZoo/runs/6997040481?check_suite_focus=true#step:6:616) that there's a total of 111739 warnings :scream: (some tests and thus some warnings were missing in my previous comment) and that the test coverage is 84%.

jjshoots:
I think changing all the tests to reflect the proper name change is probably a good idea.

kir0ul:
@jjshoots Are the files under `pettingzoo/test` (`api_test.py`, `bombardment_test.py`, `manual_control_test.py`, `max_cycles_test.py`, `parallel_test.py`, `render_test.py`, `save_obs_test.py`, `seed_test.py`, `state_test.py`) part of the test suite?

kir0ul:
> I think changing all the tests to reflect the proper name change is probably a good idea.

Seems that appending `*_test.py ` [is also standard](https://docs.pytest.org/en/stable/explanation/goodpractices.html#conventions-for-python-test-discovery), so I just renamed the test files this way as some files were already named like that.

kir0ul:
> @kir0ul I believe the tests under `pettingzoo/tests` are simply the API tests for when a user creates a custom environment. So those do not need to be part of the tests suite.

Ok, so I had to [ignore those](https://github.com/Farama-Foundation/PettingZoo/blob/62f5f734a563a405226ece9594e3a670a23c10e4/pyproject.toml#L17=) otherwise they were auto discovered by Pytest.

Now we have a [total of 222 tests](https://github.com/Farama-Foundation/PettingZoo/runs/7016975769?check_suite_focus=true#step:6:616) that are run in CI.

kir0ul:
`test_chess` was run twice so I [ignored all the tests in the `pettingzoo` folder](https://github.com/Farama-Foundation/PettingZoo/pull/731/commits/ff3888b95339f4dbb416e9fc8501aa827f43fbe5). Now we should be back at 221 tests total as it was the case before this PR.

CONTRIBUTING.md
```diff
@@ -33,3 +32,13 @@ The CI will run several checks on the new code pushed to the PettingZoo reposito
...
+
+### Testing shortcuts
+A few _Make_ commands are provided to quickly test parts of the test suite.
+
+To use it you first need to have _Make_ installed on your local machine:
+- Linux: `sudo apt install make` (depends on the distribution/package manager)
+- MacOS: `brew install make`, the command should then be available as `gmake` instead of `make` (https://formulae.brew.sh/formula/make)
+- Windows: `choco install make` (https://community.chocolatey.org/packages/make)
+
+Then `make help` should give all the available commands and you can use `Tab` when writing the command if it's too long to write. For example to run all the tests, one would write `make test-all`.
```
jjshoots, line 44: Now that I'm reading this, it's brilliant.

jjshoots:
@kir0ul I've looked through and things look pretty good, I'll need to keep this on backburner for now while we figure out why master is failing. Once master passes, we'll merge.

PR #737, opened 2022-07-06
Move Pyright to `pre-commit` + add `pydocstyle`

kir0ul:
# Description

This PR moves the Pyright checks from CI only to `pre-commit` (both local + CI), fixes some typing issues, and add also `pydocstyle` to `pre-commit`.

## Type of change

- Refactoring/maintenance

# Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,16 +38,19 @@ repos:
     hooks:
       - id: mixed-line-ending
         args: ["--fix=lf"]
-  # - repo: https://github.com/pycqa/pydocstyle
-  #   rev: 6.1.1
-  #   hooks:
-  #     - id: pydocstyle
-  #       args:
-  #         - --source
-  #         - --explain
-  #         - --convention=google
-  #         - --count
-  #       additional_dependencies: ["toml"]
+  - repo: https://github.com/pycqa/pydocstyle
+    rev: 6.1.1
+    hooks:
+      - id: pydocstyle
+        args:
+          - --source
+          - --explain
+          - --convention=google
+          - --count
+          # TODO: Remove ignoring rules D101, D102, D103, D105
+          - --add-ignore=D100,D107,D101,D102,D103,D105
+        exclude: "__init__.py$|^pettingzoo.test|^docs"
+        additional_dependencies: ["toml"]
   - repo: local
     hooks:
       - id: pyright
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 ---
 repos:
   - repo: https://github.com/python/black
-    rev: 22.3.0
+    rev: 22.6.0
     hooks:
       - id: black
   - repo: https://github.com/codespell-project/codespell
@@ -28,13 +28,32 @@ repos:
       - id: isort
         args: ["--profile", "black"]
   - repo: https://github.com/asottile/pyupgrade
-    rev: v2.32.1
+    rev: v2.34.0
     hooks:
       - id: pyupgrade
         # TODO: remove `--keep-runtime-typing` option
         args: ["--py37-plus", "--keep-runtime-typing"]
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.2.0  # Use the ref you want to point at
+    rev: v4.3.0
     hooks:
       - id: mixed-line-ending
         args: ["--fix=lf"]
+  # - repo: https://github.com/pycqa/pydocstyle
+  #   rev: 6.1.1
+  #   hooks:
+  #     - id: pydocstyle
+  #       args:
+  #         - --source
+  #         - --explain
+  #         - --convention=google
+  #         - --count
+  #       additional_dependencies: ["toml"]
+  - repo: local
+    hooks:
+      - id: pyright
+        name: pyright
+        entry: pyright
+        language: node
+        pass_filenames: false
+        types: [python]
+        additional_dependencies: ["pyright"]
--- a/pettingzoo/utils/conversions.py
+++ b/pettingzoo/utils/conversions.py
@@ -260,7 +260,8 @@ class parallel_to_aec_wrapper(AECEnv):
     def step(self, action):
         if self.dones[self.agent_selection]:
             del self._actions[self.agent_selection]
-            return self._was_done_step(action)  # type: ignore
+            assert action is None
+            return self._was_done_step(action)
         self._actions[self.agent_selection] = action  # type: ignore
         if self._agent_selector.is_last():
             obss, rews, dones, infos = self.env.step(self._actions)
--- a/pettingzoo/utils/conversions.py
+++ b/pettingzoo/utils/conversions.py
@@ -257,12 +257,11 @@ class parallel_to_aec_wrapper(AECEnv):
         self.rewards[new_agent] = 0
         self._cumulative_rewards[new_agent] = 0
 
-    def step(self, action):
+    def step(self, action: None):
         if self.dones[self.agent_selection]:
             del self._actions[self.agent_selection]
-            assert action is None
             return self._was_done_step(action)
-        self._actions[self.agent_selection] = action  # type: ignore
+        self._actions[self.agent_selection] = action
         if self._agent_selector.is_last():
             obss, rews, dones, infos = self.env.step(self._actions)
 
--- a/pettingzoo/utils/conversions.py
+++ b/pettingzoo/utils/conversions.py
@@ -1,10 +1,15 @@
 import copy
 import warnings
 from collections import defaultdict
+from typing import Dict, Optional
 
 from pettingzoo.utils import agent_selector
 from pettingzoo.utils.env import AECEnv, ParallelEnv
 from pettingzoo.utils.wrappers import OrderEnforcingWrapper
+
+ActionType = Optional[int]
+AgentID = str
+ActionDict = Dict[AgentID, ActionType]
 
 
 def parallel_wrapper_fn(env_fn):
@@ -231,7 +236,7 @@ class parallel_to_aec_wrapper(AECEnv):
         self._observations = self.env.reset(seed=seed, options=options)
         self.agents = self.env.agents[:]
         self._live_agents = self.agents[:]
-        self._actions = {agent: None for agent in self.agents}
+        self._actions: ActionDict = {agent: None for agent in self.agents}
         self._agent_selector = agent_selector(self._live_agents)
         self.agent_selection = self._agent_selector.reset()
         self.dones = {agent: False for agent in self.agents}
@@ -257,9 +262,12 @@ class parallel_to_aec_wrapper(AECEnv):
         self.rewards[new_agent] = 0
         self._cumulative_rewards[new_agent] = 0
 
-    def step(self, action: None):
+    def step(self, action: ActionType):
+        if action is not None:
+            action = int(action)
         if self.dones[self.agent_selection]:
             del self._actions[self.agent_selection]
+            assert action is None
             return self._was_done_step(action)
         self._actions[self.agent_selection] = action
         if self._agent_selector.is_last():
--- a/pettingzoo/utils/conversions.py
+++ b/pettingzoo/utils/conversions.py
@@ -260,8 +260,8 @@ class parallel_to_aec_wrapper(AECEnv):
     def step(self, action):
         if self.dones[self.agent_selection]:
             del self._actions[self.agent_selection]
-            return self._was_done_step(action)
-        self._actions[self.agent_selection] = action
+            return self._was_done_step(action)  # type: ignore
+        self._actions[self.agent_selection] = action  # type: ignore
         if self._agent_selector.is_last():
             obss, rews, dones, infos = self.env.step(self._actions)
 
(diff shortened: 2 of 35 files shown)


pyproject.toml
```diff
@@ -1,16 +1,17 @@
...
-    "**/node_modules",
-    "**/__pycache__",
 ]
-strict = ["pettingzoo/utils/env.py"]
-
-reportMissingImports = true
-reportMissingTypeStubs = false
+strict = [
+]
+# reportMissingImports = true
```
jjshoots, line 11: Are these comments needed?

pettingzoo/utils/env.py
```diff
@@ -180,11 +180,14 @@ def agent_iter(self, max_iter: int = 2**63) -> AECIterable:
...
 
-    def last(self, observe: bool = True) -> Tuple[ObsType, float, bool, Dict[str, Any]]:
+    def last(
+        self, observe: bool = True
+    ) -> Tuple[Optional[ObsType], float, bool, Dict[str, Any]]:
         """
         Returns observation, cumulative reward, done, info   for the current agent (specified by self.agent_selection)
         """
         agent = self.agent_selection
+        assert agent
```
jjshoots, line 190: Is this just checking that it's not None? Is it a legitimate concern?

pettingzoo/utils/env.py
```diff
@@ -3,12 +3,12 @@
...
 
-import gym
-import numpy as np
+import gym  # type: ignore
+import numpy as np  # type: ignore
 
 ObsType = TypeVar("ObsType")
 ActionType = TypeVar("ActionType")
-AgentID = str
+AgentID = Optional[str]
```
jjshoots, line 11: I'm not too familiar with dealing with typings, but is there a reason why this is Optional and not just a str?

.pre-commit-config.yaml
```diff
@@ -28,13 +28,32 @@ repos:
...
+  # - repo: https://github.com/pycqa/pydocstyle
+  #   rev: 6.1.1
+  #   hooks:
+  #     - id: pydocstyle
+  #       args:
+  #         - --source
+  #         - --explain
+  #         - --convention=google
+  #         - --count
+  #       additional_dependencies: ["toml"]
```
jjshoots, line 50: Are these comments needed?

pettingzoo/utils/env.py
kir0ul: Yes, it removes a typing error as it makes it clear that `None` is not allowed. Do you think `None` should be allowed here?
Just for reference, using assertions is part of the tricks to help the linter choosing the right types, see for example here: https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#built-in-types=

pyproject.toml
kir0ul: This is still a work in progress, I'm not sure yet if they'll be needed or not in the end, but in any case I'll fix this before this PR is ready for review.

pettingzoo/utils/env.py
kir0ul: `Optional` means `AgentID` is also allowed to be `None`. I don't have a lot of experience with typing either but if I don't put `Optional`, I get the following error:
```
./PettingZoo/pettingzoo/utils/env.py:239:22 - error: Cannot assign member "agent_selection" for type "AECEnv"
    Expression of type "AgentID | None" cannot be assigned to member "agent_selection" of class "AECEnv"
      Type "AgentID | None" cannot be assigned to type "AgentID"
        Type "None" cannot be assigned to type "AgentID" (reportGeneralTypeIssues)
```

pettingzoo/utils/env.py
```diff
@@ -3,12 +3,12 @@
 import warnings
 from typing import Any, Dict, Iterable, Iterator, List, Optional, Tuple, TypeVar
 
-import gym
-import numpy as np
+import gym  # type: ignore
+import numpy as np  # type: ignore
```
kir0ul, line 7: I would have expected the linter to get the types from both Gym and NumPy, not so sure why these need to be ignored... :thinking:

pettingzoo/utils/env.py
jjshoots: TIL

jjshoots:
Hey @kir0ul, just checking in, what're the future plans for this PR?

kir0ul:
> Hey @kir0ul, just checking in, what're the future plans for this PR?

Hey @jjshoots, sorry I haven't forgotten about this, I'm currently attending a 3 weeks summer school, so I'll be able to resume working on this in early August. I guess the plan would be to:
1. clean up the comments,
2. fix https://github.com/Farama-Foundation/PettingZoo/pull/737#discussion_r915834884, do we want to do it the simple way or start a refactoring?
3. maybe include `pydocstyle` (I haven't made my mind yet about including it or not as there are a lot of docstrings to fix).

pettingzoo/utils/env.py
pseudo-rnd-thoughts: I believe adding this `reportMissingTypeStubs = false` to the pyproject.toml should fix this.

jjshoots:
Hey @kir0ul, if it's not terribly difficult, I'd prefer the refactoring route, although a simple assert doesn't seem like the end of the world to me either.

pettingzoo/utils/env.py
kir0ul: Yes it does, thanks!

kir0ul: Well apparently it does on my machine but not in CI... :thinking:

jjshoots:
Hey @kir0ul, this is done now?

jjshoots:
Hey @RedTachyon, could you review this when you have the time?

kir0ul:
> Hey @kir0ul, this is done now?

Hey @jjshoots, there are still a few points of discussion but I think it's ready for review.

pettingzoo/utils/env.py
RedTachyon: What exactly is the issue that requires this?

pettingzoo/utils/env.py
```diff
@@ -81,7 +81,7 @@ def seed(self, seed: Optional[int] = None) -> None:
             "Calling seed externally is deprecated; call reset(seed=seed) instead"
         )
 
-    def observe(self, agent: str) -> ObsType:
+    def observe(self, agent: str) -> Optional[ObsType]:
```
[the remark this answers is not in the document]
RedTachyon: Can we at least add something like a TODO here? This implies there's some underlying problem with the types that should be tracked down sooner or later. Making it optional is a bit of a bandaid

pettingzoo/utils/env.py
```diff
@@ -354,11 +351,10 @@ def observation_space(self, agent: AgentID) -> gym.Space:
         warnings.warn(
             "Your environment should override the observation_space function. Attempting to use the observation_spaces dict attribute."
         )
-        return self.observation_spaces[agent]
+        return self.observation_spaces[agent]  # type: ignore
```
RedTachyon, line 354: Why the ignores here? Can you add some comments around all ignores to explain them?

pyproject.toml
```diff
@@ -1,16 +1,15 @@
 [tool.pyright]
 
 # add any files/directories with type declaration to include
-include = ["pettingzoo/utils/env.py"]
+include = [
+    "pettingzoo/utils/env.py"
```
RedTachyon, line 5: For now we only want type checking for that single file?

I think it'd be better to include everything by default, and add explicit includes for each directory/file/env/whatever, so that new code is also type checked

jjshoots: If this is too big of a job, I'm more than ok with merging this PR first, and then starting new PRs for each of the other folders.

RedTachyon: Yea, my point is changing the structure of include/exclude.

When done this way, let's say someone develops a new feature for PZ, puts it in a new file or directory or anywhere really, and pushes changes there. None of these changes will be type checked in any way.

pettingzoo/utils/env.py
kir0ul: Not sure what happened exactly but now it works with @pseudo-rnd-thoughts suggestion, so I removed it.

pettingzoo/utils/env.py
kir0ul: I was able to remove almost all `# type: ignore` statements including this one. There are still 2 I wasn't able to remove, I'll make a review comment to keep them in mind.

pettingzoo/utils/conversions.py
```diff
@@ -260,8 +260,8 @@ def add_new_agent(self, new_agent):
     def step(self, action):
         if self.dones[self.agent_selection]:
             del self._actions[self.agent_selection]
-            return self._was_done_step(action)
-        self._actions[self.agent_selection] = action
+            return self._was_done_step(action)  # type: ignore
+        self._actions[self.agent_selection] = action  # type: ignore
```
kir0ul, line 264: Not sure how to remove those 2 `# type: ignore` statements, would anyone have any suggestion?

pyproject.toml
kir0ul: Since https://github.com/Farama-Foundation/PettingZoo/pull/737/commits/c07a637a3d568044172bd074a87ad880981da8d1#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711 it'll type check any file in the `pettingzoo` folder excluding the list of folders that haven't been refactored with typing yet. I also added types to files in `pettingzoo/utils/`. Does this work?

pettingzoo/utils/env.py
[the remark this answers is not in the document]
kir0ul: Right, I added a TODO in https://github.com/Farama-Foundation/PettingZoo/pull/737/commits/9c3722d1a1b6e8207f96739079c50001f608a0c4.

pettingzoo/utils/conversions.py
jjshoots: So I just checked, `_was_done_step` actually returns `None`.

kir0ul: @jjshoots I tried your suggestion and I still get the same typing error, but maybe I missed something? :thinking:

jjshoots: Do you need to do `-> None` at the function definition?

jjshoots: Could you perhaps point me to where it is? Thanks!

kir0ul: Here is is: https://github.com/Farama-Foundation/PettingZoo/pull/737/commits/511df7479770cd28ef148e8a935c31fb9ca84052.

jjshoots: I think you could just do the same thing as earlier. That whole function returns None, not sure why it was written that way instead of 
```py
xxx()
return
```

jjshoots:
@kir0ul is this ready now?

pettingzoo/utils/conversions.py
kir0ul: I tried it but it didn't work. Now I put `def step(self, action: None):` and it seems the type checker is happy with that but I honestly don't understand it as `action` is sometimes `int` or `None`... :thinking:

jjshoots: I'm not too familiar with this, but does `Optional[int]` not work?

RedTachyon: `Optional[int]` is just a synonym for `Union[int, None]`, i.e. it can be either an int or None

jjshoots: Yes but kir0ul mentioned that it was sometimes `int` or `None`, since the typechecker is happy with it being `None`, a more semantically accurate type hint would be `Optional[int]`, no?

kir0ul: I tried `Optional[int]` too but I get:
```
./pettingzoo/utils/conversions.py
  ./pettingzoo/utils/conversions.py:264:40 - error: Argument of type "int | None" cannot be assigned to parameter "action" of type "None" in function "_was_done_step"
    Type "int | None" cannot be assigned to type "None"
      Type cannot be assigned to type "None" (reportGeneralTypeIssues)
  ./pettingzoo/utils/conversions.py:265:9 - error: Argument of type "int | None" cannot be assigned to parameter "__v" of type "None" in function "__setitem__"
    Type "int | None" cannot be assigned to type "None"
      Type cannot be assigned to type "None" (reportGeneralTypeIssues)
```
Also sometimes the type of `action` is `np.int32`.

jjshoots: This seems like a really annoying case of previously not enforcing types. Is it possible for us to fix the `np.int32` thing using a type cast.

kir0ul: Done in https://github.com/Farama-Foundation/PettingZoo/pull/737/commits/55c69e687298090dd8a19a23051e042929fe65ca#diff-3ef3437e8b392d934e60182d2116f18ce516d3685c43a4b40c2a5f84c0dfe93fR267.

kir0ul:
@jjshoots @RedTachyon I think this is ready for another round of review.

jjshoots:
@RedTachyon do you mind giving this the next pass?

jjshoots:
Optionally, could you give this a review? @pseudo-rnd-thoughts

pseudo-rnd-thoughts:
I know it is a pain, but could we split this PR into two as it is two separate things being done. It will be easier to review that way.

jjshoots:
I personally don't think that's such a great idea, the time spent splitting it would be non-trivial, but if @kir0ul doesn't mind then I'm ok with it.

kir0ul:
Conflicts have been fixed and after talking with @jjshoots we decided not to split this PR in two.

jjshoots:
Thanks a ton for the help everyone!

Issue #732, opened 2022-06-23
[Proposal] Change MPE rendering to use Pygame instead of pyglet


jjshoots:
### Proposal

Currently MPE uses the outdated rendering engine Pyglet. It should be changed to Pygame as is done with the other environments.

WillDudley:
Will look into this.

WillDudley:
See PR at https://github.com/Farama-Foundation/PettingZoo/pull/738

PR #738, opened 2022-07-06
 [Proposal] Change MPE rendering to use Pygame instead of pyglet #732 

WillDudley:
# Description

This PR makes MPE environments use Pygame rather than Pyglet. As a result of the simplicity of using Pygame instead of Pyglet, there seems to be no use of rendering.py (many methods were unused, I'm not sure if we want to keep them). 

There is an outstanding issue of: On some machines, there is an initial black screen displayed when `pygame.display.set_mode(self.screen.get_size())` is called. I'm out of time to debug this further, but the PR is still functional.

Addresses https://github.com/Farama-Foundation/PettingZoo/issues/732 

## Type of change

### Screenshots

# Checklist:

- [ ] I have run `pytest -v` and no errors are present. **The same errors are present before and after the proposed changes**
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have added tests that prove my fix is effective or that my feature works

The following code was used to sanity-check this PR:

```
import time

from pettingzoo.mpe import simple_v2, simple_adversary_v2, simple_crypto_v2, simple_push_v2, simple_reference_v2, \
    simple_speaker_listener_v3, simple_spread_v2, simple_tag_v2, simple_world_comm_v2

env_objects = [simple_v2, simple_adversary_v2, simple_crypto_v2, simple_push_v2, simple_reference_v2, \
    simple_speaker_listener_v3, simple_spread_v2, simple_tag_v2, simple_world_comm_v2]

for env_object in env_objects:
    env = env_object.env()
    env.reset()
        env.render()
        env.step(env.action_space(agent).sample() if not env.dones[agent] else None)
        time.sleep(0.1)
    env.close()
```


pettingzoo/mpe/_mpe_utils/simple_env.py
```diff
@@ -231,75 +246,80 @@ def step(self, action):
...
+    def render(self, mode="human"):
+        if not self.renderOn and mode == "human":
+            # sets self.renderOn to true and initializes display
+            self.enable_render()
+        elif mode == "human":
+            self.draw()
+
+        observation = np.array(pygame.surfarray.pixels3d(self.screen))
+        if mode == "human":
+            pygame.display.flip()
```
okgarces, line 263: Hi there,

Easy and well written code. I could read really quick.

I would suggest small changes to improve a little bit readability, really small changes.

```suggestion
        self.enable_render()

        observation = np.array(pygame.surfarray.pixels3d(self.screen))
        if mode == "human":
            self.draw()
            pygame.display.flip()
```

pettingzoo/mpe/_mpe_utils/simple_env.py
```diff
@@ -231,75 +246,80 @@ def step(self, action):
...
-            if np.all(other.state.c == 0):
-                word = "_"
-            elif self.continuous_actions:
-                word = "[" + ",".join([f"{comm:.2f}" for comm in other.state.c]) + "]"
-            else:
-                word = alphabet[np.argmax(other.state.c)]
+    def enable_render(self):
+        self.screen = pygame.display.set_mode(self.screen.get_size())
+        self.renderOn = True
+        self.draw()
```
okgarces, line 252: The self.draw could be removed from `enable_render` function. That way `enable_render` would have one specific responsibility. But I would include the flag to render or no depends if the mode is human.

```suggestion
    def enable_render(self, mode="human"):
        if not self.renderOn and mode == "human":
            self.screen = pygame.display.set_mode(self.screen.get_size())
            self.renderOn = True
```
If in the future you need to enable render for multiple modes you only changes in one place.

What do you think?

okgarces (commented):
Code LGTM and I ran a small test, which you suggested. 
We can discuss the comments, but looks great!

pettingzoo/mpe/_mpe_utils/simple_env.py
WillDudley: Thanks for the compliment! I normally just write code for my eyes only so it means a lot :)

The style of this code was inspired by cooperative_pong, which admittedly I haven't had time to fully understand. I'll quote the code from cooperative_pong below:

```
    def enable_render(self):
        self.screen = pygame.display.set_mode(self.screen.get_size())
        self.renderOn = True
        self.draw()

    def render(self, mode="human"):
        if not self.renderOn and mode == "human":
            # sets self.renderOn to true and initializes display
            self.enable_render()

        observation = np.array(pygame.surfarray.pixels3d(self.screen))
        if mode == "human":
            pygame.display.flip()
        return (
            np.transpose(observation, axes=(1, 0, 2)) if mode == "rgb_array" else None
        )
```

The addition of lines 258-259 was to compensate for the lack of a `self.draw()` call in the `step()` method.

I agree that your refactoring is more elegant, and I would be in favour of these changes. I just thought I'd note that it is a minor divergence from the style of the Butterfly environments, it seems.

pettingzoo/mpe/_mpe_utils/simple_env.py
WillDudley: See reply below, same thoughts apply.

pettingzoo/mpe/_mpe_utils/simple_env.py
jjshoots: Having the shorter version of the code is probably ideal, cooperative pong can be changed to reflect this more elegant way of doing things down the line.

okgarces: Nice! Always we can improve to make everything better. In the case of cooperative_pong, the code is quite similar, we can abstract the code and reuse it, but this should be well designed and planned. 
Thanks guys for the discussion.

okgarces (approved):
Thanks for accepting the suggestions

WillDudley:
Error seems unrelated to my proposed changes. It's not clear to me what  raised the error.

jjshoots:
@WillDudley Someone probably cancelled the remaining test seeing as everything else has passed. In either cases, LGTM. Thanks for the contribution!

PR #745, opened 2022-07-31
Account for "render_modes" or "render.modes" in metadata 

timf34:
# Description

I ran into a bug wherein my environments where the metadata's keys for  my environment (which is wrapped as a `SB3VecEnvWrapper`) where not "render_modes" but "render.modes". And so I made a quick fix as can be seen below. 

I have not tested this extensively, however it seems a pretty inconsequential quick fix that might be helpful to others. 

## Type of change


jjshoots:
Hi @timf34, could you describe which env you're using and where this `render.modes` is appearing? This does not look like intended behaviour and I think the right solution is to fix the environment to use `render_modes` instead of `render.modes`.

As an aside, would you like to be onboarded to be one of our maintainers for PZ?

jjshoots:
Closing because while this is a bug, all render APIs will change down the line in the near future.

Issue #746, opened 2022-08-01
[Bug Report] Broken atari environment path


manu-hoffmann:
Hi!

I ran into the [same issue](https://github.com/Farama-Foundation/PettingZoo/issues/744) with PettingZoo version 1.19.
Found the issue to be a broken path in file `pettingzoo/atari/pong/pong.py`.

I fixed the issue by replacing line 37:

`parent_file = glob(os.path.join(os.path.dirname(os.path.dirname(__file__)), name + "*.py"))`

I don't think this is the best solution, but did the job for me.

Thanks.

jjshoots:
Hi @manu-hoffmann, yes this is a currently known issue in PettingZoo which is why there has not been a new release yet. The developer who's supposed to fix this is kinda busy, if you would be so inclined, could you make a PR fixing this?

manu-hoffmann:
Hi @jjshoots,

I just opened a PR for fixing this issue.

PR #747, opened 2022-08-01
fix broken atari environment path

manu-hoffmann:
Fix the broken atari environment path as described [here](https://github.com/Farama-Foundation/PettingZoo/issues/746).

Type of PR
bug fix

### DIFF
--- a/docs/classic/backgammon.md
+++ b/docs/classic/backgammon.md
@@ -33,7 +33,7 @@ The rules of backgammon can be found [here.](https://www.bkgm.com/rules.html)
 
 ### Observation Space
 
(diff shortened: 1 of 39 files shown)


jjshoots:
@manu-hoffmann if it's not too much work, would you mind doing it for the other atari environments too? The structure of things should be the same.

manu-hoffmann:
Hi @jjshoots,

I've fixed the path issue for all atari environments, pulled from master and executed `black .`.

This should solve all remaining issues with the PR.

Thanks and have a nice day.

jjshoots:
Thanks for the contribution!

PR #758, opened 2022-08-12
Truncation Update for Atari Envs

jjshoots:
# Description

This updates all Atari envs to have the updated truncation fix according to Gym's newer API.


pseudo-rnd-thoughts:
LGTM

benblack769:
Seems reasonable enough. I assume this doesn't work with all the wrappers yet?

jjshoots:
@benblack769 Yeap, it doesn't work with wrappers yet, this is probably one of the updates that requires a long list of PRs to be merged into an alternative branch before they go into master.

PR #774, opened 2022-08-22
Classic Envs Update

reginald-mclean:
This update removes the use of self.dones in the Classic envs, replaces self.dones with self.terminations and self.truncations, and changes references from __was_done_step to _was_dead_step. There is also an update to pettingzoo/utils/wrappers/terminate_illegal.py that removes self.dones and uses self.terminations and self.truncations

### DIFF
--- a/pettingzoo/classic/chess/chess.py
+++ b/pettingzoo/classic/chess/chess.py
@@ -54,6 +54,8 @@ class raw_env(AECEnv):
 
         self.rewards = None
         self.infos = {name: {} for name in self.agents}
+        self.truncations = {name: False for name in self.agents}
+        self.terminations = {name: False for name in self.agents}
 
         self.agent_selection = None
 
@@ -101,7 +103,6 @@ class raw_env(AECEnv):
     def set_game_result(self, result_val):
         for i, name in enumerate(self.agents):
             self.terminations[name] = True
-            self.truncations[name] = True
             result_coef = 1 if i == 0 else -1
             self.rewards[name] = result_val * result_coef
             self.infos[name] = {"legal_moves": []}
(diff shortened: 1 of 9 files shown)


pettingzoo/classic/chess/chess.py
```diff
@@ -53,7 +53,6 @@ def __init__(self):
         }
 
         self.rewards = None
-        self.dones = None
```
jjshoots, line 56: Is there a reason this was removed but not replaced with terminations and truncations?

pettingzoo/classic/chess/chess.py
```diff
@@ -93,21 +92,23 @@ def reset(self, seed=None, return_info=False, options=None):
...
+        self.truncations = {name: False for name in self.agents}
         self.infos = {name: {} for name in self.agents}
 
         self.board_history = np.zeros((8, 8, 104), dtype=bool)
 
     def set_game_result(self, result_val):
         for i, name in enumerate(self.agents):
-            self.dones[name] = True
+            self.terminations[name] = True
+            self.truncations[name] = True
```
jjshoots, line 104: This should be True for terminations.

jjshoots (commented):
Hi @reginald-mclean, overall LGTM, aside from the couple of comments I've left. Could you also run `pip3 install black && black .` on the root directory once you're done with the fixes? This fixes the code formatting.

jjshoots:
Hey @WillDudley, I've given this a first pass review, but do you mind giving it a second pass?

WillDudley:
Looking at this now :)

WillDudley (approved):
lgtm :)

Note that the RLCard envs still need to be refactored by someone.

jjshoots:
@WillDudley Thanks for the review, I'll have a look at RLCard in the next couple of days.
Thanks for the contribution @reginald-mclean!  (On a side note if you'd like to work on RLCard, do lmk :))

WillDudley:
@jjshoots just looking at RLCard now, give me 30 mins

PR #775, opened 2022-08-24
RLCard refactor

WillDudley:
There's some issue here regarding handling of all-zero action masks, several tests raise this whenever `np.random.choice(np.flatnonzero(obs['action_mask'])` is called, or  something like that

### DIFF
--- a/pettingzoo/classic/rlcard_envs/rlcard_base.py
+++ b/pettingzoo/classic/rlcard_envs/rlcard_base.py
@@ -101,7 +101,7 @@ class RLCardBase(AECEnv):
             )
             self.next_legal_moves = []
             self.terminations = self._convert_to_dict(
-                [False for _ in range(self.num_agents)]
+                [True for _ in range(self.num_agents)]
             )
             self.truncations = self._convert_to_dict(
                 [False for _ in range(self.num_agents)]
--- a/pettingzoo/classic/rlcard_envs/rlcard_base.py
+++ b/pettingzoo/classic/rlcard_envs/rlcard_base.py
@@ -87,8 +87,11 @@ class RLCardBase(AECEnv):
         return {"observation": observation, "action_mask": action_mask}
 
     def step(self, action):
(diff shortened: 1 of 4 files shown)


jjshoots:
LGTM.

pettingzoo/classic/rlcard_envs/rlcard_base.py
```diff
@@ -117,7 +123,12 @@ def reset(self, seed=None, return_info=False, options=None):
         self._cumulative_rewards = self._convert_to_dict(
             [0 for _ in range(self.num_agents)]
         )
-        self.dones = self._convert_to_dict([False for _ in range(self.num_agents)])
+        self.terminations = self._convert_to_dict(
+            [False for _ in range(self.num_agents)]
+        )
+        self.truncations = self._convert_to_dict(
+            [False for _ in range(self.num_agents)]
+        )
```
jjshoots, line 131: Found the issue, the environment was not terminating.

jjshoots:
All tests seem to pass now, could you briefly review and lmk what you think? @WillDudley

WillDudley:
All tests pass for me too.

Just have one Q: Are we ok with RLCard envs never truncating? For example, in lucholdem we have
```
    def is_over(self):
        ''' Check if the game is over

        Returns:
            (boolean): True if the game is over
        '''
        alive_players = [1 if p.status=='alive' else 0 for p in self.players]
        # If only one player is alive, the game is over.
        if sum(alive_players) == 1:
            return True

        # If all rounds are finshed
        if self.round_counter >= 2:
            return True
        return False
```

Should we return truncated if self.round_counter >= 2? I'm leaning towards yes but I'm rather unfamiliar with the RLCard envs. I can do this refactor in a few days if we want - got a conf starting now.

Otherwise, LGTM.

jjshoots:
I'm not too familiar with RLCard envs either, but good catch, I think that would be a needed implementation.

PR #779, opened 2022-09-04
Pull Prospector

jjshoots:
This pulls prospector and all related references throughout the library.


jjshoots:
@pseudo-rnd-thoughts could you review this, should be fairly easy.

pseudo-rnd-thoughts (approved):
This looks good but could you add a comment on why it is being removed

jkterry1:
The environment doesn't serve an interesting use, has had large bugs since inception, has unfixable horrible code and you can't learn any interesting behaviors in it

PR #788, opened 2022-09-16
Update Gym version

WillDudley:
Keeping as draft for now


jkterry1:
OpenAI hasn't cut 0.26.1 yet, and it's a super minor bugfix release anyways on top of 0.26.0. You can just use >=0.26.0 if you want

PR #800, opened 2022-09-23
Update new Render API

younik:
Update to new Render API https://github.com/openai/gym/pull/2671

Todo:
 - [x] make `human` rendering automatic


docs/code_examples/parallel_rps.py
```diff
@@ -24,15 +25,16 @@
...
     The env function often wraps the environment in wrappers by default.
     You can find full documentation for these methods
     elsewhere in the developer documentation.
     """
-    env = raw_env()
+    env = raw_env(render_mode=render_mode if render_mode != "ansi" else "human")
     # This wrapper is only for environments which print results to the terminal
-    env = wrappers.CaptureStdoutWrapper(env)
+    if render_mode == "ansi":
+        env = wrappers.CaptureStdoutWrapper(env)
```
jjshoots, line 37: Line 37 won't be triggered at all.

pettingzoo/atari/base_atari_env.py
```diff
@@ -256,10 +271,8 @@ def render(self, mode="human"):
             self._screen.blit(myImage, (0, 0))
 
             pygame.display.flip()
-        elif mode == "rgb_array":
+        else:  # self.render_mode == "rgb_array":
```
jjshoots, line 274: Should the `rgb_array` conditional be here? Seems like that mode is not being accounted for unless I'm missing someething.

pettingzoo/butterfly/pistonball/pistonball.py
```diff
@@ -593,19 +596,27 @@ def get_local_reward(self, prev_position, curr_position):
...
-    def render(self, mode="human"):
-        if mode == "human" and not self.renderOn:
+    def render(self):
+        if self.render_mode is None:
+            gym.logger.WARN(
+                "You are calling render method without specifying any render mode."
+            )
+            return
+
+        if self.render_mode == "human" and not self.renderOn:
```
jjshoots, line 606: I see these lines a lot and I'm wondering if this should be in the main `AECEnv` and `ParallelEnv` class itself. Thoughts?

pettingzoo/classic/connect_four/connect_four.py
```diff
@@ -101,14 +102,15 @@ class raw_env(AECEnv):
...
-    def __init__(self):
+    def __init__(self, render_mode=None):
         super().__init__()
         # 6 rows x 7 columns
         # blank space = 0
         # agent 0 -- 1
         # agent 1 -- 2
         # flat representation in row major order
         self.screen = None
+        self.render_mode = render_mode
```
jjshoots, line 113: This line too, moving this into the superclasses would probably make sense...

pettingzoo/classic/connect_four/connect_four.py
```diff
@@ -217,16 +222,22 @@ def reset(self, seed=None, return_info=False, options=None):
...
         screen_height = 1118
         if self.screen is None:
-            if mode == "human":
+            if self.render_mode == "human":
                 pygame.init()
                 self.screen = pygame.display.set_mode((screen_width, screen_height))
             else:
                 self.screen = pygame.Surface((screen_width, screen_height))
-        if mode == "human":
+        if self.render_mode == "human":
```
jjshoots, line 240: Could we fix this loop to:

```py
if self.render_mode == "human":
  if self.screen isi None:
    # do stuff
  else:
    # do stuff
else:
  #do stuff
```

pettingzoo/classic/hanabi/hanabi.py
```diff
@@ -196,8 +197,14 @@ def __float__(self):
...
+    render_mode = kwargs.get("render_mode")
+    if render_mode == "ansi":
+        kwargs["render_mode"] = "human"
+        env = r_env = raw_env(**kwargs)
+        env = wrappers.CaptureStdoutWrapper(env)
+    else:
+        env = r_env = raw_env(**kwargs)
+
     env = wrappers.TerminateIllegalWrapper(
         env, illegal_reward=HanabiScorePenalty(r_env)
```
jjshoots, line 209: This is not your doing, but the fact that `r_env` is needed here kinda scares me. Does anyone even know what this is doing? Don't `env` and `r_env` just point to the same object?

pettingzoo/classic/rlcard_envs/leduc_holdem.py
```diff
@@ -93,8 +94,13 @@
...
 def env(**kwargs):
-    env = raw_env(**kwargs)
-    env = wrappers.CaptureStdoutWrapper(env)
+    render_mode = kwargs.get("render_mode")
+    if render_mode == "ansi":
+        kwargs["render_mode"] = "human"
+        env = raw_env(**kwargs)
+        env = wrappers.CaptureStdoutWrapper(env)
+    else:
+        env = raw_env(**kwargs)
```
jjshoots, line 103: This seems like a better way of doing that funky thing where line 37 won't run earlier.

pettingzoo/classic/rlcard_envs/texas_holdem.py
```diff
@@ -155,13 +163,13 @@ def calculate_height(screen_height, divisor, multiplier, tile_size, offset):
...
         if self.screen is None:
-            if mode == "human":
+            if self.render_mode == "human":
                 pygame.init()
                 self.screen = pygame.display.set_mode((screen_width, screen_height))
             else:
                 pygame.font.init()
                 self.screen = pygame.Surface((screen_width, screen_height))
-        if mode == "human":
+        if self.render_mode == "human":
```
jjshoots, line 172: Any possibility this can be a nicer looking for loop?

pettingzoo/classic/rlcard_envs/texas_holdem_no_limit.py
```diff
@@ -192,13 +201,13 @@ def calculate_height(screen_height, divisor, multiplier, tile_size, offset):
...
         if self.screen is None:
-            if mode == "human":
+            if self.render_mode == "human":
                 pygame.init()
                 self.screen = pygame.display.set_mode((screen_width, screen_height))
             else:
                 pygame.font.init()
                 self.screen = pygame.Surface((screen_width, screen_height))
-        if mode == "human":
+        if self.render_mode == "human":
```
jjshoots, line 210: Nicer looking for loop thing here too.

pettingzoo/classic/rps/rps.py
```diff
@@ -232,13 +240,13 @@ def offset(i, size, offset=0):
...
         if self.screen is None:
-            if mode == "human":
+            if self.render_mode == "human":
                 pygame.init()
                 self.screen = pygame.display.set_mode((screen_width, screen_height))
             else:
                 pygame.font.init()
                 self.screen = pygame.Surface((screen_width, screen_height))
-        if mode == "human":
+        if self.render_mode == "human":
```
jjshoots, line 249: Nicer looking for loop thing.

pettingzoo/utils/env.py
```diff
@@ -86,10 +86,11 @@ def observe(self, agent: str) -> Optional[ObsType]:
         """
```
jjshoots, line 86: Just leaving this comment here that maybe `self.render_mode` should be added into this superclass.

jjshoots (commented):
Almost LGTM so far, other than some of the comments I've left.

I think self.render_mode and possibly the warning can be moved into env.py itself, this would remove a lot of clutter from the many environment files.

docs/code_examples/aec_rps.py
```diff
@@ -25,15 +26,16 @@
...
     The env function often wraps the environment in wrappers by default.
     You can find full documentation for these methods
     elsewhere in the developer documentation.
     """
-    env = raw_env()
+    env = raw_env(render_mode=render_mode if render_mode != "ansi" else "human")
     # This wrapper is only for environments which print results to the terminal
-    env = wrappers.CaptureStdoutWrapper(env)
+    if render_mode == "ansi":
+        env = wrappers.CaptureStdoutWrapper(env)
```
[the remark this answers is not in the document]
younik: In line 35 there is no assignment, but just a parameter passing, thus render_mode can still be "ansi", isn't it?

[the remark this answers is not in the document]
WillDudley: So `env.render_mode` is `human` if `render_mode` is `ansi`, otherwise `env.render_mode` is whatever `render_mode` is?

Makes sense after a couple minutes of thought, but it is a bit confusing. Gimme some minutes to review this to convince myself this is the most readable way.

pettingzoo/atari/base_atari_env.py
younik: Right now, gym.make checks that render_mode is in `metadata['render_modes']` (if it is not None), thus they should be equivalent.
Anyway, for clarity it could be better to keep what you suggested or to put an assert, what do you think?

docs/code_examples/aec_rps.py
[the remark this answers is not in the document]
WillDudley: Hmm, ok. Maybe `render_mode="human" if render_mode=="ansi" else render_mode` would be more readable, along with an inline comment of something like `# switch "ansi" to "human"'`.

pettingzoo/classic/hanabi/hanabi.py
WillDudley: @jjshoots Nope, they don't.

image

younik: I have no idea why there are two variable there, but yes, they point to the same object (@WillDudley you should try with an object)

pettingzoo/butterfly/pistonball/pistonball.py
younik: Yes, it would be nice to avoid repetitions
However, if I put it on the superclass with every subclass calling super().render(), then `return` will not be effective to exit from render (it will exit only from the super().render())
The second small reason is that `AECEnv` and `ParallelEnv` are supposed to be abstract classes and `render` an abstract method. For example, if while writing a custom environment you forgot to implement `render`, PyCharm will highlight it, and calling render will raise an error (instead of doing nothing).

docs/code_examples/aec_rps.py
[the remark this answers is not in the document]
jjshoots: Yeah basically render mode will never be ansi, because it gets converted to human otherwise

pettingzoo/atari/base_atari_env.py
jjshoots: Yes, but isn't this conditional in `def render()`? If that's the case, shouldn't it return the image only if `rgb_array` is triggered? Right now it seems that `image` is only returned if render mode is `human`, otherwise it does nothing. Unless I'm not seeing the code properly that is, which is entirely possible.

pettingzoo/butterfly/pistonball/pistonball.py
jjshoots: This is a fair argument, so we are only asserting that _our_ environments have a `render_mode`, and the user-created envs don't have this requirement? If this is the case, this can be resolved. Paging @pseudo-rnd-thoughts for clarity.

pettingzoo/classic/hanabi/hanabi.py
jjshoots: Yeah with objects they usually do. Python is black magic.

jjshoots: That said, could we try removing the need for r_env? I've looked through the code several times and can't see any reason it would break without it... but Python is black magic so ¯\_(ツ)_/¯

WillDudley: Oh that's interesting and not at all confusing (: 

I assumed (because I thought `env` and `r_env` pointed to different objects), that `r_env` was so that they could do `HanabiScorePenalty(r_env)`, where `r_env` isn't wrapped by `CaptureStdoutWrapper()`. Genuinely doesn't seem to be any other reason for it.

The git blame points to Ben Black so we could ask him, or we could just remove `r_env` entirely (indifferent to which).

docs/code_examples/aec_rps.py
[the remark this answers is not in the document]
WillDudley: I took it to mean more like:

```
if render_mode == "ansi":
  env.render_mode = "human"
else:
  env.render_mode = render_mode
```

But either way it's difficult to read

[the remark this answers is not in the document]
jjshoots: Ahhh I see it now, `env.render_mode` and `render_mode` are different things. @younik Could we turn this into something a bit more readable perhaps?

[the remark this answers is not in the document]
younik: Sure, I changed it a little, let me know if this is better

pettingzoo/classic/hanabi/hanabi.py
younik: I removed it; tests are still ok and I am pretty confident saying now it is equivalent to before

pettingzoo/atari/base_atari_env.py
younik: Right now, the code displays the image in case of `human` mode otherwise returns the image.
Given the check performed by `gym.make`, this code should work equally (except if someone changes `render_mode` on the fly that should not be done, and in that case both code performs poorly).
Anyway, for clarity, I restored the previous code

pettingzoo/butterfly/pistonball/pistonball.py
pseudo-rnd-thoughts: I guess it depends on how backward compatible you want to be. 
I may be missing something, but will this check only be run in PZ envs currently

jjshoots (commented):
LGTM now, if @WillDudley is happy with this then I think we can merge. (Possibly after the clarification needed from Mark)

pettingzoo/butterfly/pistonball/pistonball.py
younik: @jjshoots The good practice for users is to always define it, because Wrapper, for example, can use this attribute. 
I can assign `render_mode=None` in the superclasses `AECEnv` and `ParallelEnv`, the environments should override it. 

Btw, I noticed this is not done for metadata, for example (that also should be defined). Do you know why? See https://github.com/openai/gym/blob/e055ee7b8e7a95d5cb41c2cb5bfd3603c67b516c/gym/core.py#L62 for example

jjshoots: @pseudo-rnd-thoughts Yes, right now it's only for PZ envs.
@younik Personally, I'm inclined to not have it defined in the superclasses if this is something that we're not enforcing in the superclasses.

pseudo-rnd-thoughts: I think this is fine, this might be solved in PZ2 but we can make a decision then about it

jjshoots: Ok, marking this as resolved then.

docs/code_examples/aec_rps.py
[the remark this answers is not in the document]
jjshoots: Having `internal_render_mode` seems somewhat unelegant, but fits the purpose well enough so I think that's ok.

WillDudley (approved):
lgtm :)

I've pinged Ben on Discord asking for his sanity check on the `r_env` revision if he has time, but nonetheless I approve the merge. If Ben notices we've misinterpreted it we can easily hotfix it.

jjshoots:
Cool, @younik could you fix `black`? Then I'll merge this PR, thanks a ton for the help!

pettingzoo/classic/hanabi/hanabi.py
benblack769: @WillDudley They do point to the same object. Integers are immutable, so += is not an in-place operation for integers.

```
>>> a = b = []
>>> a.append(1)
>>> b
[1]
>>>
```

benblack769: The reason for this weird double assignment is that we often found ourselves needing to refer to the original raw environment `r_env` for various environment specific attributes when creating some of the wrappers. And the `env` variable would get clobbered. This is before we had the `unwrapped` attribute.

This pattern got copied here, not sure if it was used in this way.

benblack769:
Looking at the r_env thing, it looks necessary to me:

```
        env = raw_env(**kwargs)
        env = wrappers.CaptureStdoutWrapper(env)
    else:
        env = raw_env(**kwargs)

    env = wrappers.TerminateIllegalWrapper(
        env, illegal_reward=HanabiScorePenalty(env)
    )
```

Note the code path:

```
        env = raw_env(**kwargs)
        env = wrappers.CaptureStdoutWrapper(env)
        HanabiScorePenalty(env)
```

`HanabiScorePenalty` looks at the `hanabi_env` attribute, which is part of the raw env, but not part of the wrapped env. If you add an unwrapped, i.e.

```
        env = raw_env(**kwargs)
        env = wrappers.CaptureStdoutWrapper(env)
        HanabiScorePenalty(env.unwrapped)
```

It should work

younik:
Thanks, @benblack769, it makes sense! With this PR, I also added a custom implementation of `__getattr__` for wrappers. Thus, the following code runs fine:
```python
from pettingzoo.classic.hanabi_v4 import raw_env
from pettingzoo.classic.hanabi.hanabi import HanabiScorePenalty
from pettingzoo.utils import wrappers

env = raw_env(render_mode="human")
env = wrappers.CaptureStdoutWrapper(env)
env = HanabiScorePenalty(env)
float(env)
```

@jjshoots I fixed pre-commit.

pettingzoo/utils/wrappers/base.py
```diff
@@ -39,6 +39,12 @@ def __init__(self, env):
         except AttributeError:
             pass
 
+    def __getattr__(self, name):
```
younik, line 42: @benblack769 for reference

pettingzoo/utils/wrappers/base.py
```diff
@@ -39,6 +39,12 @@ def __init__(self, env):
...
+        """Returns an attribute with ``name``, unless ``name`` starts with an underscore."""
+        if name.startswith("_"):
+            raise AttributeError(f"accessing private attribute '{name}' is prohibited")
+        return getattr(self.env, name)
+
```
younik, line 47: @benblack769 for reference

WillDudley:
Personally I don't quite follow this. Why does adding an attribute error for hidden attributes negate the need for `env.unwrapped`?

jjshoots:
It's not the error that negates the need for unwrapped, it's the `getattr(self.env, name)`, that's basically unwrapped.

WillDudley:
Ah, I see, lgtm then. Merge?

jjshoots:
I'm waiting to see if @benblack769 has anything to say about that new getattr.

jjshoots:
@benblack769 Could you have a look at that getattr thing and see if it's reasonable?

pettingzoo/utils/wrappers/base.py
benblack769: I wouldn't add it casually in the middle of a big PR like this, its worth thinking over whether this plays nicely with the conversion wrappers, with the parallel wrappers, with the OrderEnforcing wrapper which has its own `__getattr__` overload, etc.

PZ has somewhat more complex wrapper system than gym.

That being said, I think people have converted me over to the concept that this is a good idea.

jjshoots:
Thanks for the help everyone! Merging since I think everything's sound.

Trinkle23897:
Hi there, a lot of people complain about the API changes you have made. Is there a decent solution to address those issues?

younik:
> Hi there, a lot of people complain about the API changes you have made. Is there a decent solution to address those issues?

Hello,
We changed the API of rendering in https://github.com/openai/gym/pull/2671
This became a breaking change in gym 0.26, then we migrate also PettingZoo. 

I explained the changes and how to update environments here: https://younis.dev/blog/render-api/
I am happy to help more with it if needed.

WillDudley:
We also have a tutorial but it needs some minor bugfixes which I don't have capacity to do. 

The tutorials can be found here: https://github.com/Farama-Foundation/PettingZoo/tree/master/tutorials/Tianshou

@Trinkle23897 It would help if you could take a look at them, otherwise I'll find someone to delegate the task to. I'll also bump up the task priority, I wasn't aware of the complaints.

Trinkle23897:
It would be nice if someone can address this issue!

WillDudley:
We've upped the priority on the tutorial

PR #817, opened 2022-10-08
Docs Update 2

mgoulao:
# Description

- [X] Update docs to Gymnasium 
- [x] AEC Env attribute should have the type hints with the variables https://gymnasium.farama.org/api/core/#gymnasium.Env.action_space
- [x] AEC Env split attributes and functions on the content table to the right. See gymnasium core
- [x] Parallel content table Example usage is greater than ParallelEnv
- [x] PettingZoo wrappers should have the init of the wrappers I think
- [x] Supersuit wrappers should have content table with all of the wrappers

[https://mgoulao.github.io/PettingZoo](https://mgoulao.github.io/PettingZoo)

### DIFF
--- a/docs/api/core.md
+++ b/docs/api/core.md
@@ -63,13 +63,13 @@
 
         A dict of the observation spaces of every agent, keyed by name. This cannot be changed through play or resetting.
 
-        :type: Dict[AgentID, gym.spaces.Space]
+        :type: Dict[AgentID, gymnasium.spaces.Space]
 
     .. py:attribute:: action_spaces
 
         A dict of the action spaces of every agent, keyed by name. This cannot be changed through play or resetting.
 
-        :type: Dict[AgentID, gym.spaces.Space]
+        :type: Dict[AgentID, gymnasium.spaces.Space]
 
     .. automethod:: step
     .. automethod:: reset
--- a/docs/api/core.md
+++ b/docs/api/core.md
@@ -1,5 +1,7 @@
 # Core API
 
+## AECEnv
+
 ```{eval-rst}
 .. currentmodule:: pettingzoo.utils.env
 
@@ -7,7 +9,7 @@
 
 ```
 
-## Attributes
+### Attributes
 
 
 ```{eval-rst}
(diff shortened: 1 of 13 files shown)


docs/api/core.md
```diff
@@ -5,79 +5,89 @@
...
+.. autoattribute:: AECEnv.agents
 
-        A list of the names of all current agents, typically integers. These may be changed as an environment progresses (i.e. agents can be added or removed).
+    A list of the names of all current agents, typically integers. These may be changed as an environment progresses (i.e. agents can be added or removed).
 
-        :type: list[AgentID]
+    :type: List[AgentID]
 
-    .. py:attribute:: num_agents
+.. autoattribute:: AECEnv.num_agents
```
WillDudley, line 21: Why do types sometime appear and sometimes not?

docs/api/core.md
```diff
@@ -5,79 +5,89 @@
...
 
-        The length of the agents list.
+    The length of the agents list.
 
-        :type: int
+    :type: int
```
WillDudley, line 25: Don't need these

docs/api/core.md
```diff
@@ -5,79 +5,89 @@
 
 .. autoclass:: AECEnv
```
WillDudley, line 6: Why, in the right hand side navbar, is AECEnv in code format? The navbar format is also still different from https://gymnasium.farama.org/api/core/#

WillDudley: This is what was meant  by " AEC Env split attributes and functions on the content table to the right. See gymnasium core"

WillDudley (changes requested):

docs/api/core.md
mgoulao: The code is missing typehints.

docs/api/core.md
mgoulao: ok

docs/api/core.md
mgoulao: > Why, in the right hand side navbar, is AECEnv in code format?
Because they were generated using .. autoclass:: , ..py:function and etc. 

>  The navbar format is also still different from

I'll change the headings then

docs/api/core.md
mgoulao: Should I keep `List[AgentID]` which is more informative than `List[str]`?

WillDudley (approved):

PR #835, opened 2022-10-22
Fix pre-commit

jjshoots:
# Description

This _was_ the sequence space implementation for KAZ, but somehow the PR got merged somewhere, so now this just fixed precommit.


RedTachyon:
Do I understand correctly that this graph would have no edges? Maybe it would be more suitable to use the `Sequence` space from the newer versions of gymnasium?

jjshoots:
The original idea was for it to be one node (the agent itself) and n edges for the n entities that it can observe. But the Graph space doesn't allow edges that connect to nothing, so it's all just nodes.

I thought about sequence space, but semantically it seems kind of weird - a sequence represents a sequence of objects, like words in a sentence, whereas a graph doesn't have order to it.

But you're right, maybe a sequence space is better.

Issue #828, opened 2022-10-14
[Proposal] Classic env GUI upgrade


WillDudley:
We wish to upgrade the rendering of some of the classic environments to come in line with our standards (see https://pettingzoo.farama.org/environments/classic/).

- [x] Gin Rummy should be rendered in the same style as Texas Holdem
- [x] Leduc Holdem should be rendered in the same style as Texas Holdem
- [ ] Mahjong should be rendered using https://gamblemountain.itch.io/riichi-asset
- [x] Chess should be rendered using https://byandrox.itch.io/pixchess (reach out to Jordan for funds)
- [ ] Hanabi should be rendered using [TBC]
- [x] TicTacToe should be rendered using [TBC]

Notes:

- Usage terms of the asset packs should be double-checked
- The option to render Chess, Hanabi and TicTacToe without any dependencies/assets should be kept as a non-default option (`render_mode='ascii'`) or something

PR #842, opened 2022-10-30
ENH: add gui to chess

younik:
This PR adds GUI to the chess environment. There is also a fix for warnings when `render_mode` is None.

https://github.com/Farama-Foundation/PettingZoo/issues/828

Screenshot from 2022-10-30 18-20-57


WillDudley (approved):
This is awesome! Thanks a lot :)

PR #847, opened 2022-10-31
change paradigm of parallel api loops

WillDudley:
I didn't like how we used `for step in range(max_cycles)` before, now that we have `truncation`.


WillDudley:
@jjshoots This makes sense right?

jjshoots (approved):
Yes, this makes sense to me.

PR #870, opened 2022-12-22
wrappers documentation: get `to_parallel` from utils.conversions

AndrewRWilliams:
Updating documentation for wrappers: `to_parallel` is located in `pettingzoo.utils.conversions`, not `pettingzoo.utils` which results in an ImportError

# Description

Changing wrappers documentation related to `to_parallel`


AndrewRWilliams:
Actually, it seems like `to_parallel` is deprecated. Might be better to update documentation to `aec_to_parallel`

PR #875, opened 2023-01-03
Replace `setup.py` with `pyproject.toml`

jjshoots:


jjshoots:
@WillDudley I've fixed the 3.11 issues in this PR, do you mind giving it a review before I do precommit? Precommit changes like a million files for whatever reason.

@pseudo-rnd-thoughts Do you mind giving the setup.py to toml file change a look?

pyproject.toml
```diff
@@ -1,5 +1,106 @@
...
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    'Intended Audience :: Science/Research',
+    'Topic :: Scientific/Engineering :: Artificial Intelligence',
+]
+
+dependencies = [
+    "numpy >=1.21.0",
+    "gymnasium>0.26.0",
```
pseudo-rnd-thoughts, line 30: >= ??

pyproject.toml
```diff
@@ -1,5 +1,106 @@
...
+    "hanabi_learning_environment==0.0.4",
+]
+butterfly = ["pygame==2.1.3.dev8", "pymunk==6.2.0"]
+mpe = ["pygame==2.1.3.dev8"]
+sisl = ["pygame==2.1.3.dev8", "box2d-py==2.3.5", "scipy>=1.4.1"]
+other = ["pillow>=8.0.1"]
+tests = [
+    "pynput",
+    "pytest",
+    "codespell",
```
pseudo-rnd-thoughts, line 50: If I understand correctly, codespell, falke8, isort, bandit and pre-commit are part of the pre-commit CI.
In gymnasium, this is done completely separately that doesn't require these modules being part of the setup. 

https://github.com/Farama-Foundation/Gymnasium/blob/main/.github/workflows/pre-commit.yml

pyproject.toml
```diff
@@ -1,5 +1,106 @@
...
+    "flake8",
+    "isort",
+    "AutoROM",
+    "bandit",
+    "pytest",
+    "pytest-cov",
+    "pre-commit",
+]
+all = [
```
pseudo-rnd-thoughts, line 59: This is a stylistic change so ignore. 
In gymnasium, we have separated `all` and `test`. `all` is used for all of the modules that end-users might use, i.e., atari, box2d, etc while `test` is just for devs if they need to run the tests. 
In the CI for testing, we run `pip install gymnasium[all, test]`, feel free to follow this style. 
In my eyes, this simplifies the setup file

pyproject.toml
```diff
@@ -1,5 +1,106 @@
...
+    "pytest",
+    "pytest-cov",
+    "pre-commit",
+]
+
+[project.urls]
+Homepage = "https://farama.org"
+Repository = "https://github.com/Farama-Foundation/PettingZoo"
+Documentation = "https://pettingzoo.farama.org"
+"Bug Report" = "https://github.com/Farama-Foundation/PettingZoo/issues"
```
pseudo-rnd-thoughts, line 85: Should this be in quotes?

pseudo-rnd-thoughts (approved):
I am not super familiar with pyproject.toml file but this looks generally good, except my files

pyproject.toml
jjshoots: Got it

pyproject.toml
jjshoots: I think its still good to have pre-commit installed under testing, this allows us to do pre-commit runs locally before pushing. But I've removed flake8, isort, and bandit.

pyproject.toml
jjshoots: I was following the way it was defined in the `setup.py` file. @WillDudley what are your preferences?

pyproject.toml
jjshoots: It can't because there is a space between Bug and Report. This was the way it's done in Gymnasium as well.

pyproject.toml
pseudo-rnd-thoughts: pre-commit doesn't actually use your locally installed version of the modules. I believe that pre-commit creates a virtual environment to use but I might be wrong

pyproject.toml
WillDudley: @jjshoots No personal preference - I try to copy Gymnasium but maybe I forgot this time, or maybe it's from several mo ago.

jjshoots: Fixed to match gymnasium.

jjshoots:
@WillDudley should be ready to merge

PR #886, opened 2023-02-22
Fix CI

pseudo-rnd-thoughts:
I have updated pre-commit and rerun it, `pre-commit autoupdate` and `pre-commit run --all-files`, this seems to have fixed those issues
For the tutorials, numba doesn't support python 3.11 therefore, I have removed the CI for 3.11 of the tutorials


jjshoots (approved):
LGTM

PR #892, opened 2023-03-02
Use predownloaded targz for tests

jjshoots:


jjshoots:
This is failing because AutoROM doesn't support 3.11.

PR #894, opened 2023-03-04
Update simple_spread agent reward

sushant1212:
# Description
The non-global reward function for individual agents is -1 for each agent it collides with. The agent would always get a reward of -1 for colliding with itself which does not make sense. I've added a check to not consider an agent colliding with itself.

## Type of change
- Bug fix 

# Checklist:

### DIFF
--- a/pettingzoo/mpe/simple_spread/simple_spread.py
+++ b/pettingzoo/mpe/simple_spread/simple_spread.py
@@ -164,6 +164,8 @@ class Scenario(BaseScenario):
         rew = 0
         if agent.collide:
             for a in world.agents:
+                if(agent == a):
+                    continue
                 if self.is_collision(a, agent):
                     rew -= 1
         return rew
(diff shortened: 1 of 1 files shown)


pettingzoo/mpe/simple_spread/simple_spread.py
```diff
@@ -164,6 +164,8 @@ def reward(self, agent, world):
         rew = 0
         if agent.collide:
             for a in world.agents:
+                if(agent == a):
+                    continue
                 if self.is_collision(a, agent):
                     rew -= 1
```
jjshoots, line 170: Could you change this to `rew -= 1.0 * (self.is_collision(a, agent) and a != agent)`? This just makes it cleaner.

sushant1212: Yeah sure

jjshoots (approved):
LGTM!

PR #897, opened 2023-03-09
Fixing warnings

reginald-mclean:
# Description

This is an update to the PR I made [here](https://github.com/jjshoots/PettingZoo/pull/2). This update makes the if statements inline with other logic and makes some small changes to test logic. 

Fixing a large number of test warnings. These warnings were being caused by mismatches between an environment's specifications and the tests. For example a warning was being thrown for "simple_adversary_v2" because the state space had a maximum possible value of np.inf, but that is valid according to the env's specs. 

The method that I took to do these changes is to create lists of envs that have these violation specifications. Then if a warning should be thrown, there is an if statement that first checks if that environments name is in the associated list. For example there is a list looking for maximum state space of positive infinity. Only if an environment's name is not in the list will the warning be thrown. 

One of the files just this if statement functionality was added, the state_test file needed the env name passed to the function as the env itself was not passed.


jjshoots:
@reginald-mclean, could you fix the pre-commit? I think we can merge after that.

jjshoots (approved):
Merging should fix the precommit.

PR #901, opened 2023-03-13
Bug fixed related to a scalar np.array

zbenmo:
# Description

```
def test_evaluate():

  env = CollectCoinsEnv(pieces=['rock', 'rock'])

  action = None

  def another_action_taken(action_taken):
    nonlocal action
    action = action_taken

  # Wrapping the original environment as to make sure a valid action will be taken.
  env = EnsureValidAction(
      env,
      env.check_action_valid,
      env.provide_alternative_valid_action,
      another_action_taken
  )

  api_test(env, num_cycles=10, verbose_progress=False)

```

```
    obs_space = dict(
      board = gym.spaces.MultiBinary([8, 8]),
      player = gym.spaces.Tuple([gym.spaces.Discrete(8)] * 2),
      other_player = gym.spaces.Tuple([gym.spaces.Discrete(8)] * 2)
    )    
    self._observation_space = gym.spaces.Dict(spaces=obs_space)
    self._action_space = gym.spaces.Tuple([gym.spaces.Discrete(8)] * 2)

...

  # this cache ensures that same space object is returned for the same agent
  # allows action space seeding to work as expected
  @functools.lru_cache(maxsize=None)
  def observation_space(self, agent):
      # gymnasium spaces are defined and documented here: https://gymnasium.farama.org/api/spaces/
      return self._observation_space

  @functools.lru_cache(maxsize=None)
  def action_space(self, agent):
      return self._action_space
```

While I am aware of the concept of "action_mask", I still do it with the following wrapper.

```
from typing import TypeVar, Callable
import gymnasium as gym
from .action_wrapper_pz import ActionWrapper


Action = TypeVar("Action")


class EnsureValidAction(ActionWrapper):
  """
  A gym environment wrapper to help with the case that the agent wants to take invalid actions.
  For example consider a Chess game, where you let the action_space be any piece moving to any square on the board,
  but then when a wrong move is taken, instead of returing a big negative reward, you just take another action,
  this time a valid one. To make sure the learning algorithm is aware of the action taken, a callback should be provided.
  """
  def __init__(self, env: gym.Env,
    check_action_valid: Callable[[Action], bool],
    provide_alternative_valid_action: Callable[[Action], Action],
    alternative_action_cb: Callable[[Action], None]):

    super().__init__(env)
    self.check_action_valid = check_action_valid
    self.provide_alternative_valid_action = provide_alternative_valid_action
    self.alternative_action_cb = alternative_action_cb

  def action(self, action: Action) -> Action:
    if self.check_action_valid(action):
      return action
    alternative_action = self.provide_alternative_valid_action(action)
    self.alternative_action_cb(alternative_action)
    return alternative_action
```

```
from pettingzoo.utils.wrappers import BaseWrapper
import gymnasium as gym


class ActionWrapper(BaseWrapper):
  def __init__(self, env: gym.Env):
    super().__init__(env)

  def step(self, action):
    action = self.action(action)
    self.env.step(action)

  def action(self, action):
    pass

  def render(self, *args, **kwargs):
    self.env.render(*args, **kwargs)
```

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/pettingzoo/test/api_test.py
+++ b/pettingzoo/test/api_test.py
@@ -24,6 +24,8 @@ list and observation_spaces, action_spaces dictionaries."""
 
 def test_observation(observation, observation_0):
     if isinstance(observation, np.ndarray):
+        if observation.shape == ():
+            return # or check something.. TODO:
         if np.isinf(observation).any():
             warnings.warn(
                 "Observation contains infinity (np.inf) or negative infinity (-np.inf)"
@@ -297,11 +299,12 @@ def play_test(env, observation_0, num_cycles):
             assert env.observation_space(agent).dtype == prev_observe.dtype
         assert env.observation_space(agent).contains(
             prev_observe
-        ), "Out of bounds observation: " + str(prev_observe)
+        ) or env.observation_space(agent).contains(prev_observe[()]), "Out of bounds observation: " + str(prev_observe)
 
+        # Why the same assertion again???
         assert env.observation_space(agent).contains(
             prev_observe
-        ), "Agent's observation is outside of it's observation space"
+        ) or env.observation_space(agent).contains(prev_observe[()]), "Agent's observation is outside of it's observation space"
         test_observation(prev_observe, observation_0)
         if not isinstance(env.infos[env.agent_selection], dict):
             warnings.warn(


elliottower:
Thanks for the contribution but I'm not sure I understand, could you give an example case where this is needed? I don't see what your code would fix, afaik the code `pre_observe[()]` is the same as just calling `pre_observe` as you aren't indexing with anything. I can see there may be a problem if the previous observation is a single scalar rather than an array and `contains()` return false but I'd have to see an example and test it out.

zbenmo:
Without my "fix" the assertion failed.

pettingzoo/test/api_test.py
```diff
@@ -386,11 +386,12 @@ def play_test(env, observation_0, num_cycles):
             assert env.observation_space(agent).dtype == prev_observe.dtype
         assert env.observation_space(agent).contains(
             prev_observe
-        ), "Out of bounds observation: " + str(prev_observe)
+        ) or env.observation_space(agent).contains(prev_observe[()]), "Out of bounds observation: " + str(prev_observe)
 
+        # Why the same assertion again???
         assert env.observation_space(agent).contains(
             prev_observe
-        ), "Agent's observation is outside of it's observation space"
```
jjshoots, line 393: I think just removing this whole chunk is suffice

jjshoots (changes requested):
Thanks for the contribution and for pointing this out!

While I agree that having two of the same tests is redundant, `prev_observation[()]` is actuall yequivalent to `prev_observation`. Furthermore, doing `[()]` is quite unpythonic and what it's doing is not immediately obvious. Instead, could we figure out exactly why this "fix" is fixing things?

pettingzoo/test/api_test.py
zbenmo: I have seen that np.array(obj), meaning that the ndim is 0, or how I have accessed the item with [()], the "contains" fails. Then by checking the item itself, it works. So ndim shape np.array is not in the space, the object itself is.

zbenmo: Seems to be my bug. Closing. If issue returns, I'll open a new PR.

PR #907, opened 2023-03-20
Update third_party_envs.md

elliottower:
Add two custom envs I've created using the current version of pettingzoo (and update SUMO-RL with link to the SUMO traffic simulation)


jjshoots (approved):
LGTM, thanks for the contrib!

PR #909, opened 2023-03-20
Bump versions for all tutorials

jjshoots:


jjshoots:
Closing because I think I broke something on git.

PR #924, opened 2023-03-27
Added EzPickle support to all envs

elliottower:
# Description

Most environments already have EzPickle support, this PR extends that to all envs. Adding support is very simple, just requires 3 lines per file. Mirrored the convention of calling EzPickle.__init__() as the first line in the constructor, as was recommended by @pseudo-rnd-thoughts on Shimmy and is used throughout Gymnasium and the other envs here which support EzPickle.

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/pettingzoo/classic/chess/chess.py
+++ b/pettingzoo/classic/chess/chess.py
@@ -90,6 +90,7 @@ import gymnasium
 import numpy as np
 from gymnasium import spaces
 from gymnasium.error import DependencyNotInstalled
+from gymnasium.utils import EzPickle
 
 from pettingzoo import AECEnv
 from pettingzoo.utils import wrappers
@@ -106,7 +107,7 @@ def env(render_mode=None):
     return env
 
 
-class raw_env(AECEnv):
+class raw_env(AECEnv, EzPickle):
     metadata = {
         "render_modes": ["human", "ansi", "rgb_array"],
         "name": "chess_v5",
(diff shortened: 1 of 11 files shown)


elliottower:
Mark says this shouldn't be necessary, going to change this PR to add testing that the ezpickle functionality works as properly

elliottower:
Closing in favor of https://github.com/Farama-Foundation/PettingZoo/pull/928

benblack769:
Hmm. Awhile ago, I was thinking it might be good to EzPickle even the environments that don't seem to need it, as you did here, as EzPickle destroys any state, includeing random state. Preserving random state, as regular pickle does can lead to all sorts of really annoying unintended behavior when multitprocessing.

But now that seed() is its own funciton, I guess it is fine to rely on regular pickle, then just let the user re-seed the environment to get environments with different states.

elliottower:
> Hmm. Awhile ago, I was thinking it might be good to EzPickle even the environments that don't seem to need it, as you did here, as EzPickle destroys any state, includeing random state. Preserving random state, as regular pickle does can lead to all sorts of really annoying unintended behavior when multitprocessing.
> 
> But now that seed() is its own funciton, I guess it is fine to rely on regular pickle, then just let the user re-seed the environment to get environments with different states.

Thanks for the insight, wasn’t aware of that about EzPickle. I could see an argument to make it all consistent behavior wise but don’t think it matters too much. Also I’ve heard from Jordan you would be a good person to ask about some things related to MPE? In the other PR which I closed this in favor of, I had to add EzPickle to the simple MPE env but I think there was an inconsistency that made be unsure if it was the right thing to do. Maybe you could look that pr over? #928

PR #925, opened 2023-03-28
Fix chess import docstring

Kallinteris-Andreas:
# Description

title says it all

## Type of change

- DOC change

# Checklist:

### DIFF
--- a/pettingzoo/classic/chess/chess.py
+++ b/pettingzoo/classic/chess/chess.py
@@ -9,7 +9,7 @@
 
 This environment is part of the <a href='..'>classic environments</a>. Please read that page first for general information.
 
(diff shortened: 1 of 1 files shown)


elliottower (dismissed):
Looks good to me. Just as a note, DOC made me think it was an acronym, probably better to say docstring. And please take the time to write at least a sentence for every PR even if it seems obvious, just good to have some more info than you can fit in a title for making releases (don't want to have to dig around the code from every commit, best to have each PR be descriptive)

elliottower (approved):
Sorry for the confusion, looks good now. For some reason it said this branch had not been deployed which made me think the tests hadn't passed.

PR #928, opened 2023-03-29
Add pickling tests, adapt all envs to be picklable

elliottower:
# Description

This PR adds tests to check that each environment can be saved and loaded via pickling. Waterworld and mpe simple envs both have pygame objects (pygame.Clock and pygame Surface, respectively) and cannot be pickled without EzPickle. I have removed the changes I made to version numbers and other miscellaneous changes to make this PR as simple as possible.

Previous comments:
> > and includes them in the API test. I'm setting this as a draft because I'm not sure if we should include it in the API test, as that includes third-party environments which may not care about pickling. But perhaps it is good to enforce that all environments can be pickled, even custom ones? All tests pass internally.
> 
> I have moved them to a separate file from API tests as it was messy and causing other tests to fail due to rendering and video issues, also seems simplest to do it this way. 

## Type of change

### Screenshots

# Checklist:


elliottower:
Do we care if rendered pygame envs can be pickled? Currently they all work with render_mode=None but with pygame rendering there are some issues like chess has a pygame.time.Clock object which can't be pickled. Could add EzPickle to that file but it's a bit weird to do that as it's all python and I don't imagine people will try to pickle rendered envs. When you load an env which has been pickled and uses pygame it also doesn't have the video system initialized, as the pygame.init() call is in the constructor rather than in reset(). 
@pseudo-rnd-thoughts @jkterry1

test/pickle_test.py
```diff
@@ -0,0 +1,141 @@
+import pickle
+
+import pytest
+from gymnasium.utils.env_checker import data_equivalence
+
+from .all_modules import all_environments
+
+ALL_ENVS = list(all_environments.items())
+FAILING_ENV_NAMES = ["mpe/simple_world_comm_v2"]
```
benblack769, line 9: Only `simple_world_comm_v2` is failing the test? That is a bit weird, I would expect most, if not all of the MPE environments to fail a test like this if one of them does. `simple_world_comm_v2` is a bit more complicated in terms of how many agents there are and the complexity of the observations and reward, but it doesn't do anything different in terms of the underlying components.

test/pickle_test.py
```diff
@@ -0,0 +1,141 @@
...
+FAILING_ENV_NAMES = []
+PASSING_ENVS = [
+    (name, env_module)
+    for (name, env_module) in ALL_ENVS
+    if name not in FAILING_ENV_NAMES
+]
+
+
+@pytest.mark.skip(
+    reason="pickling pygame rendered envs does not work well, video system will not be initialized."
```
benblack769, line 85: Yeah, pickling an environment which is rendering to screen will always be a weird thing to do. Two ideas I can think of are:

1. EzPickle: Just standardize the concept that environments aren't things you can pickle, and that users will only ever be pickling the environmnt arguments via EzPickle.
2. Explicitly disable pickling for rendered environments. This is possible by overriding `__getstate__` to raise an error if a redering flag is set. Perhaps this can be done in a base layer or wrapper.

elliottower: I feel like we could just ignore rendered states or raise a warning if you pickle with render mode not None. I guess though it keeps the args so when they load it again it’ll have the render mode but won’t actually render, so I can see that being an issue. 

The inconsistency you mentioned in the other closed PR about pickle saving RNG state but EzPickle destroying it makes me think we might want to just do EzPickle across the board for consistency.

test/pickle_test.py
elliottower: Not sure why that one in particular failed, I can try to run it again and see what the issue was though.

tutorials/CleanRL/requirements.txt
```diff
@@ -1,5 +1,5 @@
 SuperSuit==3.7.2
 torch==1.13.1
-pettingzoo[butterfly]==1.22.4
```
RedTachyon, line 3: I think these three files need this line to be fixed? Maybe you just need to solve the merge conflict

elliottower: Good catch I’ll update it when I get home

elliottower: Wait I don’t think this needs to be changed, 1.22.4 got yanked so it needs to be 1.22.3

elliottower: Oh I see there was a merge conflict, got it figured out

pettingzoo/mpe/simple/simple.py
```diff
@@ -51,8 +52,9 @@
 from .._mpe_utils.simple_env import SimpleEnv, make_env
```
pseudo-rnd-thoughts, line 52: We shouldn't be using local imports

pettingzoo/mpe/simple/simple.py
```diff
@@ -51,8 +52,9 @@
...
 
 
-class raw_env(SimpleEnv):
+class raw_env(SimpleEnv, EzPickle):
     def __init__(self, max_cycles=25, continuous_actions=False, render_mode=None):
+        EzPickle.__init__(self, max_cycles, continuous_actions, render_mode)
         scenario = Scenario()
         world = scenario.make_world()
         super().__init__(
```
pseudo-rnd-thoughts, line 60: When using multi-inheritance, don't use `super().__init__` as python needs to guess which of the classes you are you talking about. Therefore change to `SimpleEnv.__init__`

pettingzoo/mpe/simple/simple.py
```diff
@@ -51,8 +52,9 @@
 from .._mpe_utils.simple_env import SimpleEnv, make_env
 
 
-class raw_env(SimpleEnv):
+class raw_env(SimpleEnv, EzPickle):
     def __init__(self, max_cycles=25, continuous_actions=False, render_mode=None):
+        EzPickle.__init__(self, max_cycles, continuous_actions, render_mode)
```
pseudo-rnd-thoughts, line 57: Use keyword arguments not positional arugments as it is easier when debugging to understand the value, `max_cycles=max_cycles`

pseudo-rnd-thoughts (changes requested):
Repeat comments for the second class

pettingzoo/mpe/simple/simple.py
elliottower: Thanks for the review, will incorporate these changes into shimmy too as I think there were some cases where positional args were used

pettingzoo/mpe/simple/simple.py
elliottower: Good catch I just left that code as is but definitely needed to adapt it

pettingzoo/mpe/simple_adversary/simple_adversary.py
```diff
@@ -55,16 +55,21 @@
...
+        EzPickle.__init__(
+            self,
+            N=N,
+            max_cycles=max_cycles,
+            continuous_actions=continuous_actions,
+            render_mode=render_mode,
+        )
         scenario = Scenario()
         world = scenario.make_world(N)
         super().__init__(
```
pseudo-rnd-thoughts, line 75: `SimpleEnv.__init__`

pettingzoo/mpe/simple_crypto/simple_crypto.py
```diff
@@ -75,7 +74,12 @@
...
-        EzPickle.__init__(self, max_cycles, continuous_actions, render_mode)
+        EzPickle.__init__(
+            self,
+            max_cycles=max_cycles,
+            continuous_actions=continuous_actions,
+            render_mode=render_mode,
+        )
         scenario = Scenario()
         world = scenario.make_world()
         super().__init__(
```
pseudo-rnd-thoughts, line 85: `SimpleEnv.__init__`

pettingzoo/mpe/simple_push/simple_push.py
```diff
@@ -50,16 +50,20 @@
...
-        EzPickle.__init__(self, max_cycles, continuous_actions, render_mode)
+        EzPickle.__init__(
+            self,
+            max_cycles=max_cycles,
+            continuous_actions=continuous_actions,
+            render_mode=render_mode,
+        )
         scenario = Scenario()
         world = scenario.make_world()
         super().__init__(
```
pseudo-rnd-thoughts, line 69: `SimpleEnv.__init__`

pettingzoo/mpe/simple_reference/simple_reference.py
```diff
@@ -56,23 +56,22 @@
...
+from pettingzoo.mpe._mpe_utils.scenario import BaseScenario
+from pettingzoo.mpe._mpe_utils.simple_env import SimpleEnv, make_env
 from pettingzoo.utils.conversions import parallel_wrapper_fn
 
-from .._mpe_utils.core import Agent, Landmark, World
-from .._mpe_utils.scenario import BaseScenario
-from .._mpe_utils.simple_env import SimpleEnv, make_env
-
 
 class raw_env(SimpleEnv, EzPickle):
```
pseudo-rnd-thoughts, line 65: Fix

pettingzoo/mpe/simple_speaker_listener/simple_speaker_listener.py
```diff
@@ -51,16 +51,20 @@
...
+from pettingzoo.mpe._mpe_utils.scenario import BaseScenario
+from pettingzoo.mpe._mpe_utils.simple_env import SimpleEnv, make_env
 from pettingzoo.utils.conversions import parallel_wrapper_fn
 
-from .._mpe_utils.core import Agent, Landmark, World
-from .._mpe_utils.scenario import BaseScenario
-from .._mpe_utils.simple_env import SimpleEnv, make_env
-
 
 class raw_env(SimpleEnv, EzPickle):
```
pseudo-rnd-thoughts, line 60: Fix

pseudo-rnd-thoughts, line 60: Fix

test/all_parameter_combs_test.py
```diff
@@ -8,38 +8,38 @@
...
-    ["classic/leduc_holdem_v4", leduc_holdem_v4, dict(num_players=2)],
-    ["classic/leduc_holdem_v4", leduc_holdem_v4, dict(num_players=3)],
-    ["classic/leduc_holdem_v4", leduc_holdem_v4, dict(num_players=4)],
-    ["classic/texas_holdem_v4", texas_holdem_v4, dict(num_players=2)],
-    ["classic/texas_holdem_v4", texas_holdem_v4, dict(num_players=3)],
-    ["classic/texas_holdem_v4", texas_holdem_v4, dict(num_players=4)],
-    ["classic/texas_holdem_no_limit_v6", texas_holdem_no_limit_v6, dict(num_players=2)],
-    ["classic/texas_holdem_no_limit_v6", texas_holdem_no_limit_v6, dict(num_players=3)],
-    ["classic/texas_holdem_no_limit_v6", texas_holdem_no_limit_v6, dict(num_players=4)],
+    # ["atari/boxing_v2", boxing_v2, dict(obs_type="grayscale_image")],
```
pseudo-rnd-thoughts, line 11: Why are these commented out?

test/pickle_test.py
pseudo-rnd-thoughts: With EzPickle added, this is true still?

pseudo-rnd-thoughts (changes requested):
With the changes, looks good

test/all_parameter_combs_test.py
elliottower: Thanks for giving this a second look through. This one was just a temporary change to only test the things that were failing when I was debugging things to shorten the iteration time, will remove them and go over things again to make sure there’s no more relative imports or super calls with ezpickle

pseudo-rnd-thoughts: Could you revert this change

elliottower: Oops lol sorry about that will do

test/pickle_test.py
elliottower: The problem was the action spaces are different for the different agents and my code didn't seed all of the action spaces, and assumed they had the same action space.

pseudo-rnd-thoughts (approved):

PR #930, opened 2023-03-31
Make the API match Gymnasium

RedTachyon:
- Add back return info on reset for Parallel Envs
- Replace `env.seed` with `env._seed`, not meant to be used by people (instead use `env.reset(seed=seed)`


elliottower (approved):

pyproject.toml
elliottower, line 1: Should be 28.1

RedTachyon: What is the actual minimum compatibility requirement? @jjshoots I think you added this in the previous PR.

Is there a specific point of incompatibility where you can't use this with Gymnasium 0.26?

(the distinction between 0.28.0 and 0.28.1 doesn't really matter, since its `>=`, and 0.28.0 is yanked)

jjshoots: Interesting, it says I was the one who added it in but somehow I don't remember why.

jjshoots (approved):
Everything looks fairly sane to me. Although admittedly I didn't do the whole "trail the whole repo with pycharm" thing. Situations like these make me wish we had a code coverage checker.

elliottower:
> Everything looks fairly sane to me. Although admittedly I didn't do the whole "trail the whole repo with pycharm" thing. Situations like these make me wish we had a code coverage checker.

Not sure what you mean exactly but the pytests do code coverage if that’s what you’re talking about

jjshoots:
@elliottower, yeah that's exactly what I mean. Code Cov would be nice but it's not an enforced standard for now.

PR #927, opened 2023-03-29
Add back return info on reset for Parallel Envs

jjshoots:
# Description

This fixes the bug that was introduces in #890, whereby reset no longer returned infos.

Fixes #926

PR #931, opened 2023-03-31
Updates to the reset info fix

RedTachyon:
WiP, see #927 #930


elliottower:
Thanks for looking into this, I'll take a look once things are finished.

PR #933, opened 2023-04-02
Update hanabi to use shimmy[openspiel] instead of  Hanabi Learning Env

elliottower:
# Description

This PR updates hanabi to use OpenSpiel's Hanabi implementation rather than hanabi learning environment. This implementation is in C++ and has been updated as recently as 4 weeks ago, whereas hanabi learning environment has not been updated since 2021. There is a flag in openspiel to use the hanabi learning environment but we don't have it enabled as there's not much point. The hanabi learning environment seems to be very limited from what I can tell, and this is a much more full featured implementation. It is also listed as thoroughly tested under openspiel's available games list: https://openspiel.readthedocs.io/en/latest/games.html. 

The previous hanabi implementation (v4) used very different ways of doing things from all of the other games in pettingzoo.classic, which makes sense as it was last updated in pettingzoo v1.8.0. 

This PR also updates the API test code to properly handle action masking from pettingzoo's classic games, contained in obs['action_mask']. Previously, the code used numpy to randomly sample from the action space, but this led to errors if the action mask has zero valid actions, whereas my code using action_space(agent).sample(mask) works how we expect it to, generating a valid action in the space, which then causes the environment to terminate (as the action is illegal). I did some other cleanups but the tests are still passing, mostly just code style changes and accounting for the possibility of an action mask. 

Unfortunately I was not able to get hanabi's seed test to work, which is bizarre because in Shimmy the openspiel seed tests actually do work but the API test doesn't. I think the API test is much more important to get working and I spent the majority of my time on that, but I can delay merging this and see if it's possible to fix the errors. The seed tests in PettingZoo should probably be updated as well, they are very confusing to debug and totally unlike the testing used in gymnasium and which we did in shimmy to test pettingzoo environments, which I think was much cleaner (adapted gymnasium's seed test code). 

I think if I do update the seed tests to match these other repos they may end up passing for Hanabi, but I'm not exactly sure what the current seed tests do in comparison to it (wouldn't want to change it and remove functionality), so maybe someone could help figure that out before I make those changes.

## Type of change

### Screenshots

# Checklist:


elliottower:
The render tests work with the most recent version of shimmy where I enabled rendering for openspiel envs, once that is released this will pass.

test/all_parameter_combs_test.py
```diff
@@ -228,7 +229,7 @@ def test_module(name, env_module, kwargs):
     api_test(_env)
 
     # some atari environments fail this test
-    if "atari/" not in name:
+    if "atari/" not in name and "hanabi" not in name:
```
jjshoots, line 232: Could we do `if name not in ["atari", "hanabi"]`?

pyproject.toml
```diff
@@ -37,7 +37,7 @@ classic = [
     "chess==1.7.0",
     "rlcard==1.0.5",
     "pygame==2.1.3.dev8",
-    "hanabi_learning_environment==0.0.4",
+    "shimmy[openspiel]>=0.2.0"
```
jjshoots, line 40: What do we think about pettingzoo depending on shimmy and shimmy depending on pettingzoo with this?

pettingzoo/test/api_test.py
```diff
@@ -382,15 +383,18 @@ def play_test(env, observation_0, num_cycles):
...
-            prev_observe
-        ), "Agent's observation is outside of it's observation space"
+        if isinstance(env.observation_space(agent), gymnasium.spaces.Box):
+            if isinstance(prev_observe, dict) and "observation" in prev_observe:
+                assert (
+                    env.observation_space(agent).dtype
+                    == prev_observe["observation"].dtype
+                )
+            else:
+                assert env.observation_space(agent).dtype == prev_observe.dtype
```
jjshoots, line 397: We try to keep the tests as never-nesting as possible, so syntactically this can be fixed.

But that aside, how can `prev_observe` be a dict when the observation space is a Box?

pettingzoo/test/api_test.py
```diff
@@ -333,10 +330,14 @@ def play_test(env, observation_0, num_cycles):
...
-            action = random.choice(np.flatnonzero(prev_observe["action_mask"]).tolist())
         else:
-            action = env.action_space(agent).sample()
+            if isinstance(prev_observe, dict) and "action_mask" in prev_observe:
+                mask = prev_observe["action_mask"]
+            elif isinstance(info, dict) and "action_mask" in info:
+                mask = info["action_mask"]
+            else:
+                mask = None
+            action = env.action_space(agent).sample(mask)
```
jjshoots, line 340: Can we not have an if loop within the else?

jjshoots (changes requested):
Some minor questions, but other than that fairly sane.

pyproject.toml
elliottower: So I talked a bit with mark about this I believe and the only problem is if there is a base requirement with each repo, leading to a circular import, so gymnasium has shimmy as an an extra/optional requirement to avoid that issue, this is doing the same thing. Should get confirmation from someone else though just to be sure this isn’t an issue.

pettingzoo/test/api_test.py
elliottower: I removed the info thing because that wasn’t actually used by pettingzoo. I believe your shimmy openspiel compatibility wrapper has action mask in info by default which was what threw me off. Again as above, this stuff is all pretty poorly designed and should be made consistent imo.

pettingzoo/classic/hanabi/hanabi.py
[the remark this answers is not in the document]
elliottower: The only functional change was I had to remove the Hanabi penalty function as it seemed not to work no matter what I did, it was previously used to penalize certain illegal moves more than others (called in the terminate illegal wrapper) but every other env in pettingzoo used -1 as the illegal reward so I just did that instead. But someone familiar with Hanabi may be able to confirm that things work the same. I tried doing a seed test with the v4 and v5 versions in separate files to confirm the actions and outputs were the same but had some issues, can go through that again and confirm the starting configurations are the same, it may have to do with differences in RNG though.

pettingzoo/test/api_test.py
jjshoots: I understand that some envs may have a dict space, but why are we checking if prev_obsrevation is a dict (line 391), when we've checked that the observation space is already a Box (line 390)?

pettingzoo/test/api_test.py
jjshoots: Sorry, I meant this more as a stylistic thing, line 333 is an `else:`, then within this `else:` is this if loop. Could we not flatten this?

jjshoots: I'll have a look at this again in the near future, along with the comment above this one.But on first glance it looks like I should put action mask into the observation for openspiel on shimmy.

pettingzoo/test/api_test.py
jjshoots: I'll look at this again in the near future, I still don't understand why we're checking the prev_observation space when we also check the whole observation space in general.

pettingzoo/test/api_test.py
```diff
@@ -331,12 +328,11 @@ def play_test(env, observation_0, num_cycles):
...
-        if terminated or truncated:
-            action = None
-        elif isinstance(prev_observe, dict) and "action_mask" in prev_observe:
-            action = random.choice(np.flatnonzero(prev_observe["action_mask"]).tolist())
+        if isinstance(prev_observe, dict) and "action_mask" in prev_observe:
+            mask = prev_observe["action_mask"]
         else:
-            action = env.action_space(agent).sample()
+            mask = None
+        action = env.action_space(agent).sample(mask)
```
jjshoots, line 335: Are we concerned that mask may not be None even when the env has terminated/truncated?

pettingzoo/test/api_test.py
```diff
@@ -481,6 +479,9 @@ def progress_report(msg):
 
     env.reset()
     observation_0, *_ = env.last()
+    if isinstance(observation_0, dict) and "observation" in observation_0:
+        observation_0 = observation_0["observation"]
+
```
jjshoots, line 484: This looks hacky but, I guess it works.

jjshoots (approved):
Some general comments, but otherwise LGTM.

jjshoots (changes requested):
Edit: once tests pass.

pettingzoo/test/api_test.py
```diff
@@ -333,10 +330,11 @@ def play_test(env, observation_0, num_cycles):
         prev_observe, reward, terminated, truncated, info = env.last()
         if terminated or truncated:
             action = None
-        elif isinstance(prev_observe, dict) and "action_mask" in prev_observe:
-            action = random.choice(np.flatnonzero(prev_observe["action_mask"]).tolist())
+        if isinstance(prev_observe, dict) and "action_mask" in prev_observe:
```
jjshoots, line 333: elif

pettingzoo/test/api_test.py
```diff
@@ -333,10 +330,11 @@ def play_test(env, observation_0, num_cycles):
...
+            mask = prev_observe["action_mask"]
         else:
-            action = env.action_space(agent).sample()
+            mask = None
+        action = env.action_space(agent).sample(mask)
```
jjshoots, line 337: Change to:
```py
        if terminated or truncated:
            action = None
        else:
            mask = prev_observe.get("action_mask") if isinstance(prev_observe, dict) else None
            action = env.action_space(agent).sample(mask)
```

jjshoots, line 337: Change to:
```py
        if terminated or truncated:
            action = None
        else:
            mask = prev_observe.get("action_mask") if isinstance(prev_observe, dict) else None
            action = env.action_space(agent).sample(mask)
```

jjshoots (changes requested):

pseudo-rnd-thoughts:
Given the complexity of the PR, there seems to be several different PRs happening here, updating `test_seeding`, docs updates and the hanabi change. Personally, I would separate out the PR and this will make it simpler to solve the issue

elliottower:
> Given the complexity of the PR, there seems to be several different PRs happening here, updating `test_seeding`, docs updates and the hanabi change. Personally, I would separate out the PR and this will make it simpler to solve the issue

Yea makes sense, I was thinking the same and believe I mentioned it in the pz dev channel about how it should probably be split unless we want to make an exception.

pseudo-rnd-thoughts:
It is simpler 90% of the time to have several prs if you want to fix several issues unless you are changing a total of 20 lines or less

elliottower:
Closing in favor of separate PRs

PR #940, opened 2023-04-07
Rename seed to _seed

RedTachyon:
This PR removes the `env.seed` method in favor of `env._seed` in some environments.

Neither should be used by end-users, and instead they should use `env.reset(seed=seed)`


RedTachyon:
I'm gonna merge this and we can review the branch

PR #944, opened 2023-04-11
Fix seed test

jjshoots:
Depends on #933.


jjshoots:
Closing because easier to fix from #933's end.

PR #946, opened 2023-04-13
Update testing

elliottower:
# Description

Splitting https://github.com/Farama-Foundation/PettingZoo/pull/933 into separate PRs for simplicity.

This PR rewrites the seed test to actually check that environment rollouts are the same, matching gymnasium. It also updates the API test to handle dict observations, and fixes the variable env test which fails under the new seed test code.

## Type of change

### Screenshots

# Checklist:


pettingzoo/classic/rlcard_envs/texas_holdem_no_limit.py
```diff
@@ -151,7 +151,7 @@ def __init__(self, num_players=2, render_mode=None):
                                 ),
                                 [100, 100],
                             ),
-                            dtype=np.float64,
+                            dtype=np.float32,
```
jjshoots, line 154: Is there a reason for this change? AFAICT, all environments tend to use float64 everywhere.

pettingzoo/test/example_envs/generated_agents_env_v0.py
```diff
@@ -75,12 +75,25 @@ def reset(self, seed=None, options=None):
         self.truncations = {}
         self.infos = {}
         self.num_steps = 0
+
+        self._obs_spaces = {}
+        self._act_spaces = {}
+        self.types = []
+        self._agent_counters = {}
+        for i in range(3):
+            self.add_type()
```
jjshoots, line 84: What's going on here?

pettingzoo/test/example_envs/generated_agents_parallel_v0.py
```diff
@@ -67,10 +67,27 @@ def add_agent(self, type):
...
-        self.agents = []
         self.num_steps = 0
+
+        # Reset spaces and types
+        self._obs_spaces = {}
+        self._act_spaces = {}
+        self.types = []
+        self._agent_counters = {}
+        for i in range(3):
+            self.add_type()
```
jjshoots, line 78: Similar question to above.

pettingzoo/test/seed_test.py
jjshoots, line 1: I don't see anything immediately broken, but I'd feel more confident is someone else had a look as well.

pettingzoo/test/seed_test.py
```diff
@@ -1,109 +1,128 @@
...
+    max_env_iters = num_cycles * len(env1.agents)
+
+    for agent1, agent2 in zip(env1.agent_iter(), env2.agent_iter()):
+        assert data_equivalence(agent1, agent2), f"Incorrect agent: {agent1} {agent2}"
+
+        obs1, reward1, termination1, truncation1, info1 = env1.last()
+        obs2, reward2, termination2, truncation2, info2 = env2.last()
+
+        if termination1 or truncation1 or termination2 or truncation2:
+            break
```
jjshoots, line 40: We probably want to do this after the asserts below.

test/variable_env_test.py
```diff
@@ -1,36 +1,109 @@
...
+    ), "parallel_to_aec and aec_to_parallel wrappers should cancel out. Result: {type(env1)}, {type(env2)}."
+
+    # fails because using parallel_to_aec with this environment can produce nondeterministic behavior
     check_environment_deterministic(env1, env2, 500)
 
 
 if __name__ == "__main__":
     test_generated_agents_aec()
     test_generated_agents_parallel()
+    test_parallel_generated_agents_conversions()
```
jjshoots, line 109: This should just be removed. The file should be run with `pytest -v -rw test/variable_env_test.py` instead.

test/variable_env_test.py
```diff
@@ -1,36 +1,109 @@
...
 def test_parallel_generated_agents_conversions():
     parallel_api_test(aec_to_parallel(generated_agents_parallel_v0.env()))
     api_test(parallel_to_aec(generated_agents_parallel_v0.parallel_env()))
 
+    # double-wrapping should produce the same behavior as unwrapped env
+    env1 = generated_agents_parallel_v0.parallel_env()
+    env2 = aec_to_parallel(parallel_to_aec(generated_agents_parallel_v0.parallel_env()))
+
+    # fails (due to parallel_to_aec conversion issues with this environment)
+    # check_environment_deterministic_parallel(env1, env2, 500)
```
jjshoots, line 43: This problem should be fixed before this PR is merged ideally.

test/variable_env_test.py
```diff
@@ -1,36 +1,109 @@
...
+    env2 = aec_to_parallel(
+        parallel_to_aec(
+            aec_to_parallel(
+                parallel_to_aec(generated_agents_parallel_v0.parallel_env())
+            )
+        )
+    )
+    assert type(env1) == type(
+        env2
+    ), "parallel_to_aec and aec_to_parallel wrappers should cancel out. Result: {type(env1)}, {type(env2)}."
```
jjshoots, line 63: We don't need this. Instead just check that double wrapped is same as unwrapped.

test/variable_env_test.py
```diff
@@ -1,36 +1,109 @@
...
+    # quadruple-wrapping should produce deterministic behavior
+    parallel_seed_test(
+        lambda: aec_to_parallel(
+            parallel_to_aec(
+                aec_to_parallel(
+                    parallel_to_aec(generated_agents_parallel_v0.parallel_env())
+                )
+            )
+        )
+    )
```
jjshoots, line 77: Don't need this, just check the double wrapped and unwrapped case.

test/variable_env_test.py
```diff
@@ -1,36 +1,109 @@
...
+
+
+@pytest.mark.skip(
+    reason="parallel_to_aec wrapper produces nondeterministic behavior with this environment."
+)
```
jjshoots, line 82: This needs to be fixed before merge.

test/variable_env_test.py
```diff
@@ -1,36 +1,109 @@
...
     env1 = parallel_to_aec(generated_agents_parallel_v0.parallel_env())
     env2 = parallel_to_aec(
         aec_to_parallel(parallel_to_aec(generated_agents_parallel_v0.parallel_env()))
     )
+    assert type(env1) == type(
+        env2
+    ), "parallel_to_aec and aec_to_parallel wrappers should cancel out. Result: {type(env1)}, {type(env2)}."
+
+    # fails because using parallel_to_aec with this environment can produce nondeterministic behavior
     check_environment_deterministic(env1, env2, 500)
```
jjshoots, line 103: Just check the unwrapped and double wrapped case ideally.

jjshoots (changes requested):

pettingzoo/classic/rlcard_envs/texas_holdem_no_limit.py
elliottower: It was failing due to the space and sampled action being different types somehow, not sure what the cause was but good to know, I’ll look into it and try to get it working with int64 instead

pettingzoo/test/example_envs/generated_agents_env_v0.py
elliottower: This is used in the init but wasn’t in the reset(), very bizarre code but essentially adds randomized obs and action spaces. Problem was that calling reset(seed) wasn’t making things consistent.

pettingzoo/test/seed_test.py
elliottower: Good call, I’ll put it at the end. I think I copied this structure from shimmy or gymnasium so one of those might also do things slightly in the wrong order.

test/variable_env_test.py
elliottower: Makes sense, only added this just because the original code checker like crazy triple wrapped vs single but definitely seems unnecessary. Figured better safe than sorry but also don’t want to have confusing clunk either.

test/variable_env_test.py
elliottower: If we want to fix this I’d really appreciate someone  actually trying to debug the code cause I’ve spent many hours on multiple days messing with this and asked @benblack769 as well and got not much progress. Waterworld as well I spent a lot of time messing with to try and get it working but it won’t pass. I believe the problem is in the underlying env not my code but it could be to do with the conversation wrapper.

jjshoots:
@elliottower

So I've been poking around at this, it appears that
1. When [this](https://github.com/Farama-Foundation/PettingZoo/blob/master/pettingzoo/test/example_envs/generated_agents_parallel_v0.py#L93) is called,
2. [this](https://github.com/Farama-Foundation/PettingZoo/blob/master/pettingzoo/test/example_envs/generated_agents_parallel_v0.py#L47) gets triggered, which creates a new agent (here called `type` for whatever confusing reason)
3. which adds new observation and action spaces for the new agent [here](https://github.com/Farama-Foundation/PettingZoo/blob/master/pettingzoo/test/example_envs/generated_agents_parallel_v0.py#L51).
4. The problem is that these spaces are not seeded, which causes the tests to fail.

The solution here is to simply:

1. Store the seed in reset using something like `self.rng_seed = seed`.
2. Under `add_type`, do 'obs_space.seed(self.rng_seed)` and `act_space.seed(self.rng_seed)` after the obs and action spaces are declared.

test/variable_env_test.py
elliottower: I'm leaving these in just in case something goes wrong in the future and we have to debug and figure out what's wrong with this environment, good to have these baselines to ensure that wrapping these different number of times works correctly (given that one of the tests that is still failing is double wrapped vs unwrapped, and triple vs single wrapped was a separate problem and has been fixed now)

test/variable_env_test.py
```diff
@@ -1,36 +1,105 @@
...
+
+def test_parallel_generated_agents_conversions_triple():
+    # single- and triple- wrapping should cancel out into the same underlying type
     env1 = parallel_to_aec(generated_agents_parallel_v0.parallel_env())
     env2 = parallel_to_aec(
         aec_to_parallel(parallel_to_aec(generated_agents_parallel_v0.parallel_env()))
     )
+    assert type(env1) == type(
+        env2
+    ), "parallel_to_aec and aec_to_parallel wrappers should cancel out. Result: {type(env1)}, {type(env2)}."
```
jjshoots, line 52: This should be an f string.

test/all_parameter_combs_test.py
```diff
@@ -227,16 +226,16 @@ def test_module(name, env_module, kwargs):
     _env = env_module.env(**kwargs)
     api_test(_env)
 
-    # some atari environments fail this test
-    if "atari/" not in name:
+    # some atari environments fail this test, waterworld fails for certain seeds
+    if "atari/" not in name and env_module not in [waterworld_v4]:
```
jjshoots, line 230: Is there a reason we're doing env_module instead of just `name`?

jjshoots (changes requested):
Sorry, but some last final knitpicks, but otherwise golden.

test/variable_env_test.py
```diff
@@ -1,36 +1,105 @@
...
+    env2 = aec_to_parallel(
+        parallel_to_aec(
+            aec_to_parallel(
+                parallel_to_aec(generated_agents_parallel_v0.parallel_env())
+            )
+        )
+    )
+    assert type(env1) == type(
+        env2
+    ), "parallel_to_aec and aec_to_parallel wrappers should cancel out. Result: {type(env1)}, {type(env2)}."
```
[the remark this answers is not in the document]
elliottower: Tried the type checks for that and they failed because they weren’t the same wrapper type, I could do isinstance parallel env to ensure it worked correctly though.

test/variable_env_test.py
```diff
@@ -1,36 +1,105 @@
...
+    # quadruple-wrapping should produce deterministic behavior
+    parallel_seed_test(
+        lambda: aec_to_parallel(
+            parallel_to_aec(
+                aec_to_parallel(
+                    parallel_to_aec(generated_agents_parallel_v0.parallel_env())
+                )
+            )
+        )
+    )
```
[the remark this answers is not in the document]
elliottower: Do you mean to check the parallel env with parallel to aec wrapper functions the same as the other AEC env implementation?

test/all_parameter_combs_test.py
elliottower: I just added env module cause I originally had Hanabi and waterworld and you can say env module in [a, b, c] instead of having to repeatedly saying if a not in name and b not in name and c not in name. But I can change this back cause I can see it being more confusing and less dynamic (if we update the version to v5 for example)

test/variable_env_test.py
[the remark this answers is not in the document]
jjshoots: Yep!

test/all_parameter_combs_test.py
elliottower: Fixed this in new commit

jjshoots (approved):

PR #951, opened 2023-04-21
Remove all local imports from PZ

pseudo-rnd-thoughts:


pettingzoo/classic/go/go.py
elliottower, line 1: Is pygame not necessary here? I see calls using it so I think it should probably be added back

test/all_parameter_combs_test.py
elliottower, line 1: This is just because we only test on some of the modules, correct?

elliottower (commented):
Looks mostly good besides a few questions. The __all__ parts are a good catch, should’ve definitely been there before

pettingzoo/classic/go/go.py
pseudo-rnd-thoughts: No, as it is already imported in the module

test/all_parameter_combs_test.py
pseudo-rnd-thoughts: Im not sure what your question is? I just replaced the `*` with the actual environments

elliottower: Makes sense, I was pretty sure that was it but just wanted to make sure you weren’t doing something else

elliottower (approved):

PR #953, opened 2023-04-23
Add action masking documentation, update usage scripts

elliottower:
# Description

This PR makes all usage scripts in documentation consistent and uses `break` (rather than `env.step(None)`, which I personally feel is a bit confusing for new users).

I also added action masking documentation to the AEC API page.

I changed the homepage usage script to use `env.action_space(agent).sample()` rather than `policy(observation, agent)` just because it feels like it's cleaner to have code which actually runs, but I can change that back if we want.

> Another consistency thing was I added `# execute the action in the environment` to the usage because that was in Gymnasium and looks nice to just orient users, I can remove that if we don't like it but I think it's a nice touch.

I decided I think this should be removed from both Gymnasium and this (it was added by a new contributor and made the Gym page a bit cluttered, but it's Mark's call, on my other PR in Gymnasium)

Example screenshot of the new homepage:
firefox_oidtqHRFOF

Action masking documentation:
firefox_l2vwkHpwWB

## Type of change

### Screenshots

# Checklist:


jjshoots (changes requested):
LGTM, but since this is documentation, I'd be more confident is someone elsd had a look as well.

PR #964, opened 2023-05-05
Added type hints in utils package

LetteraUnica:
# Description

Type hinted all python modules under pettingzoo/utils. Pyright pre-commit checks were disabled on specific lines of code because they gave an error. More details in screenshots section.

When adding type hints I spotted some bugs, so I also edited some lines of code.

## Type of change

### Screenshots

image
Error present even if we check attribute presence beforehand.

image
image
Same error but different way to check attribute presence. There are a few more errors like these, I won't show all of them here.

image
It says that `action` could be None, even if we perform a check at line 31.

image
In this case `space` is an instance of gymnasium.spaces.Space that does not have .low and .high attributes. However, we check in `__init__` that the action spaces are Box spaces which in turn have the .low and .high attributes.

# Checklist:


elliottower:
Finally got a chance to look this over a bit, thanks for helping out with this. I see there are a few TypeErrors to fix (https://github.com/Farama-Foundation/PettingZoo/actions/runs/4897571285/jobs/8745765176?pr=964):
```
 ERROR test/all_parameter_combs_test.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/doc_examples_test.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/pickle_test.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/pytest_runner_test.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/specific_env_test.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/unwrapped_test.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
ERROR test/variable_env_test.py - TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
!!!!!!!!!!!!!!!!!!! Interrupted: 7 errors during collection !!!!!!!!!!!!!!!!!!!!
```

LetteraUnica:
Yeah I saw them, however they are only present in Python version <3.10, because all the tests for python 3.10 pass. Do you think I should substitute the | operator with Union?
For example, `a: int | None` -> `a: Union[int, None]`

elliottower:
The | should work fine but you need to do from __futute__ import
annotations I believe (in other files we use those and they work fine)

LetteraUnica:
I imported annotations in all files apart from clip_out_of_bounds_wrapper which caused the error. Pushed the changes and now it all seems ok.

I forgot to mention that I haven't added type hints to utils/conversions as I am not familiar with that part of the library and the code seems very complicated.

pettingzoo/utils/average_total_reward.py
```diff
@@ -1,9 +1,15 @@
...
 import random
 
 import numpy as np
 
+from pettingzoo.utils.env import AECEnv
+
 
-def average_total_reward(env, max_episodes=100, max_steps=10000000000):
+def average_total_reward(
+    env: AECEnv, max_episodes: int = 100, max_steps: int = 10000000000
```
elliottower, line 11: Is this only for AEC?

pettingzoo/utils/capture_stdout.py
```diff
@@ -14,10 +14,10 @@ class capture_stdout:
         data = var.getvalue()
     """
 
-    def __init__(self):
+    def __init__(self) -> None:
```
elliottower, line 17: I don’t think init needs this, afaik you never return anything from init

pettingzoo/utils/env.py
```diff
@@ -262,17 +262,17 @@ def __iter__(self):
 
 
 class AECIterator(Iterator):
-    def __init__(self, env, max_iter):
+    def __init__(self, env: AECEnv, max_iter: int) -> None:
         self.env = env
         self.iters_til_term = max_iter
 
-    def __next__(self):
+    def __next__(self) -> str:
```
elliottower, line 269: I’m not 100% sure agent selection has to be a string cause the user can input it as whatever they want

elliottower: Also like with init I’ve never seen return type specified for double underscore methods like this

pettingzoo/utils/wrappers/base.py
```diff
@@ -60,7 +64,7 @@ def observation_spaces(self):
             ) from e
 
     @property
-    def action_spaces(self):
+    def action_spaces(self) -> dict[str, gymnasium.spaces.Space]:
```
elliottower, line 67: We should probably check if gymnasium has this removed yet and if so we remove it too

pettingzoo/utils/wrappers/base.py
```diff
@@ -98,13 +102,13 @@ def reset(self, seed=None, options=None):
         self.agents = self.env.agents
         self._cumulative_rewards = self.env._cumulative_rewards
 
-    def observe(self, agent):
+    def observe(self, agent: str) -> ObsType | None:
         return self.env.observe(agent)
 
-    def state(self):
+    def state(self) -> np.ndarray:
```
elliottower, line 108: Not sure if we restrict state to be a given type, it seems like we should let users do what they want

pettingzoo/utils/wrappers/base.py
```diff
@@ -71,23 +75,23 @@ def action_spaces(self):
...
+    def observation_space(self, agent: str) -> gymnasium.spaces.Space:
         return self.env.observation_space(agent)
 
-    def action_space(self, agent):
+    def action_space(self, agent: str) -> gymnasium.spaces.Space:
         return self.env.action_space(agent)
 
     @property
-    def unwrapped(self):
+    def unwrapped(self) -> AECEnv:
```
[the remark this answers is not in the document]
elliottower: Oh I’m dumb lol there’s base parallel

pettingzoo/utils/wrappers/base_parallel.py
```diff
@@ -17,37 +22,47 @@ def __init__(self, env):
 
         # Not every environment has the .state_space attribute implemented
         try:
-            self.state_space = self.env.state_space
+            self.state_space = self.env.state_space  # type: ignore
```
elliottower, line 25: Is this the best way? I think pyright ignore is better as it specifies which check it is for

pettingzoo/utils/wrappers/terminate_illegal.py
```diff
@@ -9,23 +12,23 @@ class TerminateIllegalWrapper(BaseWrapper):
...
         self._prev_obs = None
 
-    def reset(self, seed=None, options=None):
+    def reset(self, seed: int | None = None, options: dict | None = None) -> None:
         self._terminated = False
         self._prev_obs = None
         super().reset(seed=seed, options=options)
 
-    def observe(self, agent):
+    def observe(self, agent: str) -> ObsType | None:
```
elliottower, line 25: Can it be none?

pettingzoo/utils/wrappers/terminate_illegal.py
```diff
@@ -39,7 +42,7 @@ def step(self, action):
             self.terminations[self.agent_selection]
             or self.truncations[self.agent_selection]
         ):
-            self._was_dead_step(action)
+            self._was_dead_step(None)
```
elliottower, line 45: I’m not 100% sure how the dead step functionality works but we should standardize all the code, as the examples and envs use was dead(action) but it might not be necessary

pettingzoo/utils/wrappers/terminate_illegal.py
jjshoots: Yeah it can be None, just one of the more diabolical things.

pettingzoo/utils/wrappers/base_parallel.py
```diff
@@ -17,37 +22,47 @@ def __init__(self, env):
...
+    def close(self) -> None:
         return self.env.close()
 
     @property
-    def unwrapped(self):
+    def unwrapped(self) -> ParallelEnv:
         return self.env.unwrapped
 
-    def state(self):
+    def state(self) -> np.ndarray:
```
jjshoots, line 61: It's possible for this to be `Any` as well, ie: users can actually let this be a dict.

pettingzoo/utils/wrappers/base_parallel.py
```diff
@@ -1,12 +1,17 @@
...
+import numpy as np
 from gymnasium.utils import seeding
 
-from pettingzoo.utils.env import ParallelEnv
+from pettingzoo.utils.env import ActionType, ObsType, ParallelEnv
 
 
 class BaseParallelWrapper(ParallelEnv):
-    def __init__(self, env):
+    def __init__(self, env: ParallelEnv) -> None:
```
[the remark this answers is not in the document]
jjshoots: Yes, in general no need to type hint return None if it doesn't return anything.

pettingzoo/utils/wrappers/base.py
```diff
@@ -98,13 +102,13 @@ def reset(self, seed=None, options=None):
...
-    def observe(self, agent):
+    def observe(self, agent: str) -> ObsType | None:
         return self.env.observe(agent)
 
-    def state(self):
+    def state(self) -> np.ndarray:
         return self.env.state()
 
-    def step(self, action):
+    def step(self, action: ActionType) -> None:
```
jjshoots, line 111: Don't have to typehint None.

pettingzoo/utils/wrappers/base.py
```diff
@@ -71,23 +75,23 @@ def action_spaces(self):
...
-    def close(self):
+    def close(self) -> None:
         self.env.close()
 
-    def render(self):
+    def render(self) -> None | np.ndarray | str | list:
         return self.env.render()
 
-    def reset(self, seed=None, options=None):
+    def reset(self, seed: int | None = None, options: dict | None = None) -> None:
```
jjshoots, line 94: Don't have to typehint None.

pettingzoo/utils/wrappers/base.py
```diff
@@ -71,23 +75,23 @@ def action_spaces(self):
...
+    def action_space(self, agent: str) -> gymnasium.spaces.Space:
         return self.env.action_space(agent)
 
     @property
-    def unwrapped(self):
+    def unwrapped(self) -> AECEnv:
         return self.env.unwrapped
 
-    def close(self):
+    def close(self) -> None:
```
jjshoots, line 88: Don't have to typehint None.

pettingzoo/utils/wrappers/base.py
```diff
@@ -9,7 +15,7 @@ class BaseWrapper(AECEnv):
     All AECEnv wrappers should inherit from this base class
     """
 
-    def __init__(self, env):
+    def __init__(self, env: AECEnv) -> None:
```
jjshoots, line 18: Don't have to typehint None.

pettingzoo/utils/wrappers/assert_out_of_bounds.py
```diff
@@ -1,19 +1,22 @@
...
-    def __init__(self, env):
+    def __init__(self, env: AECEnv) -> None:
         super().__init__(env)
         assert all(
             isinstance(self.action_space(agent), Discrete)
             for agent in getattr(self, "possible_agents", [])
         ), "should only use AssertOutOfBoundsWrapper for Discrete spaces"
 
-    def step(self, action):
+    def step(self, action: ActionType) -> None:
```
jjshoots, line 19: Don't have to typehint None.

pettingzoo/utils/wrappers/assert_out_of_bounds.py
```diff
@@ -1,19 +1,22 @@
...
 
+from pettingzoo.utils.env import ActionType, AECEnv
 from pettingzoo.utils.wrappers.base import BaseWrapper
 
 
 class AssertOutOfBoundsWrapper(BaseWrapper):
     """Asserts if the action given to step is outside of the action space. Applied in PettingZoo environments with discrete action spaces."""
 
-    def __init__(self, env):
+    def __init__(self, env: AECEnv) -> None:
```
jjshoots, line 12: Don't have to typehint None.

pettingzoo/utils/save_observation.py
```diff
@@ -23,7 +27,12 @@ def _check_observation_saveable(env, agent):
 
 # save the observation of an agent. If agent not specified uses env selected agent. If all_agents
 # then all agents in environment observation recorded.
-def save_observation(env, agent=None, all_agents=False, save_dir=os.getcwd()):
+def save_observation(
```
jjshoots, line 30: Don't have to typehint None.

pettingzoo/utils/save_observation.py
```diff
@@ -1,10 +1,14 @@
...
 import os
 
 import gymnasium.spaces
 import numpy as np
 
+from pettingzoo.utils.env import AECEnv, AgentID, ParallelEnv
+
 
-def _check_observation_saveable(env, agent):
+def _check_observation_saveable(env: AECEnv | ParallelEnv, agent: AgentID) -> None:
```
jjshoots, line 11: Don't have to typehint None.

pettingzoo/utils/env_logger.py
```diff
@@ -98,7 +106,7 @@ def __init__(self, *args, mqueue, **kwargs):
         logging.Handler.__init__(self, *args, **kwargs)
         self.mqueue = mqueue
 
-    def emit(self, record):
+    def emit(self, record: logging.LogRecord) -> None:
```
jjshoots, line 109: Don't have to typehint None.

pettingzoo/utils/save_observation.py
```diff
@@ -23,7 +27,12 @@ def _check_observation_saveable(env, agent):
 
 # save the observation of an agent. If agent not specified uses env selected agent. If all_agents
 # then all agents in environment observation recorded.
-def save_observation(env, agent=None, all_agents=False, save_dir=os.getcwd()):
+def save_observation(
+    env: AECEnv,
```
RedTachyon, line 31: Can't this also be ParallelEnv?

pettingzoo/utils/save_observation.py
```diff
@@ -39,7 +48,8 @@ def save_observation(env, agent=None, all_agents=False, save_dir=os.getcwd()):
         os.makedirs(save_folder, exist_ok=True)
 
         observation = env.observe(a)
-        rescaled = observation.astype(np.uint8)
-        im = Image.fromarray(rescaled)
-        fname = os.path.join(save_folder, str(a) + ".png")
-        im.save(fname)
+        if observation is not None:
```
RedTachyon, line 51: Is `observation is None` a valid value in this context? My intuition is that it should never happen, so it should raise an error (as it would right now when you try to do `None.astype`)

With this change, this error would just quietly be ignored

pettingzoo/utils/wrappers/base.py
```diff
@@ -34,19 +40,17 @@ def __init__(self, env):
             pass
 
         # Not every environment has the .state_space attribute implemented
-        try:
-            self.state_space = self.env.state_space
-        except AttributeError:
-            pass
+        if hasattr(self.env, "state_space"):
+            self.state_space = self.env.state_space  # type: ignore
```
RedTachyon, line 44: Why do we need a `type: ignore` here? It might be better to stick with the previous version if that's necessary

pettingzoo/utils/wrappers/base_parallel.py
```diff
@@ -17,37 +22,47 @@ def __init__(self, env):
...
         try:
-            self.state_space = self.env.state_space
+            self.state_space = self.env.state_space  # type: ignore
         except AttributeError:
             pass
 
-    def reset(self, seed=None, options=None):
+    def reset(
+        self, seed: int | None = None, options: dict | None = None
+    ) -> tuple[Any, Any]:
```
RedTachyon, line 31: I think we can narrow down the return type here?

pettingzoo/utils/wrappers/capture_stdout.py
```diff
@@ -1,23 +1,25 @@
...
 class CaptureStdoutWrapper(BaseWrapper):
     """Takes an environment which prints to terminal, and gives it an `ansi` render mode where it captures the terminal output and returns it as a string instead."""
 
-    def __init__(self, env):
+    def __init__(self, env: AECEnv) -> None:
+        assert hasattr(env, "render_mode"), f"Environment {env} has no render_mode."
         assert (
-            env.render_mode == "human"
-        ), f"CaptureStdoutWrapper works only with human rendering mode, but found {env.render_mode} instead."
+            env.render_mode == "human"  # type: ignore
```
RedTachyon, line 12: Instead of ignoring type issues, add explicit assertions

pettingzoo/utils/wrappers/terminate_illegal.py
RedTachyon, line 45: This is meant to be a type hint fixing PR, why was this change made? Imo revert it unless there's a good reason otherwise

pettingzoo/utils/wrappers/base.py
elliottower: Ah right. Well in a few other files we have already explicitly deprecated the use of env.action_spaces(), so we might want to make a deprecation note here or remove it altogether.

References: https://github.com/Farama-Foundation/PettingZoo/blob/a825e94262eef7b1cb617bb54abfca3df68d0acf/pettingzoo/utils/conversions.py#L128
https://github.com/Farama-Foundation/PettingZoo/blob/a825e94262eef7b1cb617bb54abfca3df68d0acf/pettingzoo/utils/wrappers/base.py#L50
https://github.com/Farama-Foundation/PettingZoo/blob/a825e94262eef7b1cb617bb54abfca3df68d0acf/pettingzoo/utils/wrappers/base_parallel.py#L50

pettingzoo/utils/average_total_reward.py
LetteraUnica: I think so, because the function calls env,last() which is only present in AEC environments

pettingzoo/utils/capture_stdout.py
LetteraUnica: Removed

pettingzoo/utils/wrappers/base.py
LetteraUnica: There is already a deprecation warning for action spaces

https://github.com/Farama-Foundation/PettingZoo/blob/a825e94262eef7b1cb617bb54abfca3df68d0acf/pettingzoo/utils/wrappers/base.py#L64

pettingzoo/utils/wrappers/base_parallel.py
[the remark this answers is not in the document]
LetteraUnica: Removed

pettingzoo/utils/wrappers/terminate_illegal.py
LetteraUnica: The problem is action can be different than None even if truncations or terminations are True, so pyright gives an error. Possible solutions could be:

* The one I applied above
* Checking explicitly if action is None
* Ignoring the pyright error

Let me know what you prefer

pettingzoo/utils/wrappers/base_parallel.py
LetteraUnica: As I explained above for `BaseWrapper`, I copied the type returned by `ParallelEnv.state` method. If you want I can change both `ParallelEnv.state` and this method to return `Any`.

pettingzoo/utils/wrappers/base.py
LetteraUnica: Removed

pettingzoo/utils/wrappers/base.py
LetteraUnica: Removed

pettingzoo/utils/wrappers/assert_out_of_bounds.py
LetteraUnica: Removed

pettingzoo/utils/save_observation.py
LetteraUnica: Removed

pettingzoo/utils/save_observation.py
LetteraUnica: Removed

pettingzoo/utils/env_logger.py
LetteraUnica: Removed

pettingzoo/utils/save_observation.py
LetteraUnica: I don't think so, because the function calls `env.observe` which is only present in `AECEnv`

pettingzoo/utils/save_observation.py
LetteraUnica: I added the check because pyright complained about using `.astype` on a None value.

Other possibilities are:
* ignore the error with `# pyright: ignore`.
* check if observation is None and if so raise an error.

Let me know what you prefer

pettingzoo/utils/wrappers/base.py
LetteraUnica: Reverted. I tried to change it with hasattr to see if the pyright error went away and then forgot to revert it.

pettingzoo/utils/wrappers/capture_stdout.py
LetteraUnica: There is an assertion in the line before, but pyright still thinks that `env.render_mode` might not be present.

image

pettingzoo/utils/wrappers/terminate_illegal.py
LetteraUnica: Reverted

pettingzoo/utils/wrappers/base.py
elliottower: @jjshoots @RedTachyon any thoughts on this?

pettingzoo/utils/average_total_reward.py
elliottower: Maybe add a comment explaining that? Just so in the future we don’t think this is a bug

pettingzoo/utils/save_observation.py
elliottower: Can you test this out to be sure? I feel like observe might be implicitly defined or something (I guess no last call though). Not sure why this is the case to be honest.

pettingzoo/utils/wrappers/capture_stdout.py
elliottower: Sounds like pyright being dumb (what else is new) but you should change the type ignore to pyright ignore report MessageTypeHere or whatever the syntax is

pettingzoo/utils/wrappers/assert_out_of_bounds.py
LetteraUnica: Removed

pettingzoo/utils/wrappers/capture_stdout.py
RedTachyon: `render_mode` currently doesn't seem to exist in `AECEnv`, so there's something iffy going on here

pettingzoo/utils/average_total_reward.py
LetteraUnica: Changed the docstring from:

`Calculates the average total reward over the episodes`

To

`Calculates the average total reward over the episodes for AEC Environments`

pettingzoo/utils/wrappers/capture_stdout.py
elliottower: > Instead of ignoring type issues, add explicit assertions

We should probably do a pr one day fixing the ignored type errors to be assertions cause I know there’s a good amount

pettingzoo/utils/wrappers/terminate_illegal.py
elliottower: All of our example code uses these same three lines with dead step(action), is it possible to make the dead step function signature have type hints allowing None?

pettingzoo/utils/save_observation.py
LetteraUnica: I checked the class ParallelEnv and it doesn't define .observe() nor inherits from another class. I can allow the argument to be a ParallelEnv and ignore the pyright error.

pettingzoo/utils/save_observation.py
LetteraUnica: Added an assertion error if observation is None

pettingzoo/utils/wrappers/base_parallel.py
LetteraUnica: Changed reset to return `tuple[ObsDict, dict[str, dict]]`

pettingzoo/utils/save_observation.py
elliottower: In that case then I think it’s fine, maybe make a note that observe() isn’t implemented by parallel envs

LetteraUnica: Added a comment before calling env,observe.
https://github.com/LetteraUnica/PettingZoo/blob/type-hints/pettingzoo/utils/save_observation.py#L50

pettingzoo/utils/wrappers/terminate_illegal.py
LetteraUnica: I confirm that  `_was_dead_step(None)` is only used in the code above, so I reverted it to `_was_dead_step(action)` and ignored the pyright error.

elliottower:
Looks like all the comments are resolved, I’ll rise one final look through today but probably good to go. Would you mind making a separate PR changing the self state to be ndarray or any? Or I guess just any. I think I’ve seen something or any before as a way of suggesting but not enforcing

LetteraUnica:
Just pushed the last changes, I think we can close this PR as well.

In subsequent PRs what remains to do is:
* Make `env.state` return Any.
* Add assertions instead of ignoring pyright type errors.

Should I open an issue as a reminder? Because I don't think I'll have time to do this till next weekend.

pettingzoo/utils/deprecated_module.py
```diff
@@ -1,14 +1,18 @@
+from __future__ import annotations
+
 import importlib.util
 import pkgutil
 import re
+from types import ModuleType
```
elliottower, line 6: Are either ModuleType or Iterable used here?

pettingzoo/utils/save_observation.py
```diff
@@ -1,10 +1,14 @@
...
 import os
 
 import gymnasium.spaces
 import numpy as np
 
+from pettingzoo.utils.env import AECEnv, AgentID, ParallelEnv
+
 
-def _check_observation_saveable(env, agent):
+def _check_observation_saveable(env: AECEnv | ParallelEnv, agent: AgentID):
```
elliottower, line 11: These should have -> None to match all the other parts where you do return None (see https://stackoverflow.com/questions/48223558/typehints-none-or-leave-blank)

pettingzoo/utils/save_observation.py
elliottower, line 30: Return annotation

pettingzoo/utils/wrappers/base.py
```diff
@@ -35,18 +40,20 @@ def __init__(self, env):
...
         try:
-            self.state_space = self.env.state_space
+            self.state_space = (
+                self.env.state_space  # pyright: ignore[reportGeneralTypeIssues]
+            )
         except AttributeError:
             pass
 
-    def __getattr__(self, name):
+    def __getattr__(self, name: str):
```
elliottower, line 49: I think this is fine because double underscore private methods don’t generally have return types

pettingzoo/utils/wrappers/base.py
```diff
@@ -71,23 +78,23 @@ def action_spaces(self):
...
     @property
-    def unwrapped(self):
+    def unwrapped(self) -> AECEnv:
         return self.env.unwrapped
 
     def close(self):
         self.env.close()
 
-    def render(self):
+    def render(self) -> None | np.ndarray | str | list:
```
elliottower, line 94: Should this be Any? What if the user wants to do a custom way of rendering which isnt str list or ndarray? Like a torch tensor or jax or something? Maybe @jjshoots or @RedTachyon have thoughts

pettingzoo/utils/wrappers/base.py
```diff
@@ -98,13 +105,13 @@ def reset(self, seed=None, options=None):
...
-    def observe(self, agent):
+    def observe(self, agent: str) -> ObsType | None:
         return self.env.observe(agent)
 
-    def state(self):
+    def state(self) -> np.ndarray:
         return self.env.state()
 
-    def step(self, action):
+    def step(self, action: ActionType):
```
elliottower, line 114: return type annotation

pettingzoo/utils/wrappers/capture_stdout.py
```diff
@@ -1,23 +1,25 @@
...
 class CaptureStdoutWrapper(BaseWrapper):
     """Takes an environment which prints to terminal, and gives it an `ansi` render mode where it captures the terminal output and returns it as a string instead."""
 
-    def __init__(self, env):
+    def __init__(self, env: AECEnv):
+        assert hasattr(env, "render_mode"), f"Environment {env} has no render_mode."
         assert (
-            env.render_mode == "human"
-        ), f"CaptureStdoutWrapper works only with human rendering mode, but found {env.render_mode} instead."
+            env.render_mode == "human"  # pyright: ignore[reportGeneralTypeIssues]
```
elliottower, line 12: this first one isn’t needed, the only time when pyright comments like this will be read is at the end of the parentheses (line 13)

pettingzoo/utils/wrappers/base.py
elliottower: Actually nvm later on you use -> Any for getattr, might as well to be honest, anything besides init

pettingzoo/utils/wrappers/order_enforcing.py
```diff
@@ -106,15 +112,18 @@ def __str__(self):
...
 class AECOrderEnforcingIterator(AECIterator):
-    def __next__(self):
+    def __next__(self) -> str:
         agent = super().__next__()
+        assert hasattr(
+            self.env, "_has_updated"
+        ), "env must be wrapped by OrderEnforcingWrapper"
         assert (
-            self.env._has_updated
+            self.env._has_updated  # pyright: ignore[reportGeneralTypeIssues]
```
elliottower, line 126: same as my comment earlier dont think both are necessary

elliottower (approved):
A few minor nitpicks but overall very good, thanks again for this it is very nice to have.

pettingzoo/utils/wrappers/base.py
elliottower: This would be a separate PR though, same with the action space change

pettingzoo/utils/wrappers/capture_stdout.py
elliottower: Nvm checked and it is requited

pettingzoo/utils/wrappers/order_enforcing.py
elliottower: Nvm checked and it is required.

PR #965, opened 2023-05-06
Update enity.state.p_pos with state.p_vel of previous timestep

### LINKED ISSUE #962: MPE physical dynamics of agents not coherent with original paper
I am aware that the dynamics are copied from the original repo (which is no longer supported) hence why I am creating an issue here. I believe the dynamics not to be coherent with the definition in the paper [Emergence of Grounded Compositional Language in Multi-Agent Populations](https://arxiv.org/abs/1703.04908) (Appendix: Physical State and Dynamics).

https://github.com/Farama-Foundation/PettingZoo/blob/99f050c2b966c2a5dcdc256d764d7a76f388b28a/pettingzoo/mpe/_mpe_utils/core.py#L187

In particular `entity.state.p_pos` at time $t$ should be updated with the older value of `entity.state.p_vel` at time $t-1$.

vateseif:
# Description

This PR updates the dynamics of the agents to be coherent with the original paper [Emergence of Grounded Compositional Language in Multi-Agent Populations](https://arxiv.org/abs/1703.04908) (Appendix: Physical State and Dynamics).
In particular, the line `entity.state.p_pos += entity.state.p_vel * self.dt ` is moved to before `entity.state.p_vel ` is updated making  `p_pos` at time $t$ dependent on `p_vel` at time $t-1$ as specified in the paper.

Fixes #962

## Type of change

# Checklist:

### DIFF
--- a/pettingzoo/mpe/_mpe_utils/core.py
+++ b/pettingzoo/mpe/_mpe_utils/core.py
@@ -168,6 +168,7 @@ class World:  # multi-agent world
         for i, entity in enumerate(self.entities):
             if not entity.movable:
                 continue
+            entity.state.p_pos += entity.state.p_vel * self.dt
             entity.state.p_vel = entity.state.p_vel * (1 - self.damping)
             if p_force[i] is not None:
                 entity.state.p_vel += (p_force[i] / entity.mass) * self.dt
@@ -184,7 +185,6 @@ class World:  # multi-agent world
                         )
                         * entity.max_speed
                     )
-            entity.state.p_pos += entity.state.p_vel * self.dt
 
     def update_agent_state(self, agent):
         # set communication state (directly for now)


pseudo-rnd-thoughts:
Should we not change the version number for all MPE environments and add a note to the documentation on the change

elliottower:
> Should we not change the version number for all MPE environments and add a note to the documentation on the change + any performance regression testing

I made a separate PR for the version numbers (including this change, was just simpler than requesting wire access to this branch). Should add a note in the documentation though that’s a good point.

benblack769:
The current code reflects the reference code of the  paper. 

https://github.com/openai/multiagent-particle-envs/blob/master/multiagent/core.py#L169

However, I agree that the change in the MR makes more sense. It is standard practice to update position before velocity 
Looking at the paper itself, yeah,the change is more in line with the paper than the original code ever was.

benblack769:
Hopefully it is not too big of a change. It is an off-by-one error, after all. But yeah, need new versions for this. The timesteps in MPE are pretty large. The MADDPG paper limits the number of steps to 25.

elliottower:
Merged in #970

PR #966, opened 2023-05-06
Fix Shimmy Openspiel tests

jjshoots:
# Description

This needs to be released before [this](https://github.com/Farama-Foundation/Shimmy/pull/77) will pass.

### DIFF
--- a/pettingzoo/test/api_test.py
+++ b/pettingzoo/test/api_test.py
@@ -334,7 +334,7 @@ def play_test(env, observation_0, num_cycles):
             mask = (
                 prev_observe.get("action_mask")
                 if isinstance(prev_observe, dict)
-                else None
+                else env.infos[agent].get("action_mask")
             )
             action = env.action_space(agent).sample(mask)
 
@@ -410,7 +410,7 @@ def play_test(env, observation_0, num_cycles):
         if terminated or truncated:
             action = None
         else:
-            mask = obs.get("action_mask") if isinstance(obs, dict) else None
+            mask = obs.get("action_mask") if isinstance(obs, dict) else env.infos[agent].get("action_mask")
             action = env.action_space(agent).sample(mask)
         assert isinstance(terminated, bool), "terminated from last is not True or False"
         assert isinstance(truncated, bool), "terminated from last is not True or False"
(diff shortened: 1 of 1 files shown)


pettingzoo/test/api_test.py
```diff
@@ -439,10 +443,13 @@ def test_action_flexibility(env):
         obs, reward, terminated, truncated, info = env.last()
         if terminated or truncated:
             action = None
-        elif isinstance(obs, dict) and "action_mask" in obs:
-            action = env.action_space(agent).sample(obs["action_mask"])
         else:
-            action = 0
+            mask = (
```
elliottower, line 447: I can understand how this works but I feel like it's more clear to do if else rather than the one liner (but I recall someone saying that for testing we don't want super nested things, which I guess is fair)

pettingzoo/test/api_test.py
```diff
@@ -334,7 +334,7 @@ def play_test(env, observation_0, num_cycles):
             mask = (
                 prev_observe.get("action_mask")
                 if isinstance(prev_observe, dict)
-                else None
+                else env.infos[agent].get("action_mask")
```
elliottower, line 337: As below I feel like the .get() command is a bit confusing

pettingzoo/test/api_test.py
elliottower: I fixed this in a different way in my code:
```
        if terminated or truncated:
            action = None
        elif isinstance(prev_observe, dict) and "action_mask" in prev_observe:
            action = env.action_space(agent).sample(prev_observe["action_mask"])
        elif "action_mask" in info:
            action = env.action_space(agent).sample(info["action_mask"])
        else:
            action = env.action_space(agent).sample()
```

I feel like my way makes it much simpler to see what's going on, and is consistent with how the rest of the codebase looks. There's an argument to change the tutorials and rest of the codebase to use get() and explain in the documentation maybe that get returns None if the key doesn't exist, but otherwise I feel like we should do it my way. This code makes it clear that obs can be something other than a dict, but info has to be a dict, and that action mask can be stored in one or the other.

elliottower:
Looks like the python linux tests 3.9 failed with this error:
`FAILED test/variable_env_test.py::test_generated_agents_aec - AssertionError: action is not in action space`

I believe the variable env test calls the API test so maybe that's the issue?

pettingzoo/test/api_test.py
jjshoots: Sure, this makes sense. Changed!

elliottower (approved):

PR #973, opened 2023-05-11
Add API section to every environment with links to source code

elliottower:
I have always found it annoying that there was no direct way to view the full source code from being on an environment page, so I added this. Thanks to the previous PR that gives links to source code for API documentation, this now allows there to be links to source code at the bottom of the file.

# Description

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/docs/_scripts/gen_envs_mds.py
+++ b/docs/_scripts/gen_envs_mds.py
@@ -84,19 +84,27 @@ if __name__ == "__main__":
                 if "rlcard_envs" not in env_dir_path
                 else env_dir_path + ".py"
             )
-            if "rlcard_envs" not in env_dir_path:
+            if "rlcard_envs" in env_dir_path:
+                docs_text += f"""## API
+```{{eval-rst}}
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts:
This is an interesting point for gymnasium, I would consider changing the documentation to use the autoclass reference for the environment class

elliottower:
> This is an interesting point for gymnasium, I would consider changing the documentation to use the autoclass reference for the environment class

I tried a few others like automodule for the entire file (had all the comments at the top and was very messy) and tried specific methods but this was the simplest looking and most straightforward imo. There aren’t really proper classes for each environment as they are defined only as raw env and env, so that’s all I had to work with

elliottower:
Looks like gymnasium is a bit more structured with environments named CartPoleEnv so that would probably be more user friendly than raw_env. I also wasn’t sure if it should be API or Source Code or something, but the other pages like AEC or parallel api pages have their auto docs under an API header at the bottom

elliottower:
I merged this but the build-docs test seems to not have ran and is now failing, is it possible to enable this on every PR?

PR #997, opened 2023-06-09
added insufficient material, 50-moves and 3-fold repetition check

DmytroIvasiuk:
1) now it checks for insufficient material (before two kings would run around the board)  
    >>> is_insufficient_material = self.board.is_insufficient_material()
2) and checks for 3-fold repetition or 50-moves within one command
    >>> can_claim_draw = self.board.can_claim_draw()

### DIFF
--- a/pettingzoo/classic/chess/chess.py
+++ b/pettingzoo/classic/chess/chess.py
@@ -250,10 +250,9 @@ class raw_env(AECEnv):
         is_stale_or_checkmate = not any(next_legal_moves)
 
         # claim draw is set to be true to align with normal tournament rules
-        is_repetition = self.board.is_repetition(3)
-        is_50_move_rule = self.board.can_claim_fifty_moves()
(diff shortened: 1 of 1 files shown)


elliottower:
Took a look through the python chess docs and source code from these functions and looks good, only thing that might be worth looking into is the automatic draw rules from 2014 (fivefold repetition and 75 turn rule, see https://github.com/niklasf/python-chess/blob/736478a1c7cd798cd7a603b92a5ac7dff692349b/chess/__init__.py#L2042 and https://github.com/niklasf/python-chess/blob/736478a1c7cd798cd7a603b92a5ac7dff692349b/chess/__init__.py#L2051). The README mentions that with the new rules the game ends as a draw even if it isn't claimed. As far as I can tell the game will end due to threefold repetition or fifty turn rule here, so those shouldn't be relevant, but maybe I'm misunderstanding something.

elliottower (approved):

PR #998, opened 2023-06-09
Fix classic environments screen sizes, add type hints, fix pre-commit

elliottower:
Added screen_height argument to the rlcard envs and RPS env (they have dynamic widths depending on number of players, i used a slightly smaller default height for RPS as it was 350 previously, whereas rlcard was 1000px), added screen_scaling to connect_four because the rendering in that environment is a bit finnicky and only works for certain resolutions (the original ratio is 99 / 86, and any resolutions which multiple both numbers by integers work, but scaling by 2.5 for example results in white space because of rounding), and added screen_height to chess and go (originally used screen_width but it's simpler to have only two args to try out: screen_height for some envs, or for connect four it's screen_scaling, rather than having to remember 3 different ones for different envs)

I also added type hints to the classic envs as I was already going through and adding new args.

Almost all of the files had # noqa which ignores checks on the files globally (i.e., when doing new code it didn't even re-format according to black unless you removed the line), so I replaced that with specific ignores for the two pydocstring errors which are a result of us using the large docstrings at the top of environments to dynamically create the environment pages (starting with an H1 tag instead of full sentence with period).

I did change the default size of RPS to be bigger and go and connect to be slightly smaller and more normalized with everything else, but didn't change underlying envrionments so don't think it should need a new version

# Description

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/pettingzoo/atari/basketball_pong/basketball_pong.py
+++ b/pettingzoo/atari/basketball_pong/basketball_pong.py
@@ -1,4 +1,4 @@
-# noqa
+# noqa: D212, D415
 """
 # Basketball Pong
 
--- a/pettingzoo/atari/boxing/boxing.py
+++ b/pettingzoo/atari/boxing/boxing.py
@@ -1,4 +1,4 @@
-# noqa
+# noqa: D212, D415
 """
 # Boxing
 
--- a/pettingzoo/atari/combat_plane/combat_plane.py
+++ b/pettingzoo/atari/combat_plane/combat_plane.py
@@ -1,4 +1,4 @@
-# noqa
+# noqa: D212, D415
 """
 # Combat: Plane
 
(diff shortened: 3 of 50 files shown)


elliottower:
I may try to get dynamic scaling of pygame windows to work so users can just resize or hit maximize to change it without using arguments, think it's possible but may take a bit of work

elliottower:
Also need to look at the render FPS and make the environments go at the same speed, currently connect four for example goes incredibly fast and ends in like half a second, whereas go pauses for a second between moves or something like that, need to make it more consistent

pseudo-rnd-thoughts (commented):
Does this screen size change effect `rgb_array`?

pseudo-rnd-thoughts (approved):

PR #999, opened 2023-06-16
Update chess to v6, fix all envs to render at correct FPS, clean up pygame code

elliottower:
# Description

There was an issue in the discord about chess having the wrong first player, turns out the issue was a typo relating to the rendering images. I also normalized all of the classic environments to actually tick according to their specified render FPS, as most of them just ran as fast as possible despite saying they had a given FPS. 

This will likely mean all of them will have to use EzPickle, but I'm waiting until that test fails to change it.
Edit: they did not fail because the pickle test did not use render mode human, therefore the pygame clock was not initialized, but I figure we might as well check that things can be pickled no matter what. Previously in shimmy I had some issues pickling rendered environments, so it's good that this works. I think if you try to pickle and unpickle an env it doesn't directly work with rendering naively but I think you can do env.reset() and it will work? 

Not sure if we can do a warning if people try to pickle rendered envs, maybe it's fine to just tell them not to do that.

## Type of change

### Screenshots

# Checklist:


elliottower:
This ended up being a large number of changes in one PR so apologies for that, just figured I'd do pretty much all of the pygame related fixes here.

elliottower:
Cooperative pong had some very weird bugs due to the observations using pygame as well as the rendering, but the tests now pass locally so I think it should be good.

Issue #41, opened 2020-03-25
Chess bug


benblack769:
The following code prints out a completely invalid board state. Probably a bug in python-chess, but rather hard to track down because it is fairly rare. 

```python
from pettingzoo.classic import chess
import time
import random

env = chess.env()

orig_obs = env.reset()
game_len = 0
for x in range(100000):
    reward, done, info = env.last()

    game_len += 1

    if game_len > 1000:
        print("long game")
        print(env.board)
        exit(0)
    if done:
        orig_obs = env.reset()
        game_len = 0
        break
    action = random.choice(env.infos[env.agent_selection]['legal_moves'])

    next_obs = env.step(action)
```

PR #1004, opened 2023-06-20
Fix chess to white perspective, fix observation bug, add documentation

jacob975:
# Description

Fix the issue # 992. In this pull request, the observations in `env.board_history` is now always toward the white player. However, users can still get observations orientated toward the black player by  `env.observe('player_1")`.

Fix the issue where the env.observe("player_x")["observation"] is left-side right mirroring compare to `env.render()`.
This issue is contributed by two bugs. Firstly, in the `pettingzoo/classic/chess/chess_utils.py#L286`, It should be `OURS=1` but `OURS=0` in the last version. Vice versa for`THEIRS`. Secondly, the function `boards_to_ndarray` in `pettingzoo/classic/chess/chess_utils.py#L5` output an array `boardimage` but incorrectly with a 180 degree rotation. We use np.flip to fix it.

Corresponding updates on `pettingzoo/classic/chess/chess_utils.py#L330-#L344` to keep the en passant flag working.

Add a few type hints for functions in `pettingzoo/classic/chess/chess_utils.py` and `pettingzoo/classic/chess/chess.py`

Revise comments in `pettingzoo/classic/chess/chess.py` to match the current situation, especially on the board orientation and env.observe().

## Type of change

### Screenshots

# Checklist:

- [x ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [x ] I have run `pytest -v` and no errors are present.
- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] I have made corresponding changes to the documentation
- [x ] I solved any possible warnings that `pytest -v` has generated that are related to my code to the best of my knowledge.
- [x ] I have added tests that prove my fix is effective or that my feature works
- [x ] New and existing unit tests pass locally with my changes

### DIFF
--- a/pettingzoo/classic/chess/chess.py
+++ b/pettingzoo/classic/chess/chess.py
@@ -18,7 +18,7 @@ This environment is part of the <a href='..'>classic environments</a>. Please re
 | Agents             | 2                                  |
 | Action Shape       | Discrete(4672)                     |
 | Action Values      | Discrete(4672)                     |
-| Observation Shape  | (8,8,20)                           |
+| Observation Shape  | (8,8,111)                          |
 | Observation Values | [0,1]                              |
 
 
@@ -28,7 +28,7 @@ Chess is one of the oldest studied games in AI. Our implementation of the observ
 
 The observation is a dictionary which contains an `'observation'` element which is the usual RL observation described below, and an  `'action_mask'` which holds the legal moves, described in the Legal Actions Mask section.
 
-Like AlphaZero, the main observation space is an 8x8 image representing the board. It has 20 channels representing:
+Like AlphaZero, the main observation space is an 8x8 image representing the board. It has 111 channels representing:
 
 * Channels 0 - 3: Castling rights:
   * Channel 0: All ones if white can castle queenside
@@ -41,10 +41,13 @@ Like AlphaZero, the main observation space is an 8x8 image representing the boar
 * Channel 7 - 18: One channel for each piece type and player color combination. For example, there is a specific channel that represents black knights. An index of this channel is set to 1 if a black knight is in the corresponding spot on the game board, otherwise, it is set to 0. En passant
 possibilities are represented by displaying the vulnerable pawn on the 8th row instead of the 5th.
 * Channel 19: represents whether a position has been seen before (whether a position is a 2-fold repetition)
+* Channel 20 - 111 represents the previous 7 boards, with each board represented by 13 channels. The latest board occupies the first 13 channels, followed by the second latest board, and so on. These 13 channels correspond to channels 7 - 20.
 
-Like AlphaZero, the board is always oriented towards the current agent (the currant agent's king starts on the 1st row). In other words, the two players are looking at mirror images of the board, not the same board.
+Similar to AlphaZero, our observation space follows a stacking approach, where it accumulates the previous 8 board observations.
 
-Unlike AlphaZero, the observation space does not stack the observations previous moves by default. This can be accomplished using the `frame_stacking` argument of our wrapper.
+Unlike AlphaZero, where the board orientation may vary, in our system, the `env.board_history` always maintains the orientation towards the white agent, with the white agent's king consistently positioned on the 1st row. In simpler terms, both players are observing the same board layout.
+
+Nevertheless, we have incorporated a convenient feature, the env.observe('player_1') function, specifically for the black agent's orientation. This facilitates the training of agents capable of playing proficiently as both black and white.
 
 #### Legal Actions Mask
 
@@ -202,10 +205,21 @@ class raw_env(AECEnv, EzPickle):
         return self.action_spaces[agent]
 
     def observe(self, agent):
-        observation = chess_utils.get_observation(
-            self.board, self.possible_agents.index(agent)
-        )
+        agent_index = self.possible_agents.index(agent)
+
+        observation = chess_utils.get_observation(self.board, agent_index)
         observation = np.dstack((observation[:, :, :7], self.board_history))
+        # We need to swap the white 6 channels with black 6 channels
+        if agent_index == 1:
+            # 1. Mirror the board
+            observation = np.flip(observation, axis=0)
+            # 2. Swap the white 6 channels with the black 6 channels
+            for i in range(1, 9):
+                tmp = observation[..., 13 * i - 6 : 13 * i].copy()
+                observation[..., 13 * i - 6 : 13 * i] = observation[
+                    ..., 13 * i : 13 * i + 6
+                ]
+                observation[..., 13 * i : 13 * i + 6] = tmp
         legal_moves = (
             chess_utils.legal_moves(self.board) if agent == self.agent_selection else []
         )
@@ -272,7 +286,7 @@ class raw_env(AECEnv, EzPickle):
         self._accumulate_rewards()
 
         # Update board after applying action
-        next_board = chess_utils.get_observation(self.board, current_agent)
+        next_board = chess_utils.get_observation(self.board, 0)
         self.board_history = np.dstack(
             (next_board[:, :, 7:], self.board_history[:, :, :-13])
         )
--- a/pettingzoo/classic/chess/chess.py
+++ b/pettingzoo/classic/chess/chess.py
@@ -205,12 +205,12 @@ class raw_env(AECEnv, EzPickle):
         return self.action_spaces[agent]
 
     def observe(self, agent):
-        agent_index = self.possible_agents.index(agent)
+        current_index = self.possible_agents.index(agent)
 
-        observation = chess_utils.get_observation(self.board, agent_index)
+        observation = chess_utils.get_observation(self.board, current_index)
         observation = np.dstack((observation[:, :, :7], self.board_history))
         # We need to swap the white 6 channels with black 6 channels
-        if agent_index == 1:
+        if current_index == 1:
             # 1. Mirror the board
             observation = np.flip(observation, axis=0)
             # 2. Swap the white 6 channels with the black 6 channels
@@ -286,7 +286,8 @@ class raw_env(AECEnv, EzPickle):
         self._accumulate_rewards()
 
         # Update board after applying action
-        next_board = chess_utils.get_observation(self.board, 0)
+        # We always take the perspective of the white agent
+        next_board = chess_utils.get_observation(self.board, player=0)
         self.board_history = np.dstack(
             (next_board[:, :, 7:], self.board_history[:, :, :-13])
         )
--- a/pettingzoo/classic/chess/chess.py
+++ b/pettingzoo/classic/chess/chess.py
@@ -38,10 +38,11 @@ Like AlphaZero, the main observation space is an 8x8 image representing the boar
 * Channel 4: Is black or white
 * Channel 5: A move clock counting up to the 50 move rule. Represented by a single channel where the *n* th element in the flattened channel is set if there has been *n* moves
 * Channel 6: All ones to help neural networks find board edges in padded convolutions
-* Channel 7 - 18: One channel for each piece type and player color combination. For example, there is a specific channel that represents black knights. An index of this channel is set to 1 if a black knight is in the corresponding spot on the game board, otherwise, it is set to 0. En passant
-possibilities are represented by displaying the vulnerable pawn on the 8th row instead of the 5th.
+* Channel 7 - 18: One channel for each piece type and player color combination. For example, there is a specific channel that represents black knights. An index of this channel is set to 1 if a black knight is in the corresponding spot on the game board, otherwise, it is set to 0.
+Similar to LeelaChessZero, en passant possibilities are represented by displaying the vulnerable pawn on the 8th row instead of the 5th.
 * Channel 19: represents whether a position has been seen before (whether a position is a 2-fold repetition)
 * Channel 20 - 111 represents the previous 7 boards, with each board represented by 13 channels. The latest board occupies the first 13 channels, followed by the second latest board, and so on. These 13 channels correspond to channels 7 - 20.
+
 
 Similar to AlphaZero, our observation space follows a stacking approach, where it accumulates the previous 8 board observations.
 
--- a/pettingzoo/classic/chess/chess_utils.py
+++ b/pettingzoo/classic/chess/chess_utils.py
@@ -5,11 +5,12 @@ import numpy as np
 def boards_to_ndarray(boards):
     arr64 = np.array(boards, dtype=np.uint64)
     arr8 = arr64.view(dtype=np.uint8)
-    bits = np.unpackbits(arr8)
+    # a bit array increment from LHS to RHS
+    bits = np.unpackbits(arr8, bitorder="little")
     floats = bits.astype(bool)
     boardstack = floats.reshape([len(boards), 8, 8])
-    # We do np.flip() onto `boardstack` because the board is 180 degrees rotated after the process above.
-    boardimage = np.flip(np.transpose(boardstack, [1, 2, 0]), axis=[0, 1])
+    # We do np.flip() onto `boardstack` because the 1st line of the boardimage is the 8th line of the ndarray.
+    boardimage = np.flip(np.transpose(boardstack, [1, 2, 0]), axis=0)
     return boardimage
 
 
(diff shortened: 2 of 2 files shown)


pettingzoo/classic/chess/chess.py
```diff
@@ -202,10 +205,21 @@ def action_space(self, agent):
         return self.action_spaces[agent]
 
     def observe(self, agent):
-        observation = chess_utils.get_observation(
-            self.board, self.possible_agents.index(agent)
-        )
+        agent_index = self.possible_agents.index(agent)
```
elliottower, line 208: I think you should make the variable naming here consistent with the step function, which does the following:
```
current_agent = self.agent_selection
current_index = self.agents.index(current_agent)
```
Note that self.possible_agents always has both agents, whereas self.agents can have one agent removed if they do an illegal move and terminate, which would make the next observation incorrect.
I can see an argument that agent_index is a better name than current_index, but probably best to just keep the way things were dont previously and rename your var to current_index. 

TLDR probably want to just do `current_index = self.agents.index(current_agent)` and replace `agent_index` on the next line with `current_index`.

pettingzoo/classic/chess/chess.py
```diff
@@ -272,7 +286,7 @@ def step(self, action):
         self._accumulate_rewards()
 
         # Update board after applying action
-        next_board = chess_utils.get_observation(self.board, current_agent)
+        next_board = chess_utils.get_observation(self.board, 0)
```
elliottower, line 289: Maybe add a comment right above here noting that it's always from white's perspective (you note it in the docstring but that's long and people reading this code might be confused why it's 0). Could also explicitly do the arguments so people know what 0 means, i.e., `get_observation(orig_board=self.board, player=0`

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -8,8 +8,8 @@ def boards_to_ndarray(boards):
     bits = np.unpackbits(arr8)
     floats = bits.astype(bool)
     boardstack = floats.reshape([len(boards), 8, 8])
-    boardimage = np.transpose(boardstack, [1, 2, 0])
-    return boardimage
+    boardstack = np.flip(np.transpose(boardstack, [1, 2, 0]), axis=[0, 1])
```
elliottower, line 11: Can you leave a comment above this line or in the docstring or something explaining why you're calling np.flip? I'm assuming it's the issue you noted in discord with the left side of the board being displayed as the right side for some reason, but future devs would likely be confused by this and want some explanation. Ideally we'd get to the bottom of it and see why it's doing that mirroring in the first place but I'm fine just fixing it for now and noting that there's some other underlying problem that leads to things being mirrored

elliottower: Also not 100% sure if boardimage is the right variable name but I say probably just keep that name for consistency's sake, to show which part is stacking and which is turning it into the actual image

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -194,7 +194,7 @@ def legal_moves(orig_board):
     return legal_moves
 
 
-def get_observation(orig_board, player):
+def get_observation(orig_board, player: int):
```
elliottower, line 197: If you don't mind it would actually be nice to have type hints for all of this file

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -207,6 +207,9 @@ def get_observation(orig_board, player):
     else:
         board = board
 
+    # print("Chess black is: ", chess.BLACK) # 0
```
elliottower, line 210: Accidental printlines left in (easy to forget)

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -281,8 +284,8 @@ def get_observation(orig_board, player):
 
         """
     base = BASE
-    OURS = 0
-    THEIRS = 1
+    OURS = 1
```
elliottower, line 287: Maybe make a comment on this as well? I'm not 100% sure what this change does, and to a dev unfamiliar looking at this in the future they might be like hmm this PR seems to switch and flip a lot of different things, but I have no idea which line corresponds to which change exactly

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -321,17 +324,18 @@ def get_observation(orig_board, player):
       }
     """
     # from 0-63
-    square = board.ep_square
+    # Adjust the row number for the white pawn to the 8th if the en passant flag is set, and vice versa for black pawns.
+    square = board.ep_square  # (int) where the en passant happened
```
elliottower, line 328: Minor but probably most official sounding to say `# square where en passant happened (int)`

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -321,17 +324,18 @@ def get_observation(orig_board, player):
...
     if square:
-        ours = square > 32
+        ours = square < 32
```
elliottower, line 330: I'm assuming this is because ours was flipped to be the opposite in the line above, but could leave an inline comment saying `less than 32 is a white square` (or whichever color it is, I'm a little mixed up with all these details)

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -321,17 +324,18 @@ def get_observation(orig_board, player):
...
-        dest_col_add = 8 * 7 if ours else 0
+        dest_col_add = 0 if ours else 8 * 7
         dest_square = dest_col_add + row
         if ours:
-            result[base + 0].remove(square - 8)
+            result[base + 0].remove(square + 8)
             result[base + 0].add(dest_square)
         else:
-            result[base + 6].remove(square + 8)
+            result[base + 6].remove(square - 8)
```
elliottower, line 338: Not strictly necessary but personally I'm not 100% sure what this part of the code does so if you wanted to add a little in line comment for the if and the else part saying what the two scenarios are, that might be helpful

pettingzoo/classic/chess/chess_utils.py
jacob975: I write down the reason why I change the flag `OURS` and `THEIRS` in `chess_utils.py` as well as this pull request.

pettingzoo/classic/chess/chess.py
```diff
@@ -41,10 +41,13 @@
...
 * Channel 19: represents whether a position has been seen before (whether a position is a 2-fold repetition)
+* Channel 20 - 111 represents the previous 7 boards, with each board represented by 13 channels. The latest board occupies the first 13 channels, followed by the second latest board, and so on. These 13 channels correspond to channels 7 - 20.
 
-Like AlphaZero, the board is always oriented towards the current agent (the currant agent's king starts on the 1st row). In other words, the two players are looking at mirror images of the board, not the same board.
+Similar to AlphaZero, our observation space follows a stacking approach, where it accumulates the previous 8 board observations.
 
-Unlike AlphaZero, the observation space does not stack the observations previous moves by default. This can be accomplished using the `frame_stacking` argument of our wrapper.
+Unlike AlphaZero, where the board orientation may vary, in our system, the `env.board_history` always maintains the orientation towards the white agent, with the white agent's king consistently positioned on the 1st row. In simpler terms, both players are observing the same board layout.
+
```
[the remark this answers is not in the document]
jacob975: Thank you for reviewing. It is my pleasure.

pettingzoo/classic/chess/chess_utils.py
jacob975: I add an example and a few notes to explain how this section works if en passant happened.

pettingzoo/classic/chess/chess.py
jacob975: I replace the `agent_index` with `current_index` and also check the naming consistency.

pettingzoo/classic/chess/chess.py
jacob975: Thank you for your suggestion, I change to use `player=0` and add a comment above noting that it is always from the white's perspective.

pettingzoo/classic/chess/chess_utils.py
jacob975: I add a comment of why I use `np.flip`. I also add more explanation to this update in the pull request. 
To summarize, in the last version, the function `boards_to_ndarray` outputs an observation that is rotated by 180 degrees compared to `env.render()`.

jacob975: The naming is my typo. Now it is back to `boardimage`.

pettingzoo/classic/chess/chess_utils.py
jacob975: I add a few type hints if it is obvious, e.g. `chess.Board`.

pettingzoo/classic/chess/chess_utils.py
jacob975: Sure, I update this comment.

pettingzoo/classic/chess/chess_utils.py
jacob975: I swap the value of `OURS` and `THEIRS` because the module `chess` defined `chess. WHITE=1` and `chess.BLACK=0`. I guess it came from a typo because in `pettingzoo.chess_v6`, `player_0` play the white.

pettingzoo/classic/chess/chess_utils.py
elliottower: What underlying condition caused the previous version to flip it though? I would think we could figure out the math somehow and resolve it. But maybe it’s due to Pygame’s rendering flipping things in observe? Although the underlying chess board I imagine you can do some sort of function to get it as an array. Maybe we can look into using that or at least check to see how it works.

pettingzoo/classic/chess/chess_utils.py
elliottower: Ok this is good

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -324,18 +323,31 @@ def get_observation(orig_board, player: int):
       }
     """
     # from 0-63
-    # Adjust the row number for the white pawn to the 8th if the en passant flag is set, and vice versa for black pawns.
-    square = board.ep_square  # (int) where the en passant happened
+    # Adjust the row number for the white pawn to the 1st if the en passant flag is set, and vice versa for black pawns.
+    # For example
```
elliottower, line 327: Super minor but I’d say For example:
And do the rest on the line

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -324,18 +323,31 @@ def get_observation(orig_board, player: int):
...
+    # If the white play an en passant move, the opponent can play a special move called en passant capture.
```
elliottower, line 328: And this doesn’t actually explain it, the idea is if a pawn is next to another piece and moves diagonally to still be next to it then it removes it, right? I’d just Google around and find a nice simple explanation

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -324,18 +323,31 @@ def get_observation(orig_board, player: int):
...
+    # To show this, we denote the pawn at (row, col) = (1, `dest_square`) instead of (5, `dest_square`).
```
elliottower, line 329: I don’t get why we denote it differently with 1, dest square. Is that just because it’s impossible for a pawn to go backwards so this is just the only encoding for a pawn being in en passant position? Is this something done by AlphaZero or others? Seems a bit odd.

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -324,18 +323,31 @@ def get_observation(orig_board, player: int):
...
+    square = board.ep_square  # square where the en passant happened (int)
```
elliottower, line 330: Could be my phone glitching but I think this comment isn’t tab aligned properly

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -324,18 +323,31 @@ def get_observation(orig_board, player: int):
...
+        ours = (
+            square < 32
+        )  # Less than 32 is a white square, otherwise it's a black square
         row = square % 8
         dest_col_add = 0 if ours else 8 * 7
         dest_square = dest_col_add + row
         if ours:
-            result[base + 0].remove(square + 8)
-            result[base + 0].add(dest_square)
+            result[base + 0].remove(
```
elliottower: Same with the comments in the line being made ugly, just move them to above

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -324,18 +323,31 @@ def get_observation(orig_board, player: int):
...
+                dest_square
+            )  # Set the `dest_square` position in channel `base` to 1
         else:
-            result[base + 6].remove(square - 8)
-            result[base + 6].add(dest_square)
+            result[base + 6].remove(
+                square - 8
+            )  # Set the `square + 8` position in channel `base` to 0
+            result[base + 6].add(
+                dest_square
```
elliottower: And some down here too, looks good content wise though

pettingzoo/classic/chess/chess_utils.py
jacob975: I add a comment of `Less than 32 is a white square, otherwise it's a black square` above it.

pettingzoo/classic/chess/chess_utils.py
jacob975: Yes, you are right, a white pawn never walks to the 1st row. It is just a notation of the chance to make a en passant capture. It is written in the document `pettingzoo/classic/chess/chess.py`#41. To my knowledge I have no idea whether it is conventional, but I just follow the definition already in pettinzoo.chess_v6.

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -324,18 +323,31 @@ def get_observation(orig_board, player: int):
...
-    square = board.ep_square  # (int) where the en passant happened
+    # Adjust the row number for the white pawn to the 1st if the en passant flag is set, and vice versa for black pawns.
+    # For example
+    # If the white play an en passant move, the opponent can play a special move called en passant capture.
+    # To show this, we denote the pawn at (row, col) = (1, `dest_square`) instead of (5, `dest_square`).
+    square = board.ep_square  # square where the en passant happened (int)
     if square:
-        ours = square < 32
+        ours = (
+            square < 32
```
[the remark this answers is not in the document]
jacob975: Sorry for this. it is my typo from copilot or some toolkits I used on VScode.

pettingzoo/classic/chess/chess_utils.py
jacob975: My typo again. Now I move the comment to above.

pettingzoo/classic/chess/chess_utils.py
jacob975: OK, I remove these down.

pettingzoo/classic/chess/chess_utils.py
elliottower: Nvm it’s all one line and doesn’t get turned into two lines so looks good

pettingzoo/classic/chess/chess_utils.py
elliottower: This is to say, a single line saying “# For example” looks bad

pettingzoo/classic/chess/chess_utils.py
jacob975: The prevision 180 rotation is contributed by two factors. The line 8 `bits = np.unpackbits(arr8)` gives a bit map, by default, increments from RHS to LHS. Another factor is that the env.render() shows the first line at the bottom but `numpy` shows the first line on the top.

I am going to make comments and optimization about these in next revision.

pettingzoo/classic/chess/chess_utils.py
jacob975: I review the paper of AlphaZero and the notation system of chess called Forsyth-Edwards Notation (FEN). AlphaZero mentioned some of special movement in chess like queen castling, but nothing about en passant capture. They did not explain but I guess the reason is a neural network can retrive information of en passant flag from their board history. On the other hand, FEN is a popular notation used on chess. I also see a function for it in the module `chess`. FEN use a individual string to denote whether a en passant flag is set or not, instead of putting into the 1st line of the board.

Back to our observation space, so far I revision the en passant section to keep its functionality. However, it is not conventional to my knowledge. In addition, we have provided a borad history where a model can retrive the information of en passant flag. Therefore, I would like to comment out this section of code.

elliottower: Could you check other RL chess implementations such as openspiel or other papers? Just want to make sure it won’t mess things up for people. I think I linked it previously but RLlib has a LeelaChessZero model, maybe we can check if that would work with this change? I can try to get in contact with the person who did the pr for that if need be

jacob975: OK, I will check and inform you if I need help.

benblack769: I copied this behavior from LC0, but I took a 2nd look at the AlphaZero paper and it definitely doesn't mention this. Presumably alphazero is relying on the board history to provide the necessary information for en-passant, and the Lc0 developers decided to add this information in explicitly.

elliottower: If the current code behavior is also done by leela chess zero that seems reasonable to do imo. Having some way of accessing the information is better than none. Do you have thoughts on the FEN system mentioned above @benblack769?

benblack769: Yeah, FEN's representation is nice because it is concise and unambiguous, but neural networks aren't necessarily the best at learning concise information, it can be better to spread them out.

For a code reference, here is how Lc0 decodes FEN into its internal representation of en-passant https://github.com/LeelaChessZero/lc0/blob/master/src/chess/board.cc#L1114

pettingzoo/classic/chess/chess_utils.py
```diff
@@ -324,18 +324,18 @@ def get_observation(orig_board: chess.Board, player: int):
       }
     """
 
-    # from 0-63
     """
     The LeelaChessZero-style en passant flag.
-    Adjust the row number for the white pawn to the 1st if the en passant flag is set, and vice versa for black pawns.
+    In FEN, the en passant flag is represented by the square that can be a possible target of an en passant, e.g. the `e3` in `4k3/8/8/8/4Pp2/8/8/4K3 b - e3 99 50`.
```
elliottower, line 329: This is great, thanks for the extra clarification

elliottower (approved):

PR #1036, opened 2023-07-21
Env creation tutorial 2 : grid size inconsistent

BertrandDecoster:
There were two mistakes
 - The grid is 7x7 but the guard starts at position (7,7) which is out of bounds
 - The type of the grid used in `render() `was integers and it conflicts with the characters we assign in the function

# Description

## Type of change

# Checklist:

Most of those at not relevant, it's a doc change

### DIFF
--- a/tutorials/EnvironmentCreation/tutorial2_adding_game_logic.py
+++ b/tutorials/EnvironmentCreation/tutorial2_adding_game_logic.py
@@ -104,7 +104,7 @@ class CustomEnvironment(ParallelEnv):
         return observations, rewards, terminations, truncations, infos
 
     def render(self):
-        grid = np.zeros((7, 7))
+        grid = np.full((8,8), ' ', dtype='U1')
         grid[self.prisoner_y, self.prisoner_x] = "P"
         grid[self.guard_y, self.guard_x] = "G"
         grid[self.escape_y, self.escape_x] = "E"
--- a/tutorials/EnvironmentCreation/tutorial2_adding_game_logic.py
+++ b/tutorials/EnvironmentCreation/tutorial2_adding_game_logic.py
@@ -104,7 +104,7 @@ class CustomEnvironment(ParallelEnv):
         return observations, rewards, terminations, truncations, infos
 
     def render(self):
-        grid = np.full((7,7), ' ', dtype='U1')
+        grid = np.full((7,7), ' ')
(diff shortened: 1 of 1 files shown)


BertrandDecoster:
For info, I want to have this PR validated before doing a C/C for tutorial 3 of env creation

tutorials/EnvironmentCreation/tutorial2_adding_game_logic.py
```diff
@@ -104,7 +104,7 @@ def step(self, actions):
         return observations, rewards, terminations, truncations, infos
 
     def render(self):
-        grid = np.zeros((7, 7))
+        grid = np.full((7,7), ' ', dtype='U1')
```
elliottower, line 107: What does the second argument to np.full do here? And I googled around to see what exactly dtype U1 is but I'm confused about how this site says https://note.nkmk.me/en/python-numpy-dtype-astype/ the dtype should be <U1?

elliottower: The render space doesn't have to be the same as the observation space to be fair, but I can see how if we're overriding an array of integers with characters as the input that's a little sloppy. Granted, numpy lets you do that so it's like, ultimately it doesn't matter and is arguably cleaner code, but not as strictly correct. Could do it as a list of lists of size 49 or some other data type as well, but this is probably cleaner

BertrandDecoster: Are you sure it doesn't matter? It doesn't seem to want to write down a char in a array of floats

Screenshot 2023-07-22 at 09 12 35

BertrandDecoster: And 'U1' is a native type in numpy that supports char
See what happens without it : 
Screenshot 2023-07-22 at 09 14 04

BertrandDecoster: But you're right that numpy does type inference and selects an appropriate type, so there's no need to manually specify that it's Unicode, I can remove `U1`

elliottower: Ah I see, thanks for testing it out. Looks good just doing full with a space character

elliottower:
If you can fix pre commit I’d be happy to merge

PR #1037, opened 2023-07-21
Env creation tutorial 2 fix observation space

BertrandDecoster:
# Description

The observation space was too small. For example, for the observation returned after a `reset()`, the guard is at the extreme corner, which is indexed 48. However, the current observation space can only go up to 47

Related question : should the tests like `parallel_api_test()` check that the observations are consistent with the observation spaces? This could have been easily caught (cf tutorial 4)

## Type of change

# Checklist:

It's just a doc change

### DIFF
--- a/tutorials/EnvironmentCreation/tutorial2_adding_game_logic.py
+++ b/tutorials/EnvironmentCreation/tutorial2_adding_game_logic.py
@@ -112,7 +112,7 @@ class CustomEnvironment(ParallelEnv):
 
     @functools.lru_cache(maxsize=None)
     def observation_space(self, agent):
-        return MultiDiscrete([7 * 7 - 1] * 3)
+        return MultiDiscrete([7 * 7] * 3)
 
     @functools.lru_cache(maxsize=None)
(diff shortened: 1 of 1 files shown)


BertrandDecoster:
Same thing, once it's validated, I'll C/C the changes to tutorial 3

elliottower:
This is actually a good point, if you are interested in updating the API tests to actually check that the observations are in the correct spaces that would be great. It would be simpler to include all of these changes in a single PR but I like doing web commits which only allow you to change a single file so I get it (you can also just click on the patch in your browser and then go into the file tree and change other files that way, so multiple changes are on one patch)

elliottower:
I don't want to approve this until either I or the CI manually go through and check that the environment outputs the correct observation shape (and I'm pretty busy so if you'd be able to create that it would be a huge help, also for future PRs as you mention)

elliottower:
Going to merge this but if you're interested in adding a test to check that the observation and observation space match up (in a separate PR) that would be great

Issue #1039, opened 2023-07-21
[Bug Report] API test test_observation warns for every new game


BertrandDecoster:
### Describe the bug

The code introduced in 1.22.4 for `test_observation` does not seem to work as intended

```
def test_observation(observation, observation_0, env_name=None):
    if not isinstance(observation, np.ndarray) or (
        env_name is not None and env_name not in env_obs_dicts
    ):
        warnings.warn("Observation is not NumPy array")
        return
    ...
```

If your game has an output of type ndarray, but does not belong to a very short list of 10 games in `env_obs_dicts`, there is a message raised `"Observation is not NumPy array"` which is factually false.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/PettingZoo/issues) in the repo

elliottower:
Thanks for catching this, would you be interested in making a PR to fix this? The or should just be changed to an an, I believe

elliottower:
@BertrandDecoster just pinging you again to see if you wanted to make a PR for it, otherwise someone else can do it

PR #1066, opened 2023-08-17
Fix API test warning for obs not being numpy array for every custom env

elliottower:
Fixes https://github.com/Farama-Foundation/PettingZoo/issues/1039

# Description

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/pettingzoo/test/api_test.py
+++ b/pettingzoo/test/api_test.py
@@ -125,7 +125,7 @@ env_neg_inf_obs = [
 
 
 def test_observation(observation, observation_0, env_name=None):
-    if not isinstance(observation, np.ndarray) or (
+    if not isinstance(observation, np.ndarray) and (
         env_name is not None and env_name not in env_obs_dicts
     ):
         warnings.warn("Observation is not NumPy array")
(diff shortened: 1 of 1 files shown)


elliottower:
Tested before and after this change and it indeed prevents the numpy array warning from occurring. 

Example script:

```
from pettingzoo.test.example_envs import generated_agents_env_v0
from pettingzoo.test import api_test
env = generated_agents_env_v0.env()
api_test(env)
```

There is a slew of other errors but I get `/Users/elliottower/Documents/GitHub/PettingZoo/pettingzoo/test/api_test.py:131: UserWarning: Observation is not NumPy array
  warnings.warn("Observation is not NumPy array")` before the change and do not get it after the change

elliottower:
Not sure why this is failing the python tests, will have to look later

elliottower:
Looks like the previous API test was having every instance of dict observation space (e.g., action mask and observation in pz classic envs) not be tested in the observation_test, doing my change made it so the function did not return, and it treated the dict as if it was a numpy array. 

I made it so that it now re-runs the observation test with only the observation, so we can check that the actual observations inside of action masked envs are correct. I also added a new test that gets called in that case which tests the action mask (all binary, no more than 1 dimensional, etc).

Issue #1070, opened 2023-08-19
[Proposal] Permit AECEnv and ParallelEnv to have AgentIDs other than `str`


pimpale:
### Proposal

Currently, the only permitted AgentID type is the `str` type. The method types [here](https://github.com/Farama-Foundation/PettingZoo/blob/master/pettingzoo/utils/env.py) use the `AgentID` type in some places, and `str` in others. 

I propose to make `AgentID` a type variable like `ObsType` and `ActionType`, so that users can use other agent ids than strings.

### Motivation

I'd like to use numbers to key my agents, since they don't have meaningful names. The documentation suggests doing this [here](https://pettingzoo.farama.org/api/aec/#pettingzoo.utils.env.AECEnv.agents) as well. However, it feels a bit unwieldy to constantly convert strings to integers and vice versa.

Additionally, the current state of mixed `AgentID` and `str` types dictionary is inconsistent, and should be fixed to use `AgentID` everywhere.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/PettingZoo/issues) in the repo

elliottower:
This seems like a pretty reasonable thing but it would take a considerable amount of work in ensuring that the tests all work if the agent id isn’t a string.

If you or another contributor would like to submit a PR and include tests to check that it works then I’d be happy to accept it, but I won’t have the time personally.

pimpale:
Thanks for responding! I've created an initial PR for the change, and I'll add some tests soon.

PR #1071, opened 2023-08-20
Permit AgentIDs other than `str`

pimpale:
# Description

Adds the feature described in https://github.com/Farama-Foundation/PettingZoo/issues/1070, by making ParallelEnv and AECEnv generic over AgentID.

## Type of change

# Checklist:

### DIFF
--- a/pettingzoo/test/example_envs/generated_agents_parallel_cust_agentid_v0.py
+++ b/pettingzoo/test/example_envs/generated_agents_parallel_cust_agentid_v0.py
@@ -1,4 +1,4 @@
-from typing import Union
+from typing import Tuple, Union
 
 import gymnasium
 import numpy as np
@@ -19,11 +19,11 @@ def raw_env(**kwargs):
     return conversions.parallel_to_aec(parallel_env(**kwargs))
 
 
-def get_type(agent: tuple[str, int]):
+def get_type(agent: Tuple[str, int]):
     return agent[0]
 
 
-class parallel_env(ParallelEnv[tuple[str, int], np.ndarray, Union[int, None]]):
+class parallel_env(ParallelEnv[Tuple[str, int], np.ndarray, Union[int, None]]):
     metadata = {"render_modes": ["human"], "name": "generated_agents_parallel_v0"}
 
     def __init__(self, max_cycles=100, render_mode=None):
--- a/pettingzoo/test/example_envs/generated_agents_parallel_cust_agentid_v0.py
+++ b/pettingzoo/test/example_envs/generated_agents_parallel_cust_agentid_v0.py
@@ -1,3 +1,5 @@
+from typing import Union
+
 import gymnasium
 import numpy as np
 from gymnasium.utils import seeding
@@ -21,7 +23,7 @@ def get_type(agent: tuple[str, int]):
     return agent[0]
 
 
-class parallel_env(ParallelEnv[tuple[str, int], np.ndarray, int | None]):
+class parallel_env(ParallelEnv[tuple[str, int], np.ndarray, Union[int, None]]):
     metadata = {"render_modes": ["human"], "name": "generated_agents_parallel_v0"}
 
     def __init__(self, max_cycles=100, render_mode=None):
--- a/pettingzoo/test/example_envs/generated_agents_parallel_cust_agentid_v0.py
+++ b/pettingzoo/test/example_envs/generated_agents_parallel_cust_agentid_v0.py
@@ -20,6 +20,7 @@ def raw_env(**kwargs):
 def get_type(agent: tuple[str, int]):
     return agent[0]
 
+
 class parallel_env(ParallelEnv[tuple[str, int], np.ndarray, int | None]):
     metadata = {"render_modes": ["human"], "name": "generated_agents_parallel_v0"}
 
@@ -120,7 +121,9 @@ class parallel_env(ParallelEnv[tuple[str, int], np.ndarray, int | None]):
         if not truncated:
             for i in range(6):
                 if self.np_random.random() < 0.1 and len(self.agents) >= 10:
-                    all_terminations[self.agents[self.np_random.choice(len(self.agents))]] = True
+                    all_terminations[
+                        self.agents[self.np_random.choice(len(self.agents))]
+                    ] = True
 
             for i in range(3):
                 if self.np_random.random() < 0.1:
--- a/pettingzoo/test/example_envs/generated_agents_parallel_cust_agentid_v0.py
+++ b/pettingzoo/test/example_envs/generated_agents_parallel_cust_agentid_v0.py
@@ -1,0 +1,159 @@
+import gymnasium
+import numpy as np
+from gymnasium.utils import seeding
+
+from pettingzoo import ParallelEnv
+from pettingzoo.utils import conversions, wrappers
+
+
+def env(**kwargs):
+    env = raw_env(**kwargs)
+    env = wrappers.AssertOutOfBoundsWrapper(env)
+    env = wrappers.OrderEnforcingWrapper(env)
+    return env
+
+
+def raw_env(**kwargs):
+    return conversions.parallel_to_aec(parallel_env(**kwargs))
+
+
+def get_type(agent: tuple[str, int]):
+    return agent[0]
+
+class parallel_env(ParallelEnv[tuple[str, int], np.ndarray, int | None]):
+    metadata = {"render_modes": ["human"], "name": "generated_agents_parallel_v0"}
+
+    def __init__(self, max_cycles=100, render_mode=None):
+        super().__init__()
+        self._obs_spaces = {}
+        self._act_spaces = {}
+
+        # dummy state space, not actually used
+        self.state_space = gymnasium.spaces.MultiDiscrete([10, 10])
+        self._state = self.state_space.sample()
+
+        self.types = []
+        self._agent_counters = {}
+        self.max_cycles = max_cycles
+        self.rng_seed = None
+        self._seed()
+        self.render_mode = render_mode
+        for i in range(3):
+            self.add_type()
+
+    def observation_space(self, agent):
+        return self._obs_spaces[get_type(agent)]
+
+    def action_space(self, agent):
+        return self._act_spaces[get_type(agent)]
+
+    def state(self) -> np.ndarray:
+        return self._state
+
+    def observe(self, agent):
+        return self.observation_space(agent).sample()
+
+    def add_type(self) -> str:
+        type_id = len(self.types)
+        num_actions = self.np_random.integers(3, 10)
+        obs_size = self.np_random.integers(10, 50)
+        obs_space = gymnasium.spaces.Box(low=0, high=1, shape=(obs_size,))
+        act_space = gymnasium.spaces.Discrete(num_actions)
+        obs_space.seed(self.rng_seed)
+        act_space.seed(self.rng_seed)
+        new_type = f"type{type_id}"
+        self.types.append(new_type)
+        self._obs_spaces[new_type] = obs_space
+        self._act_spaces[new_type] = act_space
+        self._agent_counters[new_type] = 0
+        return new_type
+
+    def add_agent(self, type: str):
+        agent_id = self._agent_counters[type]
+        self._agent_counters[type] += 1
+        agent_name = (type, agent_id)
+        self.agents.append(agent_name)
+        return agent_name
+
+    def reset(self, seed=None, options=None):
+        self.rng_seed = seed
+
+        if seed is not None:
+            self._seed(seed=seed)
+        self.num_steps = 0
+
+        # Reset spaces and types
+        self._obs_spaces = {}
+        self._act_spaces = {}
+        self.state_space = gymnasium.spaces.MultiDiscrete([10, 10])
+        self._state = self.state_space.sample()
+
+        self.types = []
+        self._agent_counters = {}
+        for i in range(3):
+            self.add_type()
+
+        # Add agents
+        self.agents = []
+        for i in range(5):
+            self.add_agent(self.np_random.choice(self.types))
+
+        # seed observation and action spaces
+        for i, agent in enumerate(self.agents):
+            self.observation_space(agent).seed(seed)
+        for i, agent in enumerate(self.agents):
+            self.action_space(agent).seed(seed)
+
+        return {agent: self.observe(agent) for agent in self.agents}, {
+            agent: {} for agent in self.agents
+        }
+
+    def _seed(self, seed=None):
+        self.np_random, _ = seeding.np_random(seed)
+
+    def step(self, actions):
+        truncated = self.num_steps >= self.max_cycles
+        for agent in self.agents:
+            assert agent in actions
+        all_truncations = {agent: truncated for agent in self.agents}
+        all_terminations = {agent: False for agent in self.agents}
+        if not truncated:
+            for i in range(6):
+                if self.np_random.random() < 0.1 and len(self.agents) >= 10:
+                    all_terminations[self.agents[self.np_random.choice(len(self.agents))]] = True
+
+            for i in range(3):
+                if self.np_random.random() < 0.1:
+                    if self.np_random.random() < 0.1:
+                        type = self.add_type()
+                    else:
+                        type = self.np_random.choice(self.types)
+
+                    new_agent = self.add_agent(type)
+                    all_terminations[new_agent] = False
+                    all_truncations[new_agent] = False
+
+        all_infos = {agent: {} for agent in self.agents}
+        all_rewards = {agent: 0 for agent in self.agents}
+        all_rewards[self.agents[self.np_random.choice(len(self.agents))]] = 1
+        all_observes = {agent: self.observe(agent) for agent in self.agents}
+        self.agents = [
+            agent
+            for agent in self.agents
+            if not (all_truncations[agent] or all_terminations[agent])
+        ]
+
+        if self.render_mode == "human":
+            self.render()
+        return all_observes, all_rewards, all_terminations, all_truncations, all_infos
+
+    def render(self):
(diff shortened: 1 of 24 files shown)


pimpale:
Hi, @elliottower. I believe this PR is ready to be reviewed. For testing, I copied the environments in example_env, but made them use non-string AgentIDs, since those demonstrate termination, addition, and other behaviors. Let me know if that testing is sufficient, or if there are other tests you would like to see.

pettingzoo/utils/env.py
```diff
@@ -22,7 +18,7 @@
 """
 
 
-class AECEnv:
+class AECEnv(Generic[AgentID, ObsType, ActionType]):
```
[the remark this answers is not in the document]
elliottower: My guess is if you have to pass in ObsType or ActionType when you define an AECEnv it will make things a bit more confusing, kind of feel like it would be simplest to keep it as it was before but just replace uses of `str` in signatures to `AgentID` (I've already tried to do this a few places because it's cleaner)

elliottower:
Another potential idea although it wouldn't be as clean, would be to have a wrapper that converts into string agent IDs so that the highest level user facing environment would be interactable using integers for agent id, but then in every subsequent level beneath that it would be cast to a string so the other wrappers and functionality works normally. 

Only problem is certain things you may not want to convert to a string, but inherently string types are sort of the type which everything can be converted into so I don't see it being a huge problem

pettingzoo/utils/env.py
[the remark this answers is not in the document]
pimpale: What inheriting from Generic does is it makes the class have the ability to accept type parameters when you define a child class. When instantiating the class, the type parameters are inferred.

However, providing type parameters is purely optional. You don't have to pass in an AgentID, ObsType or ActionType when defining an AECEnv unless you want to. If you don't provide these bits of info, then they are left as unknown. This means that the change is fully backward-compatible. 

I.E, if the user does not touch their code after this change, it will still work, and it will not introduce any new type errors.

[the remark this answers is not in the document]
pimpale: Also, Gymnasium also uses Generics when defining their environment: https://github.com/Farama-Foundation/Gymnasium/blob/main/gymnasium/core.py#L21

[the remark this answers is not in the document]
elliottower: Oh I see, well in that case this looks good then, thanks for the heads up about Gymnasium doing it as well

pettingzoo/test/example_envs/generated_agents_parallel_cust_agentid_v0.py
```diff
@@ -0,0 +1,164 @@
...
+
+def raw_env(**kwargs):
+    return conversions.parallel_to_aec(parallel_env(**kwargs))
+
+
+def get_type(agent: tuple[str, int]):
+    return agent[0]
+
+
+class parallel_env(ParallelEnv[tuple[str, int], np.ndarray, Union[int, None]]):
```
elliottower, line 26: My only thought is that this is a bit confusing in that when you look at it, it doesn't immediately become clear that the first arg is the AgentID, and that the others are observation, etc.

pettingzoo/test/example_envs/generated_agents_parallel_v0.py
```diff
@@ -30,7 +32,7 @@ def __init__(self, max_cycles=100, render_mode=None):
         self._act_spaces = {}
 
         # dummy state space, not actually used
-        self.state_space = gymnasium.spaces.MultiDiscrete((10, 10))
+        self.state_space = gymnasium.spaces.MultiDiscrete([10, 10])
```
elliottower, line 35: Was this a typo previously or something?

pettingzoo/utils/conversions.py
```diff
@@ -217,17 +223,21 @@ def close(self):
...
+    ):
         self.env = parallel_env
 
         self.metadata = {**parallel_env.metadata}
         self.metadata["is_parallelizable"] = True
 
         try:
-            self.render_mode = self.env.render_mode
+            self.render_mode = (
+                self.env.render_mode  # pyright: ignore[reportGeneralTypeIssues]
```
elliottower, line 239: I don't see why your code should make this specific line need a different pyright ignore, I guess it was just throwing that error now for some reason? Kind of ugly but doesn't matter much if it's not possible to remove

pettingzoo/utils/conversions.py
```diff
@@ -311,7 +325,7 @@ def add_new_agent(self, new_agent):
         self.rewards[new_agent] = 0
         self._cumulative_rewards[new_agent] = 0
 
-    def step(self, action: ActionType):
+    def step(self, action: Optional[ActionType]):
```
elliottower, line 328: Why is the action type now optional?

pettingzoo/utils/conversions.py
```diff
@@ -335,7 +349,7 @@ def step(self, action: ActionType):
 
             self.agents = self.env.agents + [
                 agent
-                for agent in sorted(self._observations.keys())
+                for agent in sorted(self._observations.keys(), key=lambda x: hash(x))
```
elliottower, line 352: I'm assuming this is to be able to handle non-strings? Does hashing produce any undesirable side effects though? I would imagine you would probably want it to sort in a specific order otherwise you wouldn't call sorted (I'm not familiar with the previous code here)

pettingzoo/utils/conversions.py
```diff
@@ -433,7 +451,7 @@ def reset(self, seed=None, options=None):
             if not (self.aec_env.terminations[agent] or self.aec_env.truncations[agent])
         }
 
-        infos = dict(**self.aec_env.infos)
+        infos = {**self.aec_env.infos}
```
elliottower, line 454: It may be more readable to do dict() in this case, don't see a reason to change it to be honest

pettingzoo/utils/save_observation.py
```diff
@@ -1,14 +1,17 @@
...
 
 import gymnasium.spaces
 import numpy as np
 
 from pettingzoo.utils.env import AECEnv, AgentID, ParallelEnv
 
 
-def _check_observation_saveable(env: AECEnv | ParallelEnv, agent: AgentID) -> None:
+def _check_observation_saveable(
+    env: AECEnv[AgentID, Any, Any] | ParallelEnv[AgentID, Any, Any], agent: AgentID
```
elliottower, line 13: Should the Any, Any not be ObsType and ActionType?

test/variable_env_test.py
```diff
@@ -11,15 +13,21 @@
 
 
 def test_generated_agents_aec():
-    # check that that AEC env passes API test and produces deterministic behavior
```
elliottower, line 14: Good catch finding this typo

elliottower (approved):
Looks pretty much good but a few little questions and nitpicks, thanks so much for this, this sort of stronger typing and more consistency with Gymnasium is really important

elliottower:
Looks like it's saying `type is not subscriptable` https://github.com/Farama-Foundation/PettingZoo/actions/runs/5929053382/job/16077967033?pr=1071

pettingzoo/test/example_envs/generated_agents_parallel_v0.py
pimpale: It was giving me a type error, since MultiDiscrete "officially" only accepts `list`s and `ndarray`s.

pettingzoo/utils/conversions.py
pimpale: Actually, using `dict()` doesn't work, since it requires keys that are strings, which is not necessarily the case now. However, `{}` does work for non-string keys.

elliottower:
I believe the type errors will go away if you add this line: 
`from __future__ import annotations`

pettingzoo/utils/save_observation.py
pimpale: Yes, you could write it like that, but if you don't use a generic variable in more than one place in a function definition, then it acts just like Any, since it's not actually constraining anything.

pettingzoo/utils/conversions.py
elliottower: Oh interesting, good catch then

pettingzoo/test/example_envs/generated_agents_parallel_v0.py
elliottower: Oh I see, sounds good then

pettingzoo/utils/conversions.py
pimpale: In the original version of `conversions.py`, we had the line `ActionType = Optional[int]` at the top. 

This is because we have to be able to accept `None` as a potential action, due to this line: https://github.com/Farama-Foundation/PettingZoo/blob/master/pettingzoo/utils/conversions.py#L456

However, I wanted to make it more general, so I just started using `Optional[ActionType]`, so that we can use actions other than int.

pettingzoo/utils/save_observation.py
elliottower: Fair enough, I just figure it may be more readable for people to see that it's those 3 arguments even if it doesn't change the performance. Not like typing in Python really even does much anyways besides things like pyright

pettingzoo/utils/conversions.py
elliottower: Ahh okay that makes sense. Sounds good, good catch.

pettingzoo/utils/conversions.py
pimpale: Yes, the reason is in order to handle non-strings.

The reason we call sorted here, at least as far as I know, is so that the `agents` list has a deterministic ordering, since otherwise a dictionary makes no guarantees as to the order of its items.

You actually raise a good point here, since `hash` values are not consistent between different runs of the python interpreter. 

I should use the `str` function to convert it to a string for comparison instead here.

pimpale:
I've added annotations on all files where I saw the error occur.

pettingzoo/test/example_envs/generated_agents_parallel_cust_agentid_v0.py
pimpale: I'm not currently aware of any way to provide named generic arguments, but I agree that many generic arguments is kind of confusing. We only have 3 so it's not too bad yet.

pettingzoo/utils/conversions.py
pimpale: Render mode is not actually defined as a property or variable on either `ParallelEnv` or `AECEnv`.

pettingzoo/utils/conversions.py
pimpale: Done now.

elliottower:
> I've added annotations on all files where I saw the error occur.

Sounds good, thanks

elliottower:
It may not be that ideal to have to import annotations on any internal file which deals with calling these environments, but that's a cost I'm willing to pay. Plus I believe in 3.9 you don't need to import them anymore, so once 3.12 is out next year and 3.8 is phased out I think it will no longer be necessary.

pettingzoo/test/seed_test.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
elliottower, line 1: Okay this file doesn't use any type hinting at all so I think you shouldn't need to have annotations. I think it may be a problem that env.py didn't have annotations or some other root file didn't have annotations imported, so then when you ran that file it raised an error. You can fix that by importing annotations on all the other files which call that root file, but I think they won't be necessary if you fix the root cause. 

I could be wrong but this is my hunch.

pettingzoo/utils/conversions.py
```diff
@@ -73,29 +73,35 @@ def parallel_to_aec(par_env: ParallelEnv) -> AECEnv:
...
+    par_env: ParallelEnv[AgentID, ObsType, Optional[ActionType]]
+) -> AECEnv[AgentID, ObsType, Optional[ActionType]]:
     warnings.warn(
         "The `from_parallel` function is deprecated. Use the `parallel_to_aec` function instead."
     )
     return parallel_to_aec(par_env)
 
 
-class aec_to_parallel_wrapper(ParallelEnv):
+class aec_to_parallel_wrapper(ParallelEnv[AgentID, ObsType, ActionType]):
```
elliottower, line 104: This occurrence of it has ActionType rather than Optional[ActionType], looks like you just forgot to add it but maybe it's intentional for some reason?

pettingzoo/utils/conversions.py
elliottower: That's a little weird, but I don't think we should change it. I imagine there are some checks which say if hasattr(render_mode) or something similar. Other subsequent training libraries assume things are how they are so I don't want to change much without a really strong reason to do so.

elliottower (approved):

elliottower:
Tag me when you want me to re-run the tests or review again, this looks almost perfect but just a few little nitpicks I noted above. Thanks again for this, great work paying such close attention to detail and fixing all of the little issues.

pettingzoo/utils/conversions.py
pimpale: It's only necessary from ParallelEnv -> AECEnv, but not the other way around.

pettingzoo/test/seed_test.py
pimpale: That seems correct.

pettingzoo/utils/save_observation.py
pimpale: When I replaced them with the typevariables, I got a warning from pyright. Apparently it's not preferred to use a TypeVar only once in a generic function signature.

pimpale:
It looks like all tests have passed. From my end, I'm ready to merge.

pimpale:
@elliottower Following up on this.

pettingzoo/utils/save_observation.py
elliottower: Ok whatever pyright wants is fine then

elliottower:
@pimpale could you look into this? Looks like your removal of obs dict caused an issue with shimmy. Maybe it should be added back in a patch? https://github.com/Farama-Foundation/Shimmy/issues/106

pettingzoo/utils/env.py
```diff
@@ -177,7 +173,7 @@ def agent_iter(self, max_iter: int = 2**63) -> AECIterable:
 
     def last(
         self, observe: bool = True
-    ) -> tuple[ObsType | None, float, bool, bool, dict[AgentID, dict[str, Any]]]:
+    ) -> tuple[ObsType | None, float, bool, bool, dict[str, Any]]:
```
elliottower, line 176: Also looking at before and after this line looks incorrect, it should use AgentID, but also simpler to just keep the ObsDict type at the top of the file and define it to use AgentID, which you changed ofc. That way if agent id gets changed again this will not need to be changed as it just references it. And prevents breaking code like the shimmy bug.

pimpale: The reason I changed that line was because we return ` self.infos[agent]` for the info dict, and `self.infos` has the type `dict[AgentID, dict[str, Any]]`.

 That line was producing a type error previously.

elliottower: I'm a little confused why it needs to be dict[str] here though instead of agent ID

elliottower: I guess this has to do with the other PR up right now which talks about infos and the correct syntax for it. What I gather is that parallel envs need to have the dict be indexed by agent ID and have sub dicts with info for each agent, but I guess in AEC environments you just need any dict and it's assumed that it's the info only for that agent on that timestep. So it shouldn't be nested dict like the original code, but I think it should still be dict[AgentID, Any]

elliottower:
Is there a principled reason that ObsDict and ActionDict need to be removed though? I don't really see a reason why we shouldn't include them just in case people already use them and like the convenience or whatever

pettingzoo/utils/env.py
```diff
@@ -297,17 +293,21 @@ def reset(
...
         And returns a dictionary of observations (keyed by the agent name)
         """
         raise NotImplementedError
 
     def step(
-        self, actions: ActionDict
+        self, actions: dict[AgentID, ActionType]
     ) -> tuple[
-        ObsDict, dict[str, float], dict[str, bool], dict[str, bool], dict[str, dict]
+        dict[AgentID, ObsType],
```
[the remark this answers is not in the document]
elliottower: I guess the consequence of this is that it's now somewhat backwards incompatible, but I guess even I agreed that it seemed cleaner to be explicit that it's a dict of agentID and obstype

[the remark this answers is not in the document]
elliottower: Yeah okay looking back at that full file I think you're probably right that we should just make a shimmy update and try to be more clear, we don't have RewardDict or anything like that anyways, so it's simplest to make all of these types fully clear when you see them in the type hint and know what to expect without having to cross reference 'oh what is an ObsDict'. And I think I'm probably one of the only people who uses ObsDict anyways

elliottower:
Just looked and it seems like 21 different repositories use the ObsDict type... guess I'm not the only one. I can see an argument for keeping the ObsDict defined for backwards compatibility but internally still doing your changes which I think are more clean. (TLDR: just re-add the two lines defining ActionDict and ObsDict but we don't need to use them internally)

pettingzoo/utils/env.py
pimpale: Well in this case, (`last()`) we're only returning the info for the current agent, so I don't understand why the it should be indexed by AgentID. 

If someone wants to use `int` to index their agents, but still have an info that looks like `{ "extra_info": 1, "extra_info2": 2 }`, (per agent) then that seems fine to me.

pimpale:
Hmm. In that case I agree with you, I'll add them back.

elliottower:
Yeah but we want to minimize them when possible and make the upgrade process as easy as can be

PR #1074, opened 2023-08-23
Tests duplicated deleted from parametrized_envs

GiovanniGrotto:
# Description

I deleted all the repeted lines in ```parametrized_envs``` from ```all_parameters_combs_tests.py```, and changed lines where the kwargs where matching with default arguments with an empty dict, so now the test will run with default args also if they will be changes in __init__ func.

## Type of change

### Screenshots

This is an exmaple of what I changed, the first highlighted line now has an empty dict instead of hand_size=5 and the second and third lines have been deleted.
Immagine 2023-08-19 153353

# Checklist:


elliottower (approved):
Looks good to me, thanks for taking the time to fix this. Curious what the speedup will be in the CI. If you could also try to cross reference with the pytest runner file and see if we actually need both of these files (my guess is that was created first and this is the more comprehensive version of it, you could check git blame to figure out when each was created and stuff like that as well)

GiovanniGrotto:
> Looks good to me, thanks for taking the time to fix this. Curious what the speedup will be in the CI. If you could also try to cross reference with the pytest runner file and see if we actually need both of these files (my guess is that was created first and this is the more comprehensive version of it, you could check git blame to figure out when each was created and stuff like that as well)

Yes I'll check in the next days an try to speed up test as much as possible, I can also check Gymnasium wrapper for max_cycles

elliottower:
Looks good, only thing I'd say is try to at least do two parameter combinations per environment (I imagine the test will take about the same amount of time if you just remove the pytest runner test and do this, because most envs are already being run twice with the two files)

elliottower:
Good stuff, the tests are now 15 mins instead of 30 mins

PR #1076, opened 2023-08-25
Fix tutorial reset returns

ggsavin:
# Description

Environments that inherit from `ParallelEnv` and are converted to  `AECEnv` via `parallel_to_aec` require named agent keys in their `env.reset() `infos` returns. This is because the generated `env.last` function uses the info IDs from the `env.reset` return.

The tutorial for custom environment creation had an empty object returned for `infos`, which would pass the current `parallel_api_test` however cause failures if a user for example wanted to train their agents in Tianshou and needed to covert `ParallelEnv` as an `AECEnv` before necessary Tianshou wrappers..

This fix looks to update the documentation while also updating the `parallel_api_test` to catch this oversight in current and future projects.

## Type of change

### Screenshots
The CustomEnvironment is copied over verbatim, and we are trying to wrap it in a compatible Tianshou PettingZooEnv 

image

# Checklist:

### DIFF
--- a/pettingzoo/utils/conversions.py
+++ b/pettingzoo/utils/conversions.py
@@ -294,11 +294,6 @@ class parallel_to_aec_wrapper(AECEnv[AgentID, ObsType, Optional[ActionType]]):
 
     def reset(self, seed=None, options=None):
         self._observations, self.infos = self.env.reset(seed=seed, options=options)
-
-        # Every environment needs to return infos that contain the same keys as the observations
-        if bool(self.infos) == False:
-            self.infos = {agent: {} for agent in self._observations.keys()}
-
         self.agents = self.env.agents[:]
         self._live_agents = self.agents[:]
         self._actions: Dict[AgentID, Optional[ActionType]] = {
@@ -309,6 +304,12 @@ class parallel_to_aec_wrapper(AECEnv[AgentID, ObsType, Optional[ActionType]]):
         self.terminations = {agent: False for agent in self.agents}
         self.truncations = {agent: False for agent in self.agents}
         self.rewards = {agent: 0 for agent in self.agents}
+
+        # Every environment needs to return infos that contain the same keys as the observations
+        if bool(self.infos) == False:
+            warnings.warn("The `infos` dictionary returned by `env.reset` was empty. Agent IDs as keys will be used")
+            self.infos = {agent: {} for agent in self.agents}
+
         self._cumulative_rewards = {agent: 0 for agent in self.agents}
         self.new_agents = []
         self.new_values = {}
--- a/pettingzoo/utils/conversions.py
+++ b/pettingzoo/utils/conversions.py
@@ -294,6 +294,11 @@ class parallel_to_aec_wrapper(AECEnv[AgentID, ObsType, Optional[ActionType]]):
 
     def reset(self, seed=None, options=None):
         self._observations, self.infos = self.env.reset(seed=seed, options=options)
+
+        # Every environment needs to return infos that contain the same keys as the observations
+        if bool(self.infos) == False:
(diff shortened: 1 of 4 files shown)


ggsavin:
@elliottower 
I've added a commit that will populate an empty info dict with the keys from self.agents

pettingzoo/utils/conversions.py
```diff
@@ -294,6 +294,11 @@ def action_space(self, agent):
 
     def reset(self, seed=None, options=None):
         self._observations, self.infos = self.env.reset(seed=seed, options=options)
+
+        # Every environment needs to return infos that contain the same keys as the observations
+        if bool(self.infos) == False:
```
elliottower: I think you can just say if not self.infos or if self.infos = None, unless casting it to a bool is checking something else? But I think it's good to add a blank info in this case like you do, but you should also add a warning that infos is not defined, or something similar.

Also should check something like `isinstance(self.infos, dict) and [ agent not in self.infos.keys() for agent in self.agents ]`, in which case you can do a separate warning that infos is a dict but it does not have the required keys. Then say that we are correcting it basically (my philosophy here is it's not that clear how you're supposed to do it, and if a user is using someone else's environment which isn't properly configured, they should still be able to convert it even if they don't have write access to change the original environment)

ggsavin: Updated this to also have a key check, between agent ids and the keys in self.info. There is a warning here as well that the value will be overwritten. I think just an empty dict here is safer than any type of logic to re-use correct key-value pairs and drop incorrect ones. The warning will come up when the conversion is ran, and whoever is converting the environment will be able to address it

pettingzoo/utils/conversions.py
```diff
@@ -305,11 +305,14 @@ def reset(self, seed=None, options=None):
         self.truncations = {agent: False for agent in self.agents}
         self.rewards = {agent: 0 for agent in self.agents}
 
-        # Every environment needs to return infos that contain the same keys as the observations
-        if bool(self.infos) == False:
+        # Every environment needs to return infos that contain self.agents as their keys
+        if not self.infos:
             warnings.warn("The `infos` dictionary returned by `env.reset` was empty. Agent IDs as keys will be used")
```
elliottower, line 310: I would say "Adding empty infos dictionary"

pettingzoo/utils/conversions.py
```diff
@@ -305,11 +305,14 @@ def reset(self, seed=None, options=None):
...
             self.infos = {agent: {} for agent in self.agents}
-
+        elif set(self.infos.keys()) != set(self.agents):
+            warnings.warn("The `infos` dictionary keys returned by `env.reset` is not equal to the agent IDs listed in `self.agents`. Overwriting with just agent IDs as keys")
```
elliottower, line 313: I think the warning should be more like the infos returned by env.reset is not valid: must contain keys for each agent defined in `self.agents`: {self.agents}. Duplicating current info data to each agent... {self.infos}"

pettingzoo/utils/conversions.py
elliottower: Fair enough I guess, I was just thinking if they only have an info of `turn: 0` then we could add that. Maybe if it contains none of the agent IDs then we can copy it but if it contains some of them we just say invalid info and override it to be empty?

elliottower:
Also just btw make sure to make a simple test somewhere to check that this works properly with some different failure cases

elliottower:
Looks like pre-commit failed as well

elliottower:
@ggsavin if you're not able to finish this PR I'll just do some quick fixes myself, I need to make a hotfix soon to fix issues with Shimmy

elliottower:
Closing as #1082  has been merged

PR #1077, opened 2023-08-25
Add back ObsDict and ActionDict

pimpale:
# Description

Adds back ObsDict and ActionDict to the pettingzoo.utils.env module. 

Fixes https://github.com/Farama-Foundation/Shimmy/issues/106 . 

## Type of change

# Checklist:

### DIFF
--- a/pettingzoo/test/api_test.py
+++ b/pettingzoo/test/api_test.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import re
 import warnings
 from collections import defaultdict
--- a/pettingzoo/test/api_test.py
+++ b/pettingzoo/test/api_test.py
@@ -249,7 +249,7 @@ def test_observation_action_spaces(env, agent_0):
             )
         if (not isinstance(agent, str)) and agent != "env":
             warnings.warn(
-                "Agent's are recommended to have numbered string names, like player_0"
+                "Agents are recommended to have numbered string names, like player_0"
             )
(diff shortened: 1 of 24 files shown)


pimpale:
Question: how can we mark a constant as deprecated? Stack Overflow has [some solutions](https://stackoverflow.com/questions/922550/how-to-mark-a-global-as-deprecated-in-python), but they seem a bit heavyweight.

elliottower:
> Question: how can we mark a constant as deprecated? Stack Overflow has [some solutions](https://stackoverflow.com/questions/922550/how-to-mark-a-global-as-deprecated-in-python), but they seem a bit heavyweight.

Honestly don't think we need to deprecate it, they're fine for people to use and could make things simpler for new users, we just happen to not use them internally anymore for this repo (but on shimmy I'll probably just leave it as is)

elliottower:
It's giving ` TypeError: 'type' object is not subscriptable` in CI btw, not sure exactly how to fix that or what the difference is between what was there previously

pimpale:
The issue was using `dict` instead of `typing.Dict`.

elliottower (approved):

RedTachyon:
@elliottower Shouldn't this be pushed in a hotfix release asap? At the moment this completely breaks all multiagent stuff in shimmy unless you manually downgrade PZ

elliottower:
Yeah that's the plan sorry I've just not had a chance to write up release notes

PR #1084, opened 2023-09-03
Updated CustomEnvironment tutorial comments

GiovanniGrotto:
# Description

I added some comments to the CustomEnvironment prisoner tutorial 

## Type of change

# Checklist:


elliottower (approved):
These are great thanks. One comment though is to make sure it's clear which of the bullet points are true for all environments, and which are specific for this environment. You had things about prisoner positions and such in the same list as termination truncation etc, probably best to just say something like "the reset method must do the following" and then specify this environment itself does some more additional steps. 

But this is much improved so thanks a ton for the help

PR #1091, opened 2023-09-07
Super suit migration example

GiovanniGrotto:
# Description

This is a try of migrating *clip_reward_v0* and *color_reduction_v0* from **SuperSuit** *basic_wrappers.py* to **PettingZoo**.
It is a work in progress so there is a lot of things that needs to be fixed but I would like to know if at least I'm in the good track.
In particular:
- Test folder should not exist and the test inside are just to check if everything is running and will then be improved and moved in the main test folder of PZ
- Docstring have been generated by chatGPT and will be double checked improved and fixed

I'm having some trouble with type hinting, in fact some of them could be wrong, since sometimes it is not very clear what a function should accepts, is there a way or a best practice to make this easier or at least cleaner? 
I'm open to any type of comment and suggestions about how to make this migration cleaner.

## Type of change

### Screenshots

# Checklist:

### DIFF
--- a/pettingzoo/utils/wrappers/supersuit/basic_wrappers.py
+++ b/pettingzoo/utils/wrappers/supersuit/basic_wrappers.py
@@ -1,0 +1,95 @@
+from reward_lambda import reward_lambda_v0, AecRewardLambda
+from observation_lambda import observation_lambda_v0, AecObservationLambda
+from utils.basic_transforms import color_reduction
+from typing import Literal, Any
+from types import ModuleType
+from pettingzoo import AECEnv, ParallelEnv
+from gymnasium.spaces import Space
+import numpy as np
+
+
+def basic_obs_wrapper(env: AECEnv | ParallelEnv, module: ModuleType, param: Any) -> AecObservationLambda:
+    """
+    Wrap an environment to modify its observation space and observations using a specified module and parameter.
+
+    This function takes an environment, a module, and a parameter, and creates a new environment with an observation
+    space and observations modified based on the provided module and parameter.
+
+    Parameters:
+    - env (Generic[AgentID, ObsType, ActionType]): The environment to be wrapped.
+    - module: The module responsible for modifying the observation space and observations.
+    - param: The parameter used to modify the observation space and observations.
+
+    Returns:
+    - AecObservationLambda: A wrapped environment that applies the observation space and observation modifications. #TODO fix this line
+
+    Example:
+    ```python
+    modified_env = basic_obs_wrapper(original_env, my_module, my_param)
+    ```
+    In the above example, `modified_env` is a new environment that has its observation space and observations modified
+    according to the `my_module` and `my_param`.
+    """
+
+    def change_space(space: Space):  # Box?
+        module.check_param(space, param)
+        space = module.change_obs_space(space, param)
+        return space
+
+    def change_obs(obs: np.ndarray, obs_space: Space):  # not sure about ndarray
+        return module.change_observation(obs, obs_space, param)
+
+    return observation_lambda_v0(env, change_obs, change_space)
+
+
+def color_reduction_v0(env: AECEnv | ParallelEnv, mode: Literal["full", "R", "G", "B"] = "full") -> AecObservationLambda:
+    """
+   Wrap an environment to perform color reduction on its observations.
+
+   This function takes an environment and an optional mode to specify the color reduction technique. It then creates
+   a new environment that performs color reduction on the observations based on the specified mode.
+
+   Parameters:
+   - env (Generic[AgentID, ObsType, ActionType]): The environment to be wrapped.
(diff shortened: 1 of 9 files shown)


pettingzoo/utils/wrappers/supersuit/basic_wrappers.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def basic_obs_wrapper(env: AECEnv | ParallelEnv, module: ModuleType, param: Any) -> AecObservationLambda:
+    """
+    Wrap an environment to modify its observation space and observations using a specified module and parameter.
+
+    This function takes an environment, a module, and a parameter, and creates a new environment with an observation
+    space and observations modified based on the provided module and parameter.
+
+    Parameters:
```
elliottower, line 18: If you could have the docstrings be more similar to the syntax in `conversions.py` and other wrappers in PettingZoo that would be good. Also I don't know the wrapper super well but I think `my_module` and `my_param` may not be the correct names. In general using GPT to write this stuff is pretty risky because it could very well just be completely made up and I don't have the time to look through all of the specifics to ensure it's correct. If you can do so yourself and double check then that's great but I'm hesitant to include too much details because it could be incorrect. Look elsewhere throughout the repo to see how the formatting is done for the docstrings.

And for the example format, see chess.py as we have an example using the `>>>` format which gets tested under the doctests with pytest.

pettingzoo/utils/wrappers/supersuit/utils/base_aec_wrapper.py
```diff
@@ -0,0 +1,115 @@
+from pettingzoo.utils.wrappers import OrderEnforcingWrapper as PettingzooWrap
+from pettingzoo import AECEnv, ParallelEnv
+from pettingzoo.utils.env import ActionType, AgentID
+from gymnasium.spaces import Space
+
+
+class BaseWrapper(PettingzooWrap):
+    def __init__(self, env: AECEnv | ParallelEnv):
+        """
+        Creates a wrapper around `env`. Extend this class to create changes to the space.
```
elliottower, line 10: Also this looks incorrect saying `class BaseWrapper(PettingZooWrap)`, why would it use OrderEnforcingWrapper as the base class? And I'd honestly have to think more about what the base class should be, it's tough cause a lot of the functions like step() and reset() have different return types and aren't uniform between parallel and AEC.

elliottower:
Also run the pre-commit hooks as that will give you some feedback about docstrings adhering to the right format.

pettingzoo/utils/wrappers/supersuit/basic_wrappers.py
GiovanniGrotto: Ok thanks for the advice, I'll fix the docstring format and write them from scratch.

pettingzoo/utils/wrappers/supersuit/utils/base_aec_wrapper.py
GiovanniGrotto: I haven't really looked into the wrappers for the moment, this is just as it was on SuperSuit I can try to find a better base class for BaseWrapper

pettingzoo/utils/wrappers/supersuit/basic_wrappers.py
jjshoots: Agree, definitely not use the output of CGPT directly for docstrings.

jjshoots:
Looks fairly sensible so far, seems like these are just files taken from SS with docstrings added. For type hinting, I would recommend something like [the way Gymnasium does it](https://github.com/Farama-Foundation/Gymnasium/blob/main/gymnasium/core.py#L16), but ultimately this is Elliot's call. I'll wait for a more fleshed out version of the PR before doing a more in depth review. Ultimately, as long as the tests are comprehensive and they pass I don't see anything wrong.

elliottower:
> Looks fairly sensible so far, seems like these are just files taken from SS with docstrings added. For type hinting, I would recommend something like [the way Gymnasium does it](https://github.com/Farama-Foundation/Gymnasium/blob/main/gymnasium/core.py#L16), but ultimately this is Elliot's call. I'll wait for a more fleshed out version of the PR before doing a more in depth review. Ultimately, as long as the tests are comprehensive and they pass I don't see anything wrong.

Agreed, and we’ve added some better type hinting modeled after that for other internal files in pettingzoo, so that should be easy to replicate looking at env.py and stuff

elliottower:
Jets pr got merged which was an example of doubt SuperSuit style wrappers, you should try to follow it as an example

GiovanniGrotto:
> Jets pr got merged which was an example of doubt SuperSuit style wrappers, you should try to follow it as an example

Perfect I'll try to find some time and start with the migration

jjshoots:
@GiovanniGrotto For your reference, it's #1105 :)

jkterry1:
Closing this due to divergence and adding this example to notes for with @virgilt starts on this in the coming months

PR #1093, opened 2023-09-07
[DOC] Add version table to README.md

chrisyeh96:
# Description

I'd like to add a compatibility version table to the main README file to help myself and others track which versions of Gymnasium and Python are supported by each version of PettingZoo. This table is inspired by the table listed in the README of the [torchvision](https://github.com/pytorch/vision) package.

I've pieced together the compatibility information based on the [`pyproject.toml`](https://github.com/Farama-Foundation/PettingZoo/blob/master/pyproject.toml) file. However, I'm not 100% sure that I did this correctly, so I would appreciate a core PettingZoo maintainer double-checking this information.

## Type of change

Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ This does not include dependencies for all families of environments (some enviro
 
 To install the dependencies for one family, use `pip install pettingzoo[atari]`, or use `pip install pettingzoo[all]` to install all dependencies.
 
-PettingZoo supports Linux and macOS. We will accept PRs related to Windows, but do not officially support it.
+PettingZoo supports Python 3.8, 3.9, 3.10 and 3.11 on Linux and macOS. We will accept PRs related to Windows, but do not officially support it.
 
 The following table shows how PettingZoo versions correspond to supported Gymnasium and Python versions.
 
--- a/README.md
+++ b/README.md
@@ -26,7 +26,16 @@ This does not include dependencies for all families of environments (some enviro
 
 To install the dependencies for one family, use `pip install pettingzoo[atari]`, or use `pip install pettingzoo[all]` to install all dependencies.
 
-We support Python 3.8, 3.9, 3.10 and 3.11 on Linux and macOS. We will accept PRs related to Windows, but do not officially support it.
+PettingZoo supports Linux and macOS. We will accept PRs related to Windows, but do not officially support it.
+
+The following table shows how PettingZoo versions correspond to supported Gymnasium and Python versions.
+
+| Gymnasium | PettingZoo | Python |
+|-----------|------------|--------|
+| `0.29`    | `1.24`     | `>=3.8`, `<=3.11`
+| `0.28`    | `1.23`     | `>=3.7`, `<=3.11`
+| `0.27`    | `1.22`     | `>=3.7`, `<=3.11`
+| `0.26`    | `1.22`     | `>=3.7`, `<=3.10` (`3.11` supported in PettingZoo `>=1.22.3`)
 
 ## Getting started
 


elliottower:
Thanks for the contribution but the numbers are off, we always have 4 versions of Python we test on so when we removed 3.7 we added 3.11 as official support. So really it is only two data points, I believe 1.24 as you have here, and before. 

The thing is having text saying we support Python 3.8, 3.9, 3.10 and 3.11 is useful and makes it obvious without having to read a table, so if you could add that back that’d be good. Having a table for gymnasium versions and shimmy versions and SuperSuit versions though would be useful, as those are officially supported together with pettingzoo but have a few different versions and such.

chrisyeh96:
As far as I can tell, PettingZoo ...

- introduced support for for Python 3.11 in PettingZoo 1.22.3. See https://github.com/Farama-Foundation/PettingZoo/blob/1.22.3/setup.py.
- dropped support for Python 3.7 in PettingZoo 1.24.0. See https://github.com/Farama-Foundation/PettingZoo/blob/1.24.0/pyproject.toml

These versions also match what is mentioned in the [official PettingZoo release notes](https://pettingzoo.farama.org/release_notes/index.html):
- support for for Python 3.11 in PettingZoo 1.22.3: https://pettingzoo.farama.org/release_notes/index.html#release-1-22-3
- dropping support for Python 3.7 in PettingZoo 1.24.0: https://pettingzoo.farama.org/release_notes/index.html#release-1-24-0

I'm not familiar with shimmy and SuperSuit, which is why I did not include them in the table.

elliottower:
> I have added back in the text about supporting Python 3.8, 3.9, 3.10, and 3.11.
> 
> As far as I can tell, PettingZoo ...
> 
> * introduced support for for Python 3.11 in PettingZoo 1.22.3. See https://github.com/Farama-Foundation/PettingZoo/blob/1.22.3/setup.py.
> * dropped support for Python 3.7 in PettingZoo 1.24.0. See https://github.com/Farama-Foundation/PettingZoo/blob/1.24.0/pyproject.toml
> 
> These versions also match what is mentioned in the [official PettingZoo release notes](https://pettingzoo.farama.org/release_notes/index.html):
> 
> * support for for Python 3.11 in PettingZoo 1.22.3: https://pettingzoo.farama.org/release_notes/index.html#release-1-22-3
> * dropping support for Python 3.7 in PettingZoo 1.24.0: https://pettingzoo.farama.org/release_notes/index.html#release-1-24-0
> 
> I'm not familiar with shimmy and SuperSuit, which is why I did not include them in the table.

We didn’t actually fix Python 3.11 support until recently though, if you check the Python test workflow you can see we tested 3.7-3.10 until just recently. Will made a PR to try and support 3.11 a while back but it wasn’t fully working and the setup.py thing isn’t as important as our CI tests that can actually confirm things

elliottower:
The 3.11 support wasn’t fully working and there was an issue and PR that will had up about fixing it, this pr I did was what fully added it https://github.com/Farama-Foundation/PettingZoo/pull/1029 (not that any of this really matters that much but it’s good to not point people to use releases that weren’t tested on their environment)

elliottower:
Also I’ve talked with another core dev and I think we don’t really want to point people towards older versions of pettingzoo because there’s so many various issues with previous versions, it’s really best to just use the most recent release. Bigger packages like torch or numpy it makes sense, but the research community in general is already bad about updating package versions and keeping up with current Python version, etc. 

I guess as a user you find this would have been helpful to see though and that’s why you made this PR? Anyways not sure if this is actually something we want to be advertising, but maybe it’s okay to have if it’s more minimal and just shows the two release versions and corresponding gymnasium releases for Python 3.7-3.10 and 3.8-3.11

chrisyeh96:
Thanks for clarifying. So in your opinion, what should the table look like? Is the following more correct?

| Gymnasium | PettingZoo | Python
|-----------|------------|-------
| `0.29`    | `1.24`     | `>=3.8`, `<=3.11`
| `0.28`    | `1.23`     | `>=3.7`, `<=3.10`
| `0.27`    | `1.22`     | `>=3.7`, `<=3.10`
| `0.26`    | `1.22`     | `>=3.7`, `<=3.10`


Yeah, the reason I made this pull request is that I personally would really appreciate such a table.

I'm using PettingZoo to do 2 things, as part of a project that I call ["SustainGym"](https://chrisyeh96.github.io/sustaingym/):
1) create several custom PettingZoo environments
2) use [Ray RLLib](https://docs.ray.io/en/latest/rllib/index.html) to train multi-agent RL algorithms on a custom PettingZoo environment. However, Ray RLLib (v2.6.3) currently only supports the PettingZoo v1.22.3 API.

SustainGym's PettingZoo envs are wrappers around gymnasium envs. Thus, Sustaingym has both gymnasium and pettingzoo as dependencies, so I need to know which gymnasium version corresponds to which PettingZoo version (and vice versa).

Also, SustainGym has a dependency that only supports up to Python 3.9. Thus, having a table spell out PettingZoo's Python version support is also very helpful to me.

Finally, to train Ray RLLib algorithms on SustainGym, I realized that I couldn't use the latest PettingZoo API, but instead was forced to use PettingZoo v1.22.*. Therefore, I need this table to not only show the latest PettingZoo version support, but also several past versions.

elliottower:
You should be able to use the most recent pettingzoo with RLlib it’ll just be mad at you when you install pettingzoo and gymnasium first cause they have strict requirements. But you can install Ray and then do pip install pettingzoo==1.24.1 and it’ll work

chrisyeh96:
Unfortunately, the latest version of RLLib (v2.6.3) is *not* compatible with PettingZoo v1.22.*, at least in the way that I'm using it:

Here's how I'm implementing a custom multi-agent PettingZoo RL environment:

```python
from pettingzoo.utils.env import ParallelEnv

class MyEnv(ParallelEnv):
    ...
```

And here's how I'm using Ray RLLib to train agents on my custom environment:

```python
from ray.rllib.env.wrappers.pettingzoo_env import ParallelPettingZooEnv

env = MyEnv()
rllib_env = ParallelPettingZooEnv(env)

# then use rllib_env in a training loop
```

When I use RLLib v2.6.3, the code above will only work with PettingZoo v1.22.*. This is because Ray RLLib's `ParallelPettingZooEnv()` wrapper (which converts a PettingZoo `ParallelEnv` into an RLLib `MultiAgentEnv`) assumes that the PettingZoo env follows the PettingZoo v1.22.* API.

elliottower:
> Unfortunately, the latest version of RLLib (v2.6.3) is _not_ compatible with PettingZoo `v1.22.*`, at least in the way that I'm using it:
> 
> Here's how I'm implementing a custom multi-agent PettingZoo RL environment:
> 
> ```python
> from pettingzoo.utils.env import ParallelEnv
> 
> class MyEnv(ParallelEnv):
>     ...
> ```
> 
> And here's how I'm using Ray RLLib to train agents on my custom environment:
> 
> ```python
> from ray.rllib.env.wrappers.pettingzoo_env import ParallelPettingZooEnv
> 
> env = MyEnv()
> rllib_env = ParallelPettingZooEnv(env)
> 
> # then use rllib_env in a training loop
> ```
> 
> When I use RLLib v2.6.3, the code above will only work with PettingZoo `v1.22.*`. This is because Ray RLLib's `ParallelPettingZooEnv()` wrapper (which converts a PettingZoo `ParallelEnv` into an RLLib `MultiAgentEnv`) assumes that the PettingZoo env follows the PettingZoo `v1.22.*` API. RLLib's `ParallelPettingZooEnv()` wrapper is not yet compatible with newer PettingZoo versions.

I saw you commented a bit in RLlibs PRs, and appreciate you testing these things. The problem is that though the PR fixing RLlib to work with pettingzoo 1.23+ has been merged, it was not included in that release. So you will need to use the nightly install version: https://docs.ray.io/en/latest/ray-overview/installation.html#daily-releases-nightlies


Our tutorial mentions this as well, and I have verified it works with the nightly release https://pettingzoo.farama.org/tutorials/rllib/pistonball/

elliottower:
And we really don’t want to encourage people to use old versions of the package so I’m going to make the decision to not include this chart, packages like Ray we have been working hard to get to support the most recent version, and want to encourage that as much as possible.

Issue #1094, opened 2023-09-07
[Bug Report] CONTRIBUTING.md guide needs work


chrisyeh96:
### Describe the bug

I tried following the CONTRIBUTING.md guide to run `pytest` but it does not work.

I first did the following, which worked.

```bash
sudo apt update
sudo apt install swig

pip install -e .[all]
```

Then I tried running `pytest` but I received the following error message:

```
/workspaces/PettingZoo $ pytest -v
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: -n=auto
  inifile: /workspaces/PettingZoo/pyproject.toml
  rootdir: /workspaces/PettingZoo
```

I also tried using `make test-all`:

```
/workspaces/PettingZoo $ make test-all
pytest -v --cov=pettingzoo --cov-report term
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: -n=auto --cov=pettingzoo --cov-report
  inifile: /workspaces/PettingZoo/pyproject.toml
  rootdir: /workspaces/PettingZoo

make: *** [Makefile:11: test-all] Error 4
```

### System info

I installed PettingZoo in the default GitHub codespace configuration.

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/PettingZoo/issues) in the repo

elliottower:
Oh my fault i forgot to add pytest xdist to the pyproject dependencies, it allows distributed testing. Could you include it in one of your prs?

chrisyeh96:
I think the issue might be instead that CONTRIBUTING.md should say

```bash
pip install -e .[testing,all]
```
instead of

```bash
pip install -e .[all]
```

In [pyproject.toml](https://github.com/Farama-Foundation/PettingZoo/blob/master/pyproject.toml), it seems like the `all` option does not include any of the testing-related dependencies.

I've made a pull request with this change: #1097

PR #1095, opened 2023-09-07
Update parallel_test.py

chrisyeh96:
# Description

`live_agents` is already a set

## Type of change

Minor code improvement

# Tests

I tried following the instructions in [CONTRIBUTING.md](https://github.com/Farama-Foundation/PettingZoo/blob/master/CONTRIBUTING.md) but was unable to get `pytest` to work. See #1094.

### DIFF
--- a/pettingzoo/test/parallel_test.py
+++ b/pettingzoo/test/parallel_test.py
@@ -77,14 +77,13 @@ def parallel_api_test(par_env: ParallelEnv, num_cycles=1000):
             assert isinstance(truncated, dict)
             assert isinstance(info, dict)
 
-            agents_set = set(live_agents)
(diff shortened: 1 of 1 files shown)


elliottower (approved):

elliottower:
Thanks for catching this, if you could debug locally or confirm that it works exactly the same that would be good. But it’s just making a set out of an already existing set in the current code, so don’t see why this would change anything. Just good to check yknow.

chrisyeh96:
I can confirm locally that my change works exactly the same as before

jjshoots (approved):
LGTM actually, good catch. Not many people brave enough to dive into this code.

PR #1096, opened 2023-09-07
Update parallel_test.py to support Ray RLLib MultiAgentEnv

chrisyeh96:
# Description

I updated `parallel_test.py` so that it only checks the truncated/terminated status of agents in `live_agents` (which was created from `self.agents`). This allows `parallel_test.py` to not fail on an environment that tries to implement RLLib's `MultiAgentEnv` API, which requires the `terminated` and `truncated` dicts to include a special `'__all__'` key.

See https://github.com/ray-project/ray/blob/3e8a1dc6064225a5039a171b2dc5afb0beb7f0f8/rllib/env/multi_agent_env.py#L106C14-L106C21

With this change, an environment that inherits from both `pettingzoo.ParallelEnv` and `ray.rllib.env.MultiAgentEnv` will pass the PettingZoo `parallel_api_test()` checker:

```python
from ray.rllib.env import MultiAgentEnv
from sustaingym.envs.cogen import CogenEnv

class CustomMultiAgentEnv(ParallelEnv, MultiAgentEnv):
    ...
```

I do not believe that this pull request changes `parallel_test.py` in any other meaningful way.

# Tests

I tried following the instructions in [CONTRIBUTING.md](https://github.com/Farama-Foundation/PettingZoo/blob/master/CONTRIBUTING.md) but was unable to get `pytest` to work. See #1094.

### DIFF
--- a/pettingzoo/test/parallel_test.py
+++ b/pettingzoo/test/parallel_test.py
@@ -114,7 +114,8 @@ def parallel_api_test(par_env: ParallelEnv, num_cycles=1000):
                     agent
                 ), "action_space should return the exact same space object (not a copy) for an agent (ensures that action space seeding works as expected). Consider decorating your action_space(self, agent) method with @functools.lru_cache(maxsize=None)"
 
-            for agent in live_agents:
+            # we make a copy of live_agents because live_agents changes during iteration
+            for agent in set(live_agents):
                 if terminated[agent] or truncated[agent]:
                     live_agents.remove(agent)
 
--- a/pettingzoo/test/parallel_test.py
+++ b/pettingzoo/test/parallel_test.py
@@ -96,11 +96,8 @@ def parallel_api_test(par_env: ParallelEnv, num_cycles=1000):
 
                 has_finished |= {
                     agent
-                    for agent, d in [
-                        (x[0], x[1] or y[1])
-                        for x, y in zip(terminated.items(), truncated.items())
-                    ]
-                    if d
+                    for agent in live_agents
+                    if terminated[agent] or truncated[agent]
                 }
                 if not par_env.agents and has_finished != set(par_env.possible_agents):
                     warnings.warn(
@@ -117,11 +114,8 @@ def parallel_api_test(par_env: ParallelEnv, num_cycles=1000):
                     agent
                 ), "action_space should return the exact same space object (not a copy) for an agent (ensures that action space seeding works as expected). Consider decorating your action_space(self, agent) method with @functools.lru_cache(maxsize=None)"
 
-            for agent, d in [
-                (x[0], x[1] or y[1])
-                for x, y in zip(terminated.items(), truncated.items())
-            ]:
-                if d:
+            for agent in live_agents:
+                if terminated[agent] or truncated[agent]:
                     live_agents.remove(agent)
 
             assert (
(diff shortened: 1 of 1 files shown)


elliottower:
Oh I didn’t know about this about RLlib, thanks for the contribution though this is very useful.

pettingzoo/test/parallel_test.py
```diff
@@ -96,11 +96,8 @@ def parallel_api_test(par_env: ParallelEnv, num_cycles=1000):
 
                 has_finished |= {
                     agent
-                    for agent, d in [
-                        (x[0], x[1] or y[1])
```
elliottower, line 100: This x[0] code is confusing, can you explain it and do some testing to make sure you’re understanding it correctly? It looks like it says (terminated false, either true) basically.

jjshoots: Don't think that's necessary, the change does the same thing, may be a bit faster, and is easier to read as well.

elliottower:
Talked with jet a bit about the details (second pair of eyes) and looks good, only thing is it looks like the tests are failing now

elliottower:
Ahh okay I think the reason for two sets was this error: 

RuntimeError: Set changed size during iteration

So that part should be reverted but the rest looks good

elliottower:
Oh I’m confusing this for your other PR I think? Anyways if you could look into the tests and try to fix this that would be great.

chrisyeh96:
> Ahh okay I think the reason for two sets was this error:
> 
> RuntimeError: Set changed size during iteration
> 
> So that part should be reverted but the rest looks good


Ah, my bad. I accidentally iterated through the `live_agents` set while making changes to it. I've fixed this in a new commit.

pettingzoo/test/parallel_test.py
```diff
@@ -114,7 +114,8 @@ def parallel_api_test(par_env: ParallelEnv, num_cycles=1000):
                     agent
                 ), "action_space should return the exact same space object (not a copy) for an agent (ensures that action space seeding works as expected). Consider decorating your action_space(self, agent) method with @functools.lru_cache(maxsize=None)"
 
-            for agent in live_agents:
+            # we make a copy of live_agents because live_agents changes during iteration
+            for agent in set(live_agents):
```
jjshoots, line 118: Could we instead do something like:
```python
new_live_agents = []
for agent in live_agents:
  if terminated[agent] or truncated[agent]:
    break
  new_live_agents.append(agent)
live_agents = set(new_live_agents)
```

This would make it easier to read and for something like this to not happen in the future should someone else edit it.
Thanks!

jjshoots (changes requested):
See previous comment.

pettingzoo/test/parallel_test.py
chrisyeh96: I've rewritten it to completely eliminate this problem. Let me know what you think.

jjshoots: Awesome, thanks for the help!

jjshoots (approved):
Just needs precommit, otherwise LGTM. Thanks for the contribution!

chrisyeh96:
@jjshoots Is there anything I need to do on my end? I can't tell why pre-commit is failing.

jjshoots:
@chrisyeh96 you need to do black on parallel_test.py [trace here](https://github.com/Farama-Foundation/PettingZoo/actions/runs/6154237502/job/16699287555?pr=1096#step:7:65)

chrisyeh96:
@jjshoots Done.

jjshoots:
@elliottower LGTM.

elliottower (approved):

PR #1105, opened 2023-09-20
Add multi-episode wrapper for AEC and parallel envs

jjshoots:
# Description

Requested by @elliottower. This basically allows the wrapping of either an AEC or Parallel env into an environment which runs for multiple episodes before truncating.

When there are no more valid agents in the underlying environment, the environment is automatically reset.
When this happens, the `observation` and `info` returned by `step()` are replaced with that of the reset environment.
All `termination` and `truncation` signals are consumed internally, and the wrapped environment only returns a `truncation` signal after `num_episodes` have been run.
The result of this wrapper is that the environment is no longer Markovian around the environment reset.

## Type of change

# Checklist:


pettingzoo/utils/env.py
```diff
@@ -73,7 +73,6 @@ def reset(
         """Resets the environment to a starting state."""
         raise NotImplementedError
 
-    # TODO: Remove `Optional` type below
     def observe(self, agent: AgentID) -> ObsType | None:
```
elliottower, line 76: I feel like there was a reason for this being none, I think in the PR that added AgentID (can check the git blame here) they said that there were some examples that it returned none or something? Except that doesn't make sense, and seems like a bug. 

Maybe it's that some envs just update a self.observations dict or something? But yeah it should just be at least type hinted to show people that you are intended to return an ObsType.

elliottower (approved):
Looks good upon first glance will have to go through again in more detail later though

elliottower (approved):

elliottower:
Going to merge this as it’s a helpful utility on its own, though we are going to try to make a poker change that lets the chips persist for multiple hands which is an alternative to this.

PR #1109, opened 2023-09-27
Add num_rounds and num_chips to poker envs

jjshoots:
# Description

This allows the poker envs to have multi rounds and have persistent number of chips for the total episode.

This is accomplished through the `num_rounds` and `num_chips` parameter, and bootstraps off the MultiEpisodeWrappers.


pettingzoo/classic/rlcard_envs/texas_holdem.py
```diff
@@ -69,6 +72,7 @@
 
 ### Version History
 
+* v5: Added multiround option (1.x)
```
elliottower, line 75: this should be 1.24.2 as I don't want to do another major release for a while since there aren't any really breaking changes

pettingzoo/utils/wrappers/multi_episode_env.py
```diff
@@ -13,21 +13,30 @@ class MultiEpisodeEnv(BaseWrapper):
     When there are no more valid agents in the underlying environment, the environment is automatically reset.
     After `num_episodes` have been run internally, the environment terminates normally.
     The result of this wrapper is that the environment is no longer Markovian around the environment reset.
+
+    When `starting_utility` is used, all agents start with a base amount of health points (think of this as poker chips).
```
elliottower, line 17: This is an interesting way to do it, but I'm not sure if this is the best name for it. Can't really think of anything else besides `starting_reward` or something. Maybe it could be like `total_rewards` to indicate that it makes the rewards track between resets. Or `tally_rewards`. Feel like `starting_utility` doesn't mean anything to me if I don't know how it works. Would make sense and be simple if it was `total_rewards` adds or subtracts by round.

pettingzoo/utils/wrappers/multi_episode_env.py
```diff
@@ -13,21 +13,30 @@ class MultiEpisodeEnv(BaseWrapper):
...
-    def __init__(self, env: AECEnv, num_episodes: int):
+    def __init__(
+        self, env: AECEnv, num_episodes: int, starting_utility: float | None = None
+    ):
         """__init__.
 
         Args:
             env (AECEnv): env
             num_episodes (int): num_episodes
+            starting_utility (float | None): starting_utility
```
elliottower, line 31: It's odd but in these types of Args: parts I think type hinting in our other classes usually does `Optional[float]` rather than `| None`

pettingzoo/classic/rlcard_envs/texas_holdem.py
jjshoots: Ok

pettingzoo/utils/wrappers/multi_episode_env.py
```diff
@@ -59,6 +70,18 @@ def step(self, action: ActionType) -> None:
             None:
         """
         super().step(action)
+
+        # adjust utilities if this param is enabled
+        if self._starting_utility:
+            for agent in self.agents:
+                self._agent_utilities[agent] = (
+                    self._agent_utilities[agent] - self.rewards[agent]
```
elliottower, line 78: This looks like it always subtracts the reward. Shouldn't it just always add it, that way it's like total rewards and will add if you get positive reward or subtract if it's negative?

pettingzoo/utils/wrappers/multi_episode_env.py
jjshoots: Fixed.

elliottower (approved):
Looks mostly good but will test it locally a bit as well to see what the values are. I believe since the number of chips you start with is 100, and if you lose a round when you go all in you get -100 as reward, it should just represent the amount of chips you lost. But I'm not 100% sure if that's how the rewards for texas holdem work in our envs.

pettingzoo/utils/wrappers/multi_episode_env.py
jjshoots: I think utility makes more sense, this is a pretty standard way of saying "starting amount of meaningful substance", starting reward can be confused for the reward given to the agent for starting a new episode, while tally rewards sounds like it should be a boolean.

elliottower:
Ok so looks like the rewards are `+raised chips/2` for winner and `-raised chips/2` for loser, not sure where the divide by 2 comes from (raise chips I'm assuming is the amount that you raised? Or maybe it's the total pot divided by 2? What if it's more than 2 players? Maybe you could look into this and see if it makes sense or if there's anything on RLCard's paper or website that talks about it)

jjshoots:
Welp this was a dumb idea, closing.

elliottower:
(We’re going to do this specific to poker instead, as it sort of breaks other envs when you remove agents on reset)

PR #1121, opened 2023-10-20
fix: tictactoe doc

safarnejad:
# Description

No parallel API is not possible in tic-tac-toe. The docstring was wrongfully said that parallel API is available.

## Type of change

### DIFF
--- a/pettingzoo/classic/tictactoe/tictactoe.py
+++ b/pettingzoo/classic/tictactoe/tictactoe.py
@@ -12,7 +12,7 @@ This environment is part of the <a href='..'>classic environments</a>. Please re
 | Import             | `from pettingzoo.classic import tictactoe_v3` |
 |--------------------|-----------------------------------------------|
 | Actions            | Discrete                                      |
-| Parallel API       | Yes                                           |
+| Parallel API       | No                                            |
 | Manual Control     | No                                            |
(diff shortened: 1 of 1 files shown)


elliottower:
Hey thanks for the contribution. This is the case for all classic environments, I messaged another dev to get a second opinion because as you point out the code itself says parallelizable is false, so the docs should probably say that? Except you can do the turn based aec to parallel wrapper, and in that sense they still do work. I think the wrapper basically says “verify this yourself that the env works properly with this, we can’t guarantee all turn based aec encs will work the same with it”

I know in an unrelated project which uses pettingzoo classic envs, the turn based wrapper is used. Probably should say something on the classic homepage and on each environment explaining things.

jjshoots:
I think the `is_parallelizable` flag was invented before the wrapper was created. In that case, tic tac toe may have been wrongly labelled to be parallel friendly. This change makes sense with what Elliot is mentioning: you can parallel this environment, we just can't guarantee it would work.

elliottower:
We have been using the turn based to parallel wrapper extensively for another project and it works as intended, so I think I’m going to leave the parallel api true. If you wanted to make a note instead of this change to say that you need to use the wrapper, that would be okay. Otherwise I’ll probably do that

PR #1140, opened 2023-11-24
feature/provide access to wrapped attr

ffelten:
# Description

This fixes duplicated attributes in base wrapper classes and adds forwards through `__getattr__` in `BaseWrapper`. I also made possible to get env-specific attributes through `OrderEnforcing`, which was preventing all API extensions.
See: https://discord.com/channels/961771112864313344/1176822393910612038

Note: `conversions.py` are still a mess but it's more difficult to define a getattr since you convert API at the same time... So I left them untouched. It is however possible to retrieve env-specific attributes in converted env by using `.unwrapped`.

Example
```python
from pettingzoo.butterfly import knights_archers_zombies_v10
env = knights_archers_zombies_v10.env(render_mode=None)
converted_env = knights_archers_zombies_v10.parallel_env(render_mode=None)
print(env.max_cycles) # works
print(converted_env.max_cycles) # conversion shadows attr :(
print(converted_env.unwrapped.max_cycles) # this works though
```

Note2: I don't have the ROM for ALE installed locally so I couldn't run these on my machine.

## Type of change

# Checklist:

### DIFF
--- a/pettingzoo/atari/base_atari_env.py
+++ b/pettingzoo/atari/base_atari_env.py
@@ -169,11 +169,8 @@ class ParallelAtariEnv(ParallelEnv, EzPickle):
         self._screen = None
         self._seed(seed)
 
-    def _seed(self, seed=None):
-        if seed is None:
-            self.np_random, seed = seeding.np_random()
-        else:
-            self.np_random, seed = seeding.np_random(seed)
+    def _seed(self, seed):
+        self.np_random, seed = seeding.np_random(seed)
         self.ale.setInt(b"random_seed", seed)
         self.ale.loadROM(self.rom_path)
         self.ale.setMode(self.mode)
@@ -181,6 +178,8 @@ class ParallelAtariEnv(ParallelEnv, EzPickle):
     def reset(self, seed=None, options=None):
         if seed is not None:
             self._seed(seed=seed)
+        else:
+            self.np_random, seed = seeding.np_random()
         self.ale.reset_game()
         self.agents = self.possible_agents[:]
         self.terminations = {agent: False for agent in self.possible_agents}
--- a/pettingzoo/atari/base_atari_env.py
+++ b/pettingzoo/atari/base_atari_env.py
@@ -171,7 +171,9 @@ class ParallelAtariEnv(ParallelEnv, EzPickle):
 
     def _seed(self, seed=None):
         if seed is None:
-            _, seed = seeding.np_random()
+            self.np_random, seed = seeding.np_random()
+        else:
+            self.np_random, seed = seeding.np_random(seed)
         self.ale.setInt(b"random_seed", seed)
         self.ale.loadROM(self.rom_path)
         self.ale.setMode(self.mode)
--- a/pettingzoo/utils/wrappers/base.py
+++ b/pettingzoo/utils/wrappers/base.py
@@ -19,71 +19,11 @@ class BaseWrapper(AECEnv[AgentID, ObsType, ActionType]):
         super().__init__()
         self.env = env
 
-        try:
-            self.possible_agents = self.env.possible_agents
-        except AttributeError:
-            pass
-
-        self.metadata = self.env.metadata
-
-        # we don't want these defined as we don't want them used before they are gotten
-
-        # self.agent_selection = self.env.agent_selection
-
-        # self.rewards = self.env.rewards
-        # self.dones = self.env.dones
-
-        # we don't want to care one way or the other whether environments have an infos or not before reset
-        try:
-            self.infos = self.env.infos
-        except AttributeError:
-            pass
-
-        # Not every environment has the .state_space attribute implemented
-        try:
-            self.state_space = (
-                self.env.state_space  # pyright: ignore[reportGeneralTypeIssues]
-            )
-        except AttributeError:
-            pass
-
     def __getattr__(self, name: str) -> Any:
         """Returns an attribute with ``name``, unless ``name`` starts with an underscore."""
         if name.startswith("_"):
             raise AttributeError(f"accessing private attribute '{name}' is prohibited")
         return getattr(self.env, name)
-
-    @property
-    def observation_spaces(self) -> dict[AgentID, gymnasium.spaces.Space]:
-        warnings.warn(
-            "The `observation_spaces` dictionary is deprecated. Use the `observation_space` function instead."
-        )
-        try:
-            return {
-                agent: self.observation_space(agent) for agent in self.possible_agents
-            }
-        except AttributeError as e:
-            raise AttributeError(
-                "The base environment does not have an `observation_spaces` dict attribute. Use the environment's `observation_space` method instead"
-            ) from e
-
-    @property
-    def action_spaces(self) -> dict[AgentID, gymnasium.spaces.Space]:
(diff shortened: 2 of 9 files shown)


elliottower:
Looks like a weird error with one of the tests I can take a look at some point if you can’t figure it out. Those tests are a bit wonky because they have variable number of agents and different obs spaces and such (to test things work in those cases)

pettingzoo/atari/base_atari_env.py
```diff
@@ -171,7 +171,9 @@ def __init__(
 
     def _seed(self, seed=None):
         if seed is None:
-            _, seed = seeding.np_random()
+            self.np_random, seed = seeding.np_random()
+        else:
+            self.np_random, seed = seeding.np_random(seed)
```
elliottower, line 176: Looks good not sure why this wasn’t the case before

pettingzoo/utils/wrappers/base.py
```diff
@@ -19,72 +18,12 @@ def __init__(self, env: AECEnv[AgentID, ObsType, ActionType]):
...
-            pass
-
     def __getattr__(self, name: str) -> Any:
         """Returns an attribute with ``name``, unless ``name`` starts with an underscore."""
-        if name.startswith("_"):
+        if name.startswith("_") and name != "_cumulative_rewards":
             raise AttributeError(f"accessing private attribute '{name}' is prohibited")
         return getattr(self.env, name)
 
-    @property
```
elliottower, line 56: So removing this is because the underlying env has the action spaces et al defined and we just inherit directly through getattr?

pettingzoo/utils/wrappers/base.py
```diff
@@ -115,13 +46,11 @@ def state(self) -> np.ndarray:
...
-        self.rewards = self.env.rewards
-        self.terminations = self.env.terminations
-        self.truncations = self.env.truncations
-        self.infos = self.env.infos
-        self.agents = self.env.agents
-        self._cumulative_rewards = self.env._cumulative_rewards
+    def observation_space(self, agent: AgentID) -> gymnasium.spaces.Space:
+        return self.env.observation_space(agent)
+
+    def action_space(self, agent: AgentID) -> gymnasium.spaces.Space:
```
elliottower, line 52: Oh nvm you just moved it below I see

pettingzoo/utils/wrappers/base_parallel.py
```diff
@@ -1,40 +1,26 @@
...
 
     def reset(
         self, seed: int | None = None, options: dict | None = None
     ) -> tuple[dict[AgentID, ObsType], dict[AgentID, dict]]:
-        self.np_random, _ = seeding.np_random(seed)
-
-        res, info = self.env.reset(seed=seed, options=options)
-        self.agents = self.env.agents
-        return res, info
+        return self.env.reset(seed=seed, options=options)
```
elliottower, line 23: Is the reason for not setting self np random that you want that to be handled on the environment itself and not messed with by the wrapper

pettingzoo/utils/wrappers/base_parallel.py
```diff
@@ -62,32 +46,6 @@ def unwrapped(self) -> ParallelEnv:
     def state(self) -> np.ndarray:
         return self.env.state()
 
-    @property
```
elliottower, line 65: And what about removing this

pettingzoo/utils/wrappers/base_parallel.py
ffelten: Yup, makes no sense to me that it is the wrappers that set the random. I wanted to move it to `BaseAEC.reset()` at first but then figured that it would be overriden all the time...

elliottower (approved):
Overall looks good just some questions and tests to fix

pettingzoo/utils/wrappers/base_parallel.py
ffelten: Nothing related to this is implemented in this layer. These should be defined in the environment, or the wrapper modifying the obs space IMO?

elliottower: Makes sense, just saw that it was still there in the AEC base wrapper I think

pettingzoo/utils/wrappers/base_parallel.py
elliottower: Yeah makes sense

pettingzoo/atari/base_atari_env.py
jjshoots, line 176: You also need to fix line 182-183 of this file. Possibly remove the conditional and just do:
```py
self._seed(seed=seed)
```

elliottower: Oh nvm it looks like this is related to conversions.py, you mentioned you hadn't dived into that yet. Trying to think of what the best way to handle that is. I guess internally in there maybe we can just hard code adding code like `self.np_random = self.env.unwrapped.np_random` or something, but not sure if there's other things which are broken like this as well

ffelten: Argh, I think it is because the conversion wrappers do not give access to the attributes. It's much trickier to add this there since the API changes. I think the PR on supersuit side fixes this.

ffelten:
@pseudo-rnd-thoughts @jjshoots what do you guys think if we instead make this like Gymnasium 1.0 instead of using `__getattr__`? How painful would that be?

ffelten:
> I'm not quite aware of how Gymnasium 1.0 handles it. Could I get a TLDR?

https://github.com/Farama-Foundation/Gymnasium/pull/535 first point.

jjshoots:
That seems reasonable. I'm not a huge advocate for that idea, especially the whole "anywhere in the wrapper stack" thing, but I reckon lots of discussion has happened over on Gymnasium's end to conclude that that's the best way to do things. For that reason, I think it's OK.

As to how painful it has been, probably quite painful. Unlike Gymnasium which has very good in-code documentation, PZ has minimal. Lots of things will break that will require stack dives to fix.

ffelten:
> That seems reasonable. I'm not a huge advocate for that idea, especially the whole "anywhere in the wrapper stack" thing, but I reckon lots of discussion has happened over on Gymnasium's end to conclude that that's the best way to do things. For that reason, I think it's OK.
> 
> As to how painful it will be, probably quite painful. Unlike Gymnasium which has very good in-code documentation, PZ has minimal. Lots of things will break that will require stack dives to fix.

Well to summarize a bit all: This PR introduces the ability to access attributes from the underlying env. 

1. The breaking changes I foresee on the user side will probably be involving the duplicated variables formerly in the base wrappers that I removed here when using conversions (see the note in the PR description).

pseudo-rnd-thoughts:
I'm a bit confused by the changes as they don't seem all related to me.
To me, this should be a holistic thing for the whole of PZ, either you remove all `__getattr__` implementations everywhere or keep the code the same. This is the painful decision that we made in Gymnasium as we believed (while helpful) causes significant unknown issues for users

jjshoots:
@pseudo-rnd-thoughts Basically, PZ is transitioning from `env.env.env.attribute` to the idea of `__get_attr__`, but the natural question before merge is 'is this the right thing to do or should we be doing what Gymnasium does'.

PR #1153, opened 2024-01-15
Added single quotes around pip install arguments with square brackets

shahofblah:
>pip install pettingzoo[atari]

Results in zsh complaining as zsh by default uses [square brackets for globbing](https://stackoverflow.com/questions/30539798/zsh-no-matches-found-requestssecurity).

Putting pettingzoo[atari] and pettingzoo[all] inside single quotes makes these commands work with zsh.

# Description

Updated pip install commands in README.md to work with zsh.

## Type of change

- Documentation update

# Checklist:

### DIFF
--- a/docs/content/basic_usage.md
+++ b/docs/content/basic_usage.md
@@ -9,7 +9,7 @@ To install the base PettingZoo library: `pip install pettingzoo`.
 
 This does not include dependencies for all families of environments (some environments can be problematic to install on certain systems).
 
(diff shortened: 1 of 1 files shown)


elliottower:
Good catch, I believe some of our other documentation uses double quotes, any chance you could check to see that all usages of pip install are consistent? Not sure if double or single matters either

elliottower:
I'll fix the other ones, thanks for the PR

PR #1159, opened 2024-01-21
Remove redundant code in tutorials

jjshoots:
# Description

This removes some redundant code in the tutorials which add unnecessary confusion to new users.

## Type of change

### DIFF
--- a/docs/code_examples/aec_rps.py
+++ b/docs/code_examples/aec_rps.py
@@ -1,5 +1,3 @@
-import functools
-
 import gymnasium
 import numpy as np
 from gymnasium.spaces import Discrete
@@ -76,7 +74,9 @@ class raw_env(AECEnv):
         )
 
         # we want to define the spaces as fixed objects so we can seed them
-        self._observation_spaces = {agent: Discrete(4) for agent in self.possible_agents}
+        self._observation_spaces = {
+            agent: Discrete(4) for agent in self.possible_agents
+        }
         self._action_spaces = {agent: Discrete(3) for agent in self.possible_agents}
 
         # observation and action spaces are defined as functions which take in an agent id
--- a/docs/code_examples/aec_rps.py
+++ b/docs/code_examples/aec_rps.py
@@ -75,10 +75,14 @@ class raw_env(AECEnv):
             zip(self.possible_agents, list(range(len(self.possible_agents))))
         )
 
+        # we want to define the spaces as fixed objects so we can seed them
+        self._observation_spaces = {agent: Discrete(4) for agent in self.possible_agents}
+        self._action_spaces = {agent: Discrete(3) for agent in self.possible_agents}
+
         # observation and action spaces are defined as functions which take in an agent id
         # and returns the relevant spaces.
-        self.observation_space = lambda agent: Discrete(4)
-        self.action_space = lambda agent: Discrete(3)
+        self.observation_space = lambda agent: self.observation_spaces[agent]
+        self.action_space = lambda agent: self.action_spaces[agent]
         self.render_mode = render_mode
 
     def render(self):
--- a/docs/code_examples/aec_rps.py
+++ b/docs/code_examples/aec_rps.py
@@ -81,8 +81,8 @@ class raw_env(AECEnv):
 
         # observation and action spaces are defined as functions which take in an agent id
         # and returns the relevant spaces.
-        self.observation_space = lambda agent: self.observation_spaces[agent]
-        self.action_space = lambda agent: self.action_spaces[agent]
+        self.observation_space = lambda agent: self._observation_spaces[agent]
+        self.action_space = lambda agent: self._action_spaces[agent]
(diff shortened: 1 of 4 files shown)


jjshoots:
@elliottower I may have to take a different approach to this actually. Tests are failing because the space object have to return the exact same object.

elliottower:
Interesting commit messages hahah looks good though, the LRU cache I guess is not necessary for the api test to pass here? 

May still be worth mentioning the situations in which you do need the cache code (not 100% certain of when this is myself, I feel like even when they are separate fixed objects I’ve had to do the cache code). 

The tutorials might also benefit from explaining what exactly happens when you do env.seed() and explain that you in some cases want to seed action spaces or even observation spaces if there is anything stochastic in there (e.g., our generated agents test envs).

jjshoots:
Tbh the cache code seems redundant, `lru_cache` internally converts the function into a dict to retrieve past information, and that's kind of no different to use just calling the items of the dict directly. I don't think there's any conceivable speed gain.

jjshoots:
Ok turns out I was wrong, `lru_cache` is doing something else under the hood to make things marginally faster:
```python
import timeit
import functools

stuff: dict[str, int] = dict()
for i in range(1000):
    stuff[str(i)] = i

@functools.lru_cache(maxsize=None)
def get_stuff_lru(_id: str):
    return stuff[_id]

get_stuff_no_lru = lambda _id: stuff[_id]

def test_lru():
    for key in stuff.keys():
        get_stuff_lru(key)

def test_no_lru():
    for key in stuff.keys():
        get_stuff_no_lru(key)

print("Using lru: ", timeit.timeit("test_lru()", number=100000, setup="from __main__ import test_lru"))
print("Just a dict: ", timeit.timeit("test_no_lru()", number=100000, setup="from __main__ import test_no_lru"))
```

Result:
```
Using lru:  3.5688032259986358
Just a dict:  4.305833258000348
```

jjshoots:
Actually, I'm indifferent about leaving it in the error message. I think having the function return the same object in the error message is sufficient. The lru cache idea to me just seems like extra fluff in terms of what PZ should be. If users want to make their environments faster this should be the least of their concerns. I view this as sort of legacy code that's added to fix a the seeding tests that were introduced later.

elliottower:
> Actually, I'm indifferent about leaving it in the error message. I think having the function return the same object in the error message is sufficient. The lru cache idea to me just seems like extra fluff in terms of what PZ should be. If users want to make their environments faster this should be the least of their concerns. I view this as sort of legacy code that's added to fix a the seeding tests that were introduced later.

Makes sense, agreed that it is pretty hacky and fluff. I think if it does get removed it should be very clear how to return the correct action space or observation space, and the best practices for defining them, as removing the functools thing in the error message may lead to people being confused as to how to make the test pass.

Gymnasium errors linking to their documentation and specific pages seems like a great thing to have here, but the problem is our starter scripts (the AEC and parallel RPS ones at least) are pretty old and not necessarily the best practices either.

jjshoots:
I think the error message about returning the exact same object is reasonable. That's a pretty basic OOP concept.

elliottower:
> I think the error message about returning the exact same object is reasonable. That's a pretty basic OOP concept.

Fair enough, I was just thinking in terms of like, how some of the tutorials and envs define`self._action_spaces` as a dict or some just use `self.action_spaces` or some omit that and just do a lambda function, but if you do a lambda function you'd want to have this lru cache (if I'm understanding correctly). But those are things that we shouldn't be too heavy handed with and let people do whatever they want, best to leave the error messages to be as explicit and simple as possible

jjshoots:
With the lambda function, there is no need for lru cache, I think that's kind of the most pure Python way of doing it in a way that satisfies PZ's requirements.

elliottower:
> With the lambda function, there is no need for lru cache, I think that's kind of the most pure Python way of doing it in a way that satisfies PZ's requirements.
> 
> Your suggestion is fair, we can just leave the lru cache error message and remove lru cache from the tutorials.

Could you fully remove it from the other tutorials in this PR as well? This stuff also has it https://pettingzoo.farama.org/tutorials/custom_environment/2-environment-logic/#code

jjshoots:
Sure, got it. I'll do a ripgrep and nuke all instances of it except in the warnings once I'm back behind a bigger monitor.

jjshoots:
What the fk is that error, looks like it's related to #1160

elliottower:
Oh didn’t see this but that’s highly bizarre no idea why pytest plugins would be trying to be imported from pz classic

David-GERARD:
Hi @jjshoots , 

I'm PettingZoo's new project manager. I see that this PR is quite old, let me know if you think you'll come back to this project, or if I should close it as a wontfix.

Best,
David

David-GERARD:
Closing this PR as it has been inactive for some time. Ping me or email me if you want it reopened @jjshoots

PR #1166, opened 2024-01-24
Mirror Gymnasium's fix for the pyright upgrade

jjshoots:
This fixes the plethora by pinning the pyright version - mirroring what Gymnasium is doing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -73,7 +73,7 @@ repos:
             language: node
             pass_filenames: false
             types: [python]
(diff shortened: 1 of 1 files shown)


elliottower:
Are there currently pyright version issues? As far as I can tell those haven’t been a problem in a while

elliottower:
Oh just ran it locally as well and yeah they do seem to have changed something in a new version which was leading to errors in another PR, thanks for catching this.

elliottower (approved):

PR #1170, opened 2024-01-25
Correct obs type in parallel_rps doc example

dm-ackerman:
The value was an int, but all the testing expects observations to be numpy types. This has been changed with updated comments to explain.

## Type of change

# Checklist:

### DIFF
--- a/docs/code_examples/parallel_rps.py
+++ b/docs/code_examples/parallel_rps.py
@@ -1,6 +1,7 @@
 import functools
 
 import gymnasium
+import numpy as np
 from gymnasium.spaces import Discrete
 
 from pettingzoo import ParallelEnv
@@ -83,6 +84,7 @@ class parallel_env(ParallelEnv):
     @functools.lru_cache(maxsize=None)
     def observation_space(self, agent):
         # gymnasium spaces are defined and documented here: https://gymnasium.farama.org/api/spaces/
+        # Discrete(4) means an integer in range(0, 4)
         return Discrete(4)
 
     # Action space should be defined here.
@@ -128,7 +130,8 @@ class parallel_env(ParallelEnv):
         """
         self.agents = self.possible_agents[:]
         self.num_moves = 0
-        observations = {agent: NONE for agent in self.agents}
+        # the observations should be numpy arrays even if there is only one value
+        observations = {agent: np.array(NONE) for agent in self.agents}
         infos = {agent: {} for agent in self.agents}
         self.state = observations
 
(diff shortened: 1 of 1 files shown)


docs/code_examples/parallel_rps.py
```diff
@@ -161,9 +164,11 @@ def step(self, actions):
         env_truncation = self.num_moves >= NUM_ITERS
         truncations = {agent: env_truncation for agent in self.agents}
 
-        # current observation is just the other player's most recent action
+        # Current observation is just the other player's most recent action
+        # This is converted to a numpy value of type int to match the type
+        # that we declared in observation_space()
         observations = {
-            self.agents[i]: int(actions[self.agents[1 - i]])
+            self.agents[i]: np.array(actions[self.agents[1 - i]], dtype=np.int_)
```
elliottower, line 171: Is np.int_ the correct dtype?

docs/code_examples/parallel_rps.py
```diff
@@ -128,7 +130,8 @@ def reset(self, seed=None, options=None):
         """
         self.agents = self.possible_agents[:]
         self.num_moves = 0
-        observations = {agent: NONE for agent in self.agents}
+        # the observations should be numpy arrays even if there is only one value
+        observations = {agent: np.array(NONE) for agent in self.agents}
```
elliottower, line 134: And honestly this code is pretty old and should probably be totally redone, doing np.array(NONE) is super sloppy and confusing in itself as well.

docs/code_examples/parallel_rps.py
dm-ackerman: Actually, it looks like np.int64 may be the most correct choice. int_ is the same on most platforms, but maybe not all

docs/code_examples/parallel_rps.py
dm-ackerman: Yeah, the NONE was a confusing for me. I misread it as None and couldn't figure out why it wasn't working as expected

dm-ackerman: Maybe as a quick, slightly less confusing option, change NONE to NO_MOVE? That doesn't fix the root problem, but it avoids overlap with the None keyword

elliottower: Ohhh okay I re-read the code and I see it's just NO_MOVE = 3, okay that's clear and concise enough IMO. With parallel envs you can sometimes have agents do no move, so you need to handle that.

dm-ackerman:
I updated the type to int64, and made the initial state a bit more clear

docs/code_examples/parallel_rps.py
elliottower: If you could fix this I can merge it

dm-ackerman: I pushed a fix the pr

elliottower (approved):

PR #1175, opened 2024-02-05
Fix test failing due to __getattr__ in module

dm-ackerman:
# Description

The deprecated_handler() function is called by module level __getattr__ functions. It raised an ImportError when a module wasn't an env. This breaks getattr's behaviour when it is called with a default value. It expects __getattr__ to raise an AttributeError if the attr is not found.

Pytest was apparently updated to call getattr(obj, key, default) for several values of key. The import error isn't caught by getattr, breaking the test.

Changing ImportError to AttributeError fixes this. The missing attribute raises the correct error, which is caught by getattr, which sets the value to the default and everything works as intended.

## Type of change

# Checklist:

### DIFF
--- a/pettingzoo/utils/deprecated_module.py
+++ b/pettingzoo/utils/deprecated_module.py
@@ -37,7 +37,9 @@ def deprecated_handler(
         # It wasn't able to find this module
         # You should do your deprecation notice here.
         if not is_env(env_name):
-            raise ImportError(f"cannot import name '{env_name}' from '{module_name}'")
+            raise AttributeError(
+                f"cannot import name '{env_name}' from '{module_name}'"
+            )
         name, version = env_name.rsplit("_v")
 
         for loader, alt_env_name, is_pkg in pkgutil.iter_modules(module_path):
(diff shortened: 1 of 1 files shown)


dm-ackerman:
Well that's annoying, doc tests fail.

It works locally, but seems the workflow uses a different test command. This appears to be a problem with pytest-markdown-docs being incompatible with version 8 of pytest. A random person posted a fix and I opened an issue with pytest-markdown-docs asking them to add the changes and publish a new version. They were super responsive and looks like they are checking it out now. Hopefully it gets approved and sent out.

I patched pytest-markdown-docs locally and it fixes the error. So I'm optimistic that if they push an update, this test failure will be solved.

dm-ackerman:
The [pytest-markdown-docs folks were amazing and got pytestv8 support all sorted out](https://github.com/modal-labs/pytest-markdown-docs/issues/18). I upgrade the version and it works locally after correcting some problems it found in docs. Added a PR with those fixes, in addition to these, to hopefully be fully compatible with pytest v8: #1177 Closing this. Use that instead.

PR #1177, opened 2024-02-07
Add compatibility with pytest version 8

dm-ackerman:
# Description

Fixes a problem in deprecation handler that failed with version 8 of pytest

Upgraded to a version of pytest-markdown-docs that is compatible with pytest version 8.
That found several errors in the sample docs that have been fixed.

## Type of change

# Checklist:

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -78,11 +78,18 @@ The [AEC API](/api/aec/) supports sequential turn based environments, while the 
 Environments can be interacted with using a similar interface to [Gymnasium](https://gymnasium.farama.org):
 
 ```python
-  from pettingzoo.butterfly import knights_archers_zombies_v10
-  env = knights_archers_zombies_v10.env(render_mode="human")
-  env.reset(seed=42)
-  for agent in env.agent_iter():
-      observation, reward, termination, truncation, info = env.last()
-      action = policy(observation, agent)
-      env.step(action)
+from pettingzoo.butterfly import knights_archers_zombies_v10
+env = knights_archers_zombies_v10.env(render_mode="human")
+env.reset(seed=42)
+
+for agent in env.agent_iter():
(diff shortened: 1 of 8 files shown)


docs/index.md
elliottower, line 1: This one is a bit iffy because gymnasium still has policy(observation) in their home page pseudo code to show simplicity, but it does fail the docs tests cause it’s undefined. I’m going to defer this to @jjshoots to offer an opinion but my feeling is we should keep it as is

pyproject.toml
elliottower, line 1: We should probably just set specific versions of all these plugins so things don’t randomly break, then we can go through and update things from time to time. If you wanted to pin the current versions for the other purest things at least, these are only for devs so pinning versions isn’t a huge issue

docs/index.md
elliottower: Nvm looks like gymnasium had it the same way as this, looks good to me

pettingzoo/utils/deprecated_module.py
```diff
@@ -47,7 +49,7 @@ def deprecated_handler(
                     if int(alt_version) > int(version):
                         return DeprecatedModule(name, version, alt_version)
                     else:
-                        raise ImportError(
+                        raise AttributeError(
```
elliottower, line 52: It might be worth adding a comment as to why it’s an attribute error not an import error, just for the future

elliottower (approved):
Looks good besides the few things

pettingzoo/utils/deprecated_module.py
dm-ackerman: will do

dm-ackerman: Done

pyproject.toml
dm-ackerman: Done, all versions in testing are pinned

PR #1180, opened 2024-02-10
Add __setattr__ to wrappers. Fixes #1176

### LINKED ISSUE #1176: [Bug Report] TerminateIllegalWrapper breaks agent selection
### Describe the bug

When running an env using TerminateIllegalWrapper and not using the action mask, the agent selection becomes corrupted when an illegal move is made.

Here is an example from tictactoe (code below). Notice in the first game that player 1 starts (as expected) and it alternates between players 1 and 2 (as expected) until player 1 makes an illegal move, which is caught by the wrapper.
However, in the second game, player 1 makes two moves in a row. That should not happen. Also note that the illegal  move flagged is not actually illegal per the game rules.
This behaviour has been reported for other games.

```
New game
--------
calling reset(seed=42)
player_1 is making action: 1 current board:[0, 0, 0, 0, 0, 0, 0, 0, 0]
player_2 is making action: 5 current board:[0, 1, 0, 0, 0, 0, 0, 0, 0]
player_1 is making action: 7 current board:[0, 1, 0, 0, 0, 2, 0, 0, 0]
player_2 is making action: 4 current board:[0, 1, 0, 0, 0, 2, 0, 1, 0]
player_1 is making action: 1 current board:[0, 1, 0, 0, 2, 2, 0, 1, 0]
[WARNING]: Illegal move made, game terminating with current player losing. 
obs['action_mask'] contains a mask of all legal moves that can be chosen.

New game
--------
calling reset(seed=42)
player_1 is making action: 5 current board:[0, 0, 0, 0, 0, 0, 0, 0, 0]
player_1 is making action: 0 current board:[0, 0, 0, 0, 0, 1, 0, 0, 0]
[WARNING]: Illegal move made, game terminating with current player losing. 
obs['action_mask'] contains a mask of all legal moves that can be chosen.
```

### Code example

```shell
from pettingzoo.classic import tictactoe_v3

env = tictactoe_v3.env()

def do_game(seed):
    print("\nNew game")
    print("--------")
    
    print(f"calling reset(seed={seed})")
    env.reset(seed)
        observation, reward, termination, truncation, info = env.last()

        if termination or truncation:
            env.step(None)
        else:
            mask = observation["action_mask"]
            action = env.action_space(agent).sample()  # **no action_mask applied**
            print(f"{env.agent_selection} is making action: {action} current board:{env.board}")
            env.step(action)

do_game(42)
do_game(42)
```

### System info

```
>>> import sys; sys.version
>>> pettingzoo.__version__
'1.24.3'
```

### Additional context

_No response_

### Checklist

dm-ackerman:
# Description

Adding `__setattr__` ensures that variables get set to the appropriate location. By default, any public value set by the wrapper is sent to the env to be set there instead of on the wrapper. If a variable is meant to be set by the wrapper, it should be listed in the `_local_vars` class variable of the wrapper. This is not ideal, but seems to be the most reasonable design.

An example of needing to specify which vars to keep locally is here:
https://python-patterns.guide/gang-of-four/decorator-pattern/#implementing-dynamic-wrapper
The solution is to list which vars should be in the wrapper and check them when setting a value. That is the approach used in this commit, but more generalized.

In line with `__getattr__`, private values cannot be set on underlying envs. There are two exceptions:
`_cumulative_rewards` was previously exempted in `__getattr__` because it is used by many envs.
`_skip_agent_selection` is added because it is used by the dead step handling. If a wrapper can't set this, that functionality will break.

Fixes #1176, Depends on #1177 (for tests to pass)

## Type of change

# Checklist:

### DIFF
--- a/pettingzoo/utils/wrappers/base.py
+++ b/pettingzoo/utils/wrappers/base.py
@@ -14,15 +14,48 @@ class BaseWrapper(AECEnv[AgentID, ObsType, ActionType]):
     All AECEnv wrappers should inherit from this base class
     """
 
+    # This is a list of object variables (as strings), used by THIS wrapper,
+    # which should be stored by the wrapper object and not by the underlying
+    # environment. They are used to store information that the wrapper needs
+    # to behave correctly. The list is used by __setattr__() to determine where
+    # to store variables. It is very important that this list is correct to
+    # prevent confusing bugs.
+    # Wrappers inheriting from this class should include their own _local_vars
+    # list with object variables used by that class. Note that 'env' is hardcoded
+    # as part of the __setattr__ function so should not be included.
+    _local_vars = []
+
     def __init__(self, env: AECEnv[AgentID, ObsType, ActionType]):
         super().__init__()
         self.env = env
 
     def __getattr__(self, name: str) -> Any:
         """Returns an attribute with ``name``, unless ``name`` starts with an underscore."""
-        if name.startswith("_") and name != "_cumulative_rewards":
+        if name.startswith("_") and name not in [
+            "_cumulative_rewards",
+            "_skip_agent_selection",
+        ]:
             raise AttributeError(f"accessing private attribute '{name}' is prohibited")
         return getattr(self.env, name)
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        """Set attribute ``name`` if it is this class's value, otherwise send to env."""
+        # these are the attributes that can be set on this wrapper directly
+        if name == "env" or name in self._local_vars:
+            self.__dict__[name] = value
+        else:
+            # If this is being raised by your wrapper while you are trying to access
+            # a variable that is owned by the wrapper and NOT part of the env, you
+            # may have forgotten to add the variable to the _local_vars list.
+            if name.startswith("_") and name not in [
+                "_cumulative_rewards",
+                "_skip_agent_selection",
+            ]:
+                raise AttributeError(
+                    f"setting private attribute '{name}' is prohibited"
+                )
+            # send to the underlying environment to handle
+            setattr(self.__dict__["env"], name, value)
 
     @property
     def unwrapped(self) -> AECEnv:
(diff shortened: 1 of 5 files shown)


dm-ackerman:
I can't block the automated tests from running. I'm expecting them to fail because #1177 has not be merged yet. The failures should be resolved with that PR.

dm-ackerman:
Apparently this breaks a couple tutorials. The problem for a couple seems to be the use of supersuit wrappers that don't work with this change

elliottower:
I’m not a big fan of using _local_vars, I think we should do however Gymnasium wrappers are done with inheriting attributes and such, to ensure consistency between the two packages. They also have put a lot more thought into these kinds of decisions and have way more users who will point out design flaws so I trust their decisions

elliottower:
https://github.com/Farama-Foundation/Gymnasium/blob/c57f6d82219f35250b7cdd1b0c871687f9ea4e70/gymnasium/wrappers/common.py#L229

Looks like everything is done based off env.property instead of defining new properties (don’t do self.action space = self.env.action space for example)

dm-ackerman:
I'm looking at the gymnasium approach. PZ has different considerations so there are some complications. But may be able to pull their idea. I'll have to try that.

pseudo-rnd-thoughts:
I would do some performance testing but I believe this might impact performance massively as this is run on literally every variable update

If you **need** to add this, I think something else has gone horribly wrong and would only do this as a last, last, last solution

dm-ackerman:
Yeah, you may be right on performance. I saw the threads on Gymnasium mentioning that. I don't think this will be the final version. May do something more in line with Gymnasium, but PettingZoo wrappers have more issues than Gymnasium ones so there's more needed (I think) than just copying over the solution. I don't have access to test things until next week so can't do anything on this right now.

David-GERARD:
@dm-ackerman let me know if you're planning to do more work on this, otherwise I will close the PR before the next release

dm-ackerman:
I need to look back at this. the issue it's trying to solve is real and important, but as someone else mentioned, there is a performance hit that is significant. The approach gymanasium uses to solve this doesn't work here for reasons I can't recall. I am swamped with deadlines  but i will try to look at this in the next couple weeks.

David-GERARD:
@dm-ackerman sounds good, I'll leave it open then. Good luck with your deadlines!

jkterry1:
Hey, I'm going to close this and create reference to this PR in the notes for when we finally kill off supersuit and create pettingzoo.wrappers

PR #1192, opened 2024-03-20
Update the TicTacToe environment

dm-ackerman:
# Description

Update the TicTacToe environment to v4

Major changes:
* ~~Fix termination handling to work with stable baselines3. This now trains as expected with the SB3 PPO code~~
* Rewrite of underlying board code
* Add test functions for the underlying board code
* ~~Bump version to 4~~

Minor changes:
* adds some code comments
* removes unneeded code
* make board class more encapsulated
* only create rendering screen if rendering
* minor performance improvements (~30% faster)
* streamline win detection code
* remove redundant calculation of winning configurations

## Type of change

# Checklist:

- [X] I have run `pytest -v` and no errors are present.
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have added tests that prove my fix is effective or that my feature works


dm-ackerman:
AgileRL tutorial is broken again. I think I can fix that and I'll pin the version so it won't be an issue in the future

dm-ackerman:
Tests are failing due to AgileRL version change. #1193 should resolve that.

pettingzoo/classic/tictactoe/board.py
```diff
@@ -1,79 +1,113 @@
+class BadTicTacToeMoveException(Exception):
```
elliottower, line 1: This is a sloppy way of doing it, you should do it the same way the other envs handle illegal moves and action masks (terminate illegal wrapper)

pettingzoo/classic/tictactoe/tictactoe.py
```diff
@@ -153,88 +155,70 @@ def __init__(
         if self.render_mode == "human":
             self.clock = pygame.time.Clock()
 
-    # Key
```
elliottower, line 156: This comment seems useful, is it no longer accurate?

dm-ackerman: It was not correct. The observations are not a list of lists and they are not the values shown. They are either 1 or 0. I didn't change that, it was like before.

The board values themselves are what is shown and that is documented in the board class.

pettingzoo/classic/tictactoe/board.py
dm-ackerman: Illegal moves are still handled via the wrapper and masks if using the wrapper (still the default). This only gets triggered if using the raw env. The previous version of the raw env failed with an assertion on an illegal move. This one fails with an exception. I think that's a better approach.

pettingzoo/classic/tictactoe/test_board.py
elliottower, line 1: I like having specific env tests but should this go here? I guess it might be weird to have in the regular tests folder because those are usually more high level, but whatever standard we do here we should do for any other env which we test in the future, so we should try to make sure it’s in the right spot.

tutorials/SB3/test/test_sb3_action_mask.py
```diff
@@ -30,7 +30,7 @@
 MEDIUM_ENVS = [
     leduc_holdem_v4,  # with 10x as many steps it gets higher total rewards (9 vs -9), 0.52 winrate, and 0.92 vs 0.83 total scores
     hanabi_v5,  # even with 10x as many steps, total score seems to always be tied between the two agents
-    tictactoe_v3,  # even with 10x as many steps, agent still loses every time (most likely an error somewhere)
+    tictactoe_v4,  # even with 10x as many steps, agent still loses every time (most likely an error somewhere)
```
elliottower, line 33: So how does this new env fare here? Should the comment be updated? Funny that this would be a sign of the underlying env having issues, I assumed it was my tutorial code being bad (probably both to be fair)

pettingzoo/classic/tictactoe/test_board.py
dm-ackerman: I followed the precedent of chess, which has test_chess.py in the chess/ directory. I think it makes the most sense to put in the env directory as it is, but I'm fine if you want to move it.

pettingzoo/classic/tictactoe/board.py
dm-ackerman: Is the problem the fact that it's an exception or that it is a custom exception? If the latter, it can be replaced with a ValueError and still accomplish the same thing.

If the former, I was of the understanding that using assert to verify proper input is a misuse. Also, assertions can be disabled by -O which will change the program's behavior. That's one of the reasons the [google style guide](https://google.github.io/styleguide/pyguide.html#244-decision) (2nd bullet) says not to use them in this way.

Custom exceptions have value (though the tictactoe one may be too specific). If there was a consistently used IllegalMoveException across envs, I think the terminate illegal wrapper could be implemented clearer, more robustly, and with lower overhead (just catch that exception). You can't reasonably do that with an assert.

tutorials/SB3/test/test_sb3_action_mask.py
dm-ackerman: It trains much better. It needs a little more time than the easy step limit. At 4x as many steps, it passed 5 times in a row for me. I can update that comment/code tomorrow

pettingzoo/classic/tictactoe/test_board.py
elliottower: OK that’s fine, I forgot there was a chess one

pettingzoo/classic/tictactoe/board.py
elliottower: That’s fair, I think in that case we should try to make all of the envs do exceptions in this same way, it’s just skippy to have one env function differently. One thing I’ve seen used for these kinds of input type checking and stuff is pydantic (used by langchain for example) which makes the program fail with an error if the type is wrong, don’t think it would apply for illegal moves but a lot of the assertions we have are about the types of things, it’s arguable that we should do that and have more strict typing, but that worthless be a lot of work and diverge from other farama projects. Have been thinking about it for a while though and if we’re going to updating an env anyways and planning to do another breaking change I may just pull the trigger. Will probably talk to other devs to see their thoughts.

elliottower (approved):

pettingzoo/classic/tictactoe/board.py
dm-ackerman: Ok. Rather than bog this down, I'll change the exceptions back to asserts. I'll also add an issue when I get a chance proposing changing all the asserts for illegal moves to a (new) IllegalMoveError and changing the handling of the illegal move wrapper to match it. People can give feedback there.

tutorials/SB3/test/test_sb3_action_mask.py
elliottower: Awesome super appreciate your testing this

pettingzoo/classic/tictactoe/board.py
elliottower: OK sounds good. I may make an issue proposing pydantic and using TNT logging package, or other more modern ways of handling things vs asserts

elliottower: Made a post in the internal core devs channel about it will see what ppl say

pettingzoo/classic/tictactoe/test_board.py
```diff
@@ -113,16 +112,16 @@ def test_tictactoe_bad_move() -> None:
...
     # 1) move out of bounds should be rejected
     for outside_space in [-1, 9]:
-        with pytest.raises(BadTicTacToeMoveException):
+        with pytest.raises(AssertionError):
             board.play_turn(0, outside_space)
 
     # 2) move by unknown agent should be rejected
     for unknown_agent in [-1, 2]:
-        with pytest.raises(BadTicTacToeMoveException):
+        with pytest.raises(AssertionError):
```
elliottower, line 120: Why pytest raises? Haven’t seen that before

pettingzoo/classic/tictactoe/test_board.py
```diff
@@ -113,16 +112,16 @@ def test_tictactoe_bad_move() -> None:
...
             board.play_turn(unknown_agent, 0)
 
     # 3) move in occupied space by either agent should be rejected
     board.play_turn(0, 4)  # this is fine
     for agent in [0, 1]:
-        with pytest.raises(BadTicTacToeMoveException):
+        with pytest.raises(AssertionError):
```
elliottower, line 126: And I’m pretty sure you can do AssertionError(“illegal move made”) or other descriptions

pettingzoo/classic/tictactoe/test_board.py
dm-ackerman: As far as I know, that's the [recommended pytest way](https://docs.pytest.org/en/8.0.x/how-to/assert.html#assertions-about-expected-exceptions) to confirm that a block of code raises a specific assertion. It catches and ignores that exception (because it's expected) but raises an error if the code in that block doesn't raise the expected error.

elliottower: Ok fair, we should probably also go and make the code adhere to that as well, because I’m pretty sure elsewhere that isn’t done, but I can see it being good practice. Not a problem here tbh since it’s so small I think it’s fine to have here and then we can do an issue or future prs to fix other stuff to adhere to this practice

pettingzoo/classic/tictactoe/test_board.py
dm-ackerman: Yeah, good idea. Apparently there is a way to directly check the message of the assert to make sure it's the correct assert. I've added that.

elliottower (approved):

dm-ackerman:
Apparently the failure to train with SB3 wasn't a bug with the env, it was a SB3 tutorial bug. I removed the change that "fixed" the env because it was wrong. I believe the rest of the changes are valid, but I don't think it requires a version bump, so I moved it back to v3.

PR #1193, opened 2024-03-20
Pin AgileRL version

dm-ackerman:
# Description

The new version breaks the tutorial. Pinning to the previous version.

## Type of change

# Checklist:

- [X] I have run `pytest -v` and no errors are present.


dm-ackerman:
I was going to update the files to match the new version of AgileRL, which seemed straightforward. But it ended up also changing the trained models that are in the repo. I don't know whether that is supposed to happen, so I decided to just pin the version instead.

I'm not sure if we actually need those models to be stored in the repo.

elliottower (approved):

PR #1196, opened 2024-03-26
Agilerl updates

mikepratt1:
…ted version

# Description

1) Updating agilerl tutorials - updated key names within the net_config dicitonaries to match the most recent version of agilerl
2) Updated AgileRL versions - AgileRL is now no longer compatible with python 3.8 or less

## Type of change

# Checklist:

- [ x] I have run `pytest -v` and no errors are present.


.github/workflows/linux-tutorials-test.yml
elliottower, line 1: I don’t like having a completely separate CI for Agile, is there no way to do it within the same job but check if the Python version is 3.8 and not run that script if it is? Or somehow do it in a way that it’s still under the umbrella for all the other tutorials

tutorials/AgileRL/agilerl_matd3.py
```diff
@@ -21,7 +21,7 @@
     # Define the network configuration
     NET_CONFIG = {
         "arch": "mlp",  # Network architecture
-        "h_size": [32, 32],  # Actor hidden size
+        "hidden_size": [32, 32],  # Actor hidden size
```
elliottower, line 24: Minor but very bizarre that the original names were used, feel like it’s so unintuitive to have h_size or whatnot, especially for new users. Good to change it to be readable.

.github/workflows/linux-tutorials-test.yml
elliottower: Nevermind, was viewing this on my phone and didn't see that they still appear under Tutorial Tests, this is good enough

PR #1205, opened 2024-05-06
Order enforcing wrapper fix

### LINKED ISSUE #1174: [Bug Report] Incorrect behaviour from __getattr__ in OrderEnforcingWrapper?
### Describe the bug

I'm getting an error from OrderEnforcingWrapper when trying to access env.action_spaces.

It comes from `__getattr__`, which I included in the code section for context.

I see several issues with that function as it's written:
1) `value == "unwrapped"` will never be triggered because the base class has an unwrapped property which returns the same result
2) `value == "render_mode"` duplicates the effect of the base class's `__getattr__`.
3) both "observation_spaces" and "action_spaces" have a misleading message (referring to `possible_agents`), and more importantly, don't check whether the base class has the attribute before raising an error claiming it doesn't exist.
4) Lastly, all of the cases except for the last `elif` statement are not related to the stated goal of the docstring and the intent of the class (as I understand it).

The chained wrappers are a bit unclear to me and it looks like there was already some work related to this on #1140 , so I wanted to see if I'm misunderstanding anything before submitting a change.

I'd like to address the items above as follows:
1) remove this
2) remove this
3) return the spaces if they exist, error otherwise; fix the text of the error
4) move all except last `elif` to `BaseWrapper`'s `__getattr__`

### Code example

```shell
def __getattr__(self, value: str) -> Any:
    """Raises an error message when data is gotten from the env.

    Should only be gotten after reset
    """
    if value == "unwrapped":
        return self.env.unwrapped
    elif value == "render_mode" and hasattr(self.env, "render_mode"):
        return self.env.render_mode  # pyright: ignore[reportGeneralTypeIssues]
    elif value == "possible_agents":
        try:
            return self.env.possible_agents
        except AttributeError:
            EnvLogger.error_possible_agents_attribute_missing("possible_agents")
    elif value == "observation_spaces":
        raise AttributeError(
            "The base environment does not have an possible_agents attribute. Use the environments `observation_space` method instead"
        )
    elif value == "action_spaces":
        raise AttributeError(
            "The base environment does not have an possible_agents attribute. Use the environments `action_space` method instead"
        )
    elif value == "agent_order":
        raise AttributeError(
            "agent_order has been removed from the API. Please consider using agent_iter instead."
        )
    elif (
        value
        in {
            "rewards",
            "terminations",
            "truncations",
            "infos",
            "agent_selection",
            "num_agents",
            "agents",
        }
        and not self._has_reset
    ):
        raise AttributeError(f"{value} cannot be accessed before reset")
    else:
        return super().__getattr__(value)
```

### System info

```
>>> import sys; sys.version
>>> pettingzoo.__version__
'1.24.3'
```

### Additional context

_No response_

### Checklist

dm-ackerman:
# Description

Fix several issues related to `OrderEnforcingWrapper`

* Update docstrings to correctly describe what is happening
* remove checks in `__getattr__` that are unrelated to the wrapper (see #1174)
* fix pyright errors in `AECOrderEnforcingIterator`
* clean up a few unneeded `return` and `else` statements

Fixes #1174 

## Type of change

# Checklist:

### DIFF
--- a/pettingzoo/utils/env_logger.py
+++ b/pettingzoo/utils/env_logger.py
@@ -62,20 +62,6 @@ class EnvLogger:
         )
 
     @staticmethod
-    def warn_close_unrendered_env() -> None:
-        """Warns: ``[WARNING]: Called close on an unrendered environment.``."""
-        EnvLogger._generic_warning(
-            "[WARNING]: Called close on an unrendered environment."
-        )
-
-    @staticmethod
-    def warn_close_before_reset() -> None:
-        """Warns: ``[WARNING]: reset() needs to be called before close.``."""
-        EnvLogger._generic_warning(
-            "[WARNING]: reset() needs to be called before close."
-        )
-
-    @staticmethod
     def warn_on_illegal_move() -> None:
         """Warns: ``[WARNING]: Illegal move made, game terminating with current player losing.``."""
         EnvLogger._generic_warning(
--- a/pettingzoo/utils/wrappers/order_enforcing.py
+++ b/pettingzoo/utils/wrappers/order_enforcing.py
@@ -110,22 +110,15 @@ class AECOrderEnforcingIterator(AECIterator[AgentID, ObsType, ActionType]):
     def __init__(
         self, env: OrderEnforcingWrapper[AgentID, ObsType, ActionType], max_iter: int
     ):
-        assert hasattr(
-            env, "_has_updated"
+        assert isinstance(
+            env, OrderEnforcingWrapper
         ), "env must be wrapped by OrderEnforcingWrapper"
-        # this is set during the super call to init, so setting it here
-        # is redundant. However, it silences pyright errors because it tells
-        # pyright that self.env is an OrderEnforcingWrapper (which may not be
-        # strictly true, but it should have OrderEnforcingWrapper somewhere
-        # in the wrapper list). This might be better handled by Protocols,
-        # but this approach works.
-        self.env = env  # silence pyright errors
         super().__init__(env, max_iter)
 
     def __next__(self) -> AgentID:
         agent = super().__next__()
         assert (
-            self.env._has_updated
+            self.env._has_updated  # pyright: ignore[reportGeneralTypeIssues]
         ), "need to call step() or reset() in a loop over `agent_iter`"
-        self.env._has_updated = False
+        self.env._has_updated = False  # pyright: ignore[reportGeneralTypeIssues]
         return agent
--- a/pettingzoo/utils/wrappers/order_enforcing.py
+++ b/pettingzoo/utils/wrappers/order_enforcing.py
@@ -107,13 +107,25 @@ class AECOrderEnforcingIterable(AECIterable[AgentID, ObsType, ActionType]):
 
 
 class AECOrderEnforcingIterator(AECIterator[AgentID, ObsType, ActionType]):
+    def __init__(
+        self, env: OrderEnforcingWrapper[AgentID, ObsType, ActionType], max_iter: int
+    ):
+        assert hasattr(
+            env, "_has_updated"
+        ), "env must be wrapped by OrderEnforcingWrapper"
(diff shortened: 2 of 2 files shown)


pettingzoo/utils/env_logger.py
```diff
@@ -61,20 +61,6 @@ def warn_action_out_of_bound(
             f"[WARNING]: Received an action {action} that was outside action space {action_space}. Environment is {backup_policy}"
         )
 
-    @staticmethod
```
elliottower, line 64: Why exactly was this stuff removed? Looks kind of messy code and isn’t done in the other wrappers so can probably see it just want to make sure it’s on purpose that it got removed

pettingzoo/utils/wrappers/order_enforcing.py
```diff
@@ -134,13 +107,25 @@ def __iter__(self) -> AECOrderEnforcingIterator[AgentID, ObsType, ActionType]:
...
-            self.env, "_has_updated"
+            env, "_has_updated"
         ), "env must be wrapped by OrderEnforcingWrapper"
+        # this is set during the super call to init, so setting it here
+        # is redundant. However, it silences pyright errors because it tells
+        # pyright that self.env is an OrderEnforcingWrapper (which may not be
+        # strictly true, but it should have OrderEnforcingWrapper somewhere
+        # in the wrapper list). This might be better handled by Protocols,
+        # but this approach works.
+        self.env = env  # silence pyright errors
```
elliottower, line 122: Imo we’re better off just ignoring the pyright error directly vs setting a variable too early or having extra variables that aren’t set exactly how they should be, simpler that way

pettingzoo/utils/wrappers/order_enforcing.py
```diff
@@ -134,13 +107,25 @@ def __iter__(self) -> AECOrderEnforcingIterator[AgentID, ObsType, ActionType]:
...
+        super().__init__(env, max_iter)
+
+    def __next__(self) -> AgentID:
+        agent = super().__next__()
         assert (
-            self.env._has_updated  # pyright: ignore[reportGeneralTypeIssues]
+            self.env._has_updated
```
elliottower, line 128: I think we can just do ignore general type issues at the top of the file instead of these ugly in line ones (assuming that’s the main reason you changed this)

elliottower (approved):
Have a few qualms and don’t fully remember the issue but code looks fine to me upon first glance. Maybe it should be tested though that this previous error doesn’t happen? And hopefully put it in the same place that the other wrappers are tested

pettingzoo/utils/env_logger.py
dm-ackerman: It is dead code that is never called. I think it was maybe intended to be used in the OrderEnfocingWrapper - and maybe was used in an older version of it, but it's not used anymore. The code is specific to that wrapper, if it's not used there, there's no other place it would be useful.

pettingzoo/utils/wrappers/order_enforcing.py
dm-ackerman: fair enough. I changed this back.

pettingzoo/utils/wrappers/order_enforcing.py
dm-ackerman: I re-added this per above comment. I think the end of line one is better in this case, even though it is ugly. Ignoring any error that might show up in the file isn't great. This is a generic catch-all error too, so it has a bigger chance of hiding something important that we don't know about.

dm-ackerman:
> Have a few qualms and don’t fully remember the issue but code looks fine to me upon first glance. Maybe it should be tested though that this previous error doesn’t happen? And hopefully put it in the same place that the other wrappers are tested

The issue was mainly that the wrapper wasn't doing what it said it was and was doing things that didn't relate to enforcing order. I think this was all left over from other changes that were made previously.  There isn't really an error from that issue that can be tested, it was mostly noticing that the details didn't match the docs.

PR #1206, opened 2024-05-07
TerminateIllegalWrapper fix

### LINKED ISSUE #1176: [Bug Report] TerminateIllegalWrapper breaks agent selection

dm-ackerman:
# Description

Fix bug in TerminateIllegalWrapper that set values on the wrapper instead of the env.

Fixes #1176

## Type of change

# Checklist:


pettingzoo/utils/wrappers/terminate_illegal.py
```diff
@@ -42,7 +42,6 @@ def step(self, action: ActionType) -> None:
         if self._prev_obs is None:
             self.observe(self.agent_selection)
         if isinstance(self._prev_obs, dict):
-            assert self._prev_obs is not None
```
elliottower, line 45: Why’d you remove this assertion? Think it should check if it exists first before checking the value, but I guess the assertion will fail either way if not

elliottower (approved):
Overall looks good thanks for making the test too, though it should probably be in the file where the other wrapper tests are done just for simplicity. And have we changed all of the other wrappers to do this self.env.unwrapped business yet?

pettingzoo/utils/wrappers/terminate_illegal.py
dm-ackerman: I removed it because it is redundant and will never be called if the value is None.
The line before confirms self._prev_obs is a dict. If it's a dict, it's not None, so the assertion can never fail.

dm-ackerman:
> Overall looks good thanks for making the test too, though it should probably be in the file where the other wrapper tests are done just for simplicity. And have we changed all of the other wrappers to do this self.env.unwrapped business yet?

This was the only one I noticed where unwrapped matters.

I moved the wrapper test in with the others

jjshoots (approved):
Looks pretty reasonable to me. The test for terminate illegal is a bit hack with the double call to the env rollout, but otherwise looks sane enough.

PR #1209, opened 2024-05-20
Local Ratio should not be 0

SamuelMoor-Smith:
# Description

1

## Type of change

But

### Screenshots

# Checklist:


jjshoots:
I don't think this is a good idea. A better approach would be to fix the reason that `local_ratio` is needed in the first place. Leaving this open for now.

jkterry1:
I was the original author of the environment, and I agree that this change doesn't seem like a good default

PR #1218, opened 2024-06-24
Fix memory leak in waterworld_v4 by clearing old handlers before adding new ones

florin-pop:
# Description

We recently tested waterworld_v4 with rllib and noticed that it was slowly leaking memory. Over 1M episodes it would slowly fill the 128GB of memory allocated to the test environment. One culprit seems to be the `handlers` list which is not cleared when resetting the environment so each time `add_handlers()` is called, new collision handlers are created and appended to this list. 

## Type of change

### Screenshots

Before fix

Screenshot from 2024-06-24 05-08-14

After fix

Screenshot from 2024-06-24 05-07-37

# Checklist:


jjshoots (approved):
Great catch! Thanks for finding this!

PR #1229, opened 2024-09-11
Fix tictactoe reset bug (when one agent died)

Cheaple:
# Description

In a tictactoe environment `env`, if `env` get resetted when an possible agent is not in the `env.agents` list (for invalid action or other reasons), that agent would not be selected forever, since `env._agent_selector` always gets re-initialized with the list `env.agent` which misses that agent. To fix this situation, `env._agent_selector` should get re-initialized with `env.possible_agents` (which contains all possible agents.) rather than `env.agents`.

## Type of change

- 
### Screenshots

# Checklist:

  - only `Format YAML files.` failed.
  - there is an error, which also happens before my commit

### DIFF
--- a/pettingzoo/classic/tictactoe/tictactoe.py
+++ b/pettingzoo/classic/tictactoe/tictactoe.py
@@ -226,7 +226,7 @@ class raw_env(AECEnv, EzPickle):
         self.truncations = {i: False for i in self.agents}
         self.infos = {i: {} for i in self.agents}
         # selects the first agent
-        self._agent_selector.reinit(self.agents)
+        self._agent_selector.reinit(self.possible_agents)
         self.agent_selection = self._agent_selector.reset()
 
         if self.render_mode is not None and self.screen is None:


benblack769:
Hi, can you post a particular set of inputs/calls which causes the failure case you found? The current code sets `self.agents = self.possible_agents[:]` at the beginning of the reset function, which means that all agents should be included after reset. 

Additionally, this environment follows the same pattern as the chess `https://github.com/Farama-Foundation/PettingZoo/blob/master/pettingzoo/classic/chess/chess.py#L241` environments. So I'm wondering what sort of inputs could cause the error mentioned.

David-GERARD:
@Cheaple can you respond to @benblack769, and update your fork to the latest version of PettingZoo master?

jkterry1:
@Cheaple this also needs to bump the environment version number

jkterry1:
I'm closing this and moving it to an issue since this will require a environment version bump and related items not in this PR and additional further investigation

PR #1246, opened 2024-12-03
[Feat] Deprecate Python 3.8, add Python 3.12 + update CI/CD

David-GERARD:
# Description

As planned for the release of version 1.24.4, we are removing support for python >3.9, and adding python 3.12 .

## Type of change

# Checklist:

### DIFF
--- a/tutorials/AgileRL/requirements.txt
+++ b/tutorials/AgileRL/requirements.txt
@@ -1,7 +1,4 @@
-# Enforce Python version constraint
-python_version < '3.12'
-
-agilerl==0.1.22; python_version >= '3.9'
+agilerl==0.1.22; python_version >= '3.9' and python_version < '3.12'
 pettingzoo[classic,atari,mpe]>=1.23.1
 SuperSuit>=3.9.0
 torch>=2.0.1
--- a/tutorials/AgileRL/requirements.txt
+++ b/tutorials/AgileRL/requirements.txt
@@ -1,3 +1,6 @@
+# Enforce Python version constraint
+python_version < '3.12'
+
 agilerl==0.1.22; python_version >= '3.9'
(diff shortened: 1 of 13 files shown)


David-GERARD:
Pygame doesn't support python 3.13 yet, and due to the number of errors in CI, I am removing it and focussing this PR on python 3.12 only.

David-GERARD:
As per https://github.com/gorakhargosh/pathtools/issues/13, pathtools, a dependancy of AgileRL doesn't build python 3.12, and it doesn't seem like they will change it (see closed PR https://github.com/gorakhargosh/pathtools/pull/14).

As such:

- I removed python 3.12 from the AgileRL tutorial
- I pinned the version in requirements.txt to be <3.12
- I added a warning in the documentation in `PettingZoo/docs/tutorials/agilerl/index.md

tutorials/AgileRL/requirements.txt
```diff
@@ -1,4 +1,4 @@
-agilerl==0.1.22; python_version >= '3.9'
+agilerl==0.1.22; python_version >= '3.9' and python_version < '3.12'
```
David-GERARD, line 1: @mwydmuch I've added this to signal that the AgileRL tutorial fails with python 3.12, can you confirm this is good practice?

mwydmuch: I was unsure about this marker syntax, but after checking, it seems to be the correct way to define it. Overall, I think we have no choice. It would be a bit to limit PettingZoo to Python >=3.9,<3.12 only because of one submodule.

mwydmuch (approved):

tutorials/AgileRL/requirements.txt
David-GERARD: Agreed, thanks!

PR #1277, opened 2025-03-14
Make AECEnv/ParallelEnv rps examples deterministic by setting the RNG seed properly.

### LINKED ISSUE #1223: RPS is not deterministic and not seeded correctly.
### Describe the bug

I am just running the rock paper scissor example of pettingzoo [here](https://pettingzoo.farama.org/content/environment_creation/). I found out if I run the code multiple times, I get different plays. This is a bug. I think it is because the passed seed is not used at all in the code. However, I am not sure how we seed the environments in petting zoo. In Gym, we seed, and then use the env.np_seed to generate stuff if IIRC. 

### Code example

_No response_

### System info

_No response_

### Additional context

_No response_

### Checklist

yjhan96:
# Description

Previously, both AECEnv and ParallelEnv versions of RPS examples in the doc were not running deterministically even though the example set the seed explicitly. This was because both environments silently ignored the set seed. This PR fixes such issue by setting the random seed explicitly in the respective environments using `gymnasium.utils.seeding`.

Fixes #1223 

## Type of change

Bug fix (non-breaking change which fixes an issue)

# Checklist:

### DIFF
--- a/docs/code_examples/aec_rps_usage.py
+++ b/docs/code_examples/aec_rps_usage.py
@@ -1,4 +1,4 @@
-import aec_rps
+from . import aec_rps
 
 env = aec_rps.env(render_mode="human")
 env.reset(seed=42)
--- a/docs/code_examples/aec_rps_usage.py
+++ b/docs/code_examples/aec_rps_usage.py
@@ -1,4 +1,4 @@
-from . import aec_rps
+import aec_rps
 
 env = aec_rps.env(render_mode="human")
 env.reset(seed=42)
--- a/docs/code_examples/parallel_rps_usage.py
+++ b/docs/code_examples/parallel_rps_usage.py
@@ -1,4 +1,4 @@
-import parallel_rps
+from . import parallel_rps
 
 env = parallel_rps.parallel_env(render_mode="human")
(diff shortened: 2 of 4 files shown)


docs/code_examples/aec_rps_usage.py
```diff
@@ -1,4 +1,4 @@
-from . import aec_rps
+import aec_rps
```
David-GERARD, line 1: The problem with this change is that it make the documentation test workflow fail as it's not treated as a relative import. 

Ideas for fixing the bug:

- Revert to using from . import aec_rps
- Use os and sys to set the path correctly an enable the import

docs/code_examples/parallel_rps_usage.py
```diff
@@ -1,7 +1,7 @@
-from . import parallel_rps
+import parallel_rps
```
David-GERARD, line 1: Same thing here

David-GERARD (changes requested):

David-GERARD (approved):

David-GERARD:
@yjhan96 LGTM, I'll let you merge ;)

PR #1280, opened 2025-03-27
[FEAT]: Add deprecation warnings to MPE environment to prepare for their future migration to the MPE2 package

### LINKED ISSUE #1265: [Feat] Add deprecation warnings to MPE environments and documentation
Add [Deprecation Warnings](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) to the MPE environments in PettingZoo in anticipation of implementation of issues Farama-Foundation/MPE2#13 and #1250.

David-GERARD:
# Description
In a future version, the MPE environments will be removed from PettingZoo, and will have to be installed from [MPE2](https://mpe2.farama.org).

This PR aims at giving notice for this future change, without breaking codebase right away.

> [!NOTE]
> This is my first time implementing deprecation warnings, and need some feedback on good practice, as well as what unit testing (if any) to implement.

Closes #1265 

## Type of change

- Soon to be breaking change 

# Checklist:

### DIFF
--- a/pettingzoo/mpe/__init__.py
+++ b/pettingzoo/mpe/__init__.py
@@ -3,3 +3,7 @@ from pettingzoo.utils.deprecated_module import deprecated_handler
 
 def __getattr__(env_name):
     return deprecated_handler(env_name, __path__, __name__)
+
+
+# TODO: in a future release, remove all the MPE environments and uncomment the following lines to raise a warning
+# instead of importing the environments
--- a/pettingzoo/mpe/__init__.py
+++ b/pettingzoo/mpe/__init__.py
@@ -1,9 +1,18 @@
+import warnings
+
 from pettingzoo.utils.deprecated_module import deprecated_handler
 
 
 def __getattr__(env_name):
     return deprecated_handler(env_name, __path__, __name__)
 
+
+warnings.warn(
+    "The environment `pettingzoo.mpe` has been moved to `mpe2` and will be removed in a future release."
+    "Please update your imports.",
+    DeprecationWarning,
+    stacklevel=2,
+)
 
 # TODO: in a future release, remove all the MPE environments and uncomment the following lines to raise a warning
 # instead of importing the environments
--- a/pettingzoo/mpe/simple/simple.py
+++ b/pettingzoo/mpe/simple/simple.py
@@ -7,6 +7,14 @@
 :name: simple
 ```
 
+```{eval-rst}
+.. warning::
+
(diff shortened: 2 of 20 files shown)


pettingzoo/mpe/__init__.py
```diff
@@ -1,5 +1,18 @@
+import warnings
+
 from pettingzoo.utils.deprecated_module import deprecated_handler
 
 
 def __getattr__(env_name):
     return deprecated_handler(env_name, __path__, __name__)
+
+
+warnings.warn(
```
yjhan96, line 10: Should we create a helper function in utils/ for warnings instead?

pettingzoo/mpe/simple/simple.py
```diff
@@ -7,6 +7,14 @@
...
 ```
 
+```{eval-rst}
+.. warning::
+
+    The environment `pettingzoo.mpe.simple_v3` has been moved to the new `MPE2 package <https://mpe2.farama.org>`_, and will be removed from PettingZoo in a future release.
+    Please update your import to `mpe2.simple_v3`.
+
+```
+
```
yjhan96, line 17: It'd be good to use @warnings.deprecated decorator to throw deprecation warning when the class defined in this file was used. But I think warnings.deprecated came out in Python 3.12, so it may not be supported for previous Python versions.

yjhan96 (commented):
Overall it looks reasonable - left some minor refactor comments.

Do we plan to introduce mpe2/ directory in the near future?

pettingzoo/mpe/simple/simple.py
David-GERARD: @yjhan96 can you link an example? I can't find any. But yes if it's only supported for python 3.12 then we can't use it because we need it to work for >=3.9.

pettingzoo/mpe/__init__.py
David-GERARD: So I though about doing that, but since this is meant as temporary before the removal of mpe from pettingzoo + it's a fairly simple warning so a function would be a wrapper without any processing, I didn't go with it. 

If you think there's a good reason to make one though I can make it.

yjhan96: Since it's temporary, leaving as is sounds good to me.

pettingzoo/mpe/simple/simple.py
yjhan96: https://docs.python.org/3/library/warnings.html#warnings.deprecated shows some examples of how @warnings.deprecated is used.

yjhan96 (approved):
Looks good to me. Let's ship it! Thanks for the PR.

PR #1281, opened 2025-04-10
[TEST, DO NOT MERGE]: pin gymnasium to 1.1.0 to test compatibility

David-GERARD:
# DO NOT MERGE

This PR is just to check that pinning gymnasium to 1.1.0 doesn't result in bugs.


David-GERARD:
@jkterry1 I confirm that the ci passes with gymnasium pinned to 1.1.0

PR #1284, opened 2025-04-15
Ignore pygame deprecation warnings during testing

dm-ackerman:
# Description

pygame is using a deprecated packaging api. Since it has been raising warnings for years, it seems unlikely that the pygame devs are interested in fixing that any time soon. This change ignores those warnings during testing. If pygame devs ever see the light and correct this, these filters should be removed.

## Type of change

# Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,3 +106,7 @@ reportMissingImports = false
 
 [tool.pytest.ini_options]
 addopts = [ "--ignore-glob=*/__init__.py", "-n=auto", "--ignore=tutorials", "--ignore=docs/_scripts", "--ignore=conf.py"]
+filterwarnings = [
+    "ignore::DeprecationWarning:pygame.*",
+    "ignore::DeprecationWarning:pkg_resources.*"
+]


David-GERARD:
@dm-ackerman is the warning really an issue? What is improved by ignoring it?

dm-ackerman:
Several reasons:
* The warning isn't useful as we can't do anything about it.
* We ask people submitting PRs to resolve warnings created by their changes, but the base code throws a ton of warnings (about 200 on my system). That sends the message that we don't actually care if people fix warnings, which encourages them not to, which leads to more warnings.
* It normalizes deviance by making warnings OK, which reduces the value of warnings by encouraging people to ignore them. Warnings become a flood of meaningless noise rather than actionable items to look at.
* It makes it less obvious when changes cause a warning (3 warnings when there used to be 0 is obvious. The current number varies with the number of processes, so it's not always obvious that the number has increased).

I concede that none of these rises to the level of major problem, but the change to fix it isn't major either. It's a very targeted filtering of a specific type of warning from two specific sources.

Is there a reason not to do this?

PR #1308, opened 2025-09-03
Docs: replace third-party dilemmas environment

mrochk:
# Description

I have rewritten the third-party environment called "PettingZoo Dilemma Envs" in a more simple way as I felt like the code was quite messy and hard to use. 

## Type of change

Minimal documentation change (redirect to my repository instead).


jkterry1:
Hey,

I really like your work.

Could you please create a new PR that lists both works instead of just yours until one gains clear community preference?

PR #1335, opened 2026-05-24
Fix gin_rummy knock/gin reward reverting to RLCard default on seeded reset (#1312)

### LINKED ISSUE #1312: [Bug Report]  Gin Rummy environment seed / pay-off inconsistency for RLCard
### Describe the bug

I’m using the gin_rummy_v4 environment from PettingZoo’s classic set (which wraps the RLCard implementation), and I’ve discovered two related issues:

1. The seed parameter (or overall seeding behavior) of the environment does not seem to follow the usual env.reset(seed=…) semantics in PettingZoo. In particular, the underlying RLCard game seems to be used via rlcard.make(…) and thus the seed applied in PettingZoo does not get consistently propagated to the RLCard backend.
2.  Because of (1) the payoff/reward defaults appear inconsistent with the documented defaults in PettingZoo. The documentation says for the Gin Rummy environment (wrapped by RLCard) that the default knock reward is 0.5, but the underlying RLCard defaults use 0.2 as the knock reward.  ￼ Hence I’m seeing “knock amount = 0.2” instead of 0.5 when I use default settings or even custom setting.
3. These two issues together (seed not respected + inconsistent reward default) cause reproducibility problems and unexpected payoff behavior when using the environment for research or benchmarking.

### Code example

```shell
from pettingzoo.classic import gin_rummy_v4
env = gin_rummy_v4.env()
env.reset(seed=seed)
```

### System info

- PettingZoo version: (pettingzoo[classic]>=1.24.3)
- RLCard version: (ray[rllib]>=2.31.0)
- Python version: (e.g., 3.11.14)
- OS: (MacOS 26, Linux 24.04)

### Additional context

Suggested Fix / Ask:

1. Ensure that in the gin_rummy wrapper (and any other RLCard-wrapped environment) the reset(seed=…) value is passed properly to RLCard’s make or internal RNG so that the seed is effective and reproducible.
2. Force the default knock_reward and gin_reward to the documented values (0.5 and 1.0 respectively) regardless of RLCard’s internal defaults – or at minimum document clearly when RLCard defaults differ.
3. Add automated tests to verify that the seed parameter produces deterministic results (e.g., fixed draws, shuffle order) and that reward defaults match documentation.
4. Optionally, update the documentation to more clearly state that because RLCard is used under the hood, some internal defaults (like 0.2) may still apply unless overridden, or update the wrapper to guarantee the override.

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/PettingZoo/issues) in the repo

Nikelroid:
Fixes #1312

## Problem

`gin_rummy` ignores the configured `knock_reward` / `gin_reward` whenever the env
is seeded. `reset(seed=...)` silently reverts them to RLCard's defaults
(knock `0.2`, gin `1.0`), breaking reproducible experiments.

## Cause

`__init__` patches the scorer once (`self.env.game.judge.scorer.get_payoff =
self._get_payoff`), but `RLCardBase._seed()` rebuilds `self.env` via
`rlcard.make()`, which restores RLCard's default scorer. Since `reset()` only
calls `_seed()` when a seed is given, the bug appears only on seeded resets.

## Fix

Re-apply the override after reseeding, in a `gin_rummy`-local `_seed`:

```python
def _seed(self, seed=None):
    super()._seed(seed=seed)
    self.env.game.judge.scorer.get_payoff = self._get_payoff
```

I avoided the base-class alternative (reseed in place via `self.env.seed(seed)`):
it breaks `texas_holdem_no_limit_v6`, whose dealer button is randomized only on a
fresh game object, so reusing the env makes seeding non-deterministic there.

## Tests

New `pettingzoo/classic/rlcard_envs/test_gin_rummy.py`:
- configured scorer survives `reset(seed=...)`;
- a seeded knock pays `0.5`, not `0.2` (fails on `master`, passes here);
- `seed_test(gin_rummy_v4.env)` still passes.

`seed_test` + `api_test` pass for all four RLCard envs; `pre-commit` is clean.

### DIFF
--- a/pettingzoo/classic/rlcard_envs/gin_rummy.py
+++ b/pettingzoo/classic/rlcard_envs/gin_rummy.py
@@ -176,6 +176,15 @@ class raw_env(RLCardBase, EzPickle):
         if self.render_mode == "human":
             self.clock = pygame.time.Clock()
 
+    def _seed(self, seed=None):
+        # RLCardBase._seed() rebuilds self.env via rlcard.make(), which
+        # discards the custom payoff scorer installed in __init__. Without
+        # re-applying it here, the configured knock_reward / gin_reward
+        # silently revert to RLCard's defaults (0.2 / 1.0) on every
+        # reset(seed=...). See issue #1312.
+        super()._seed(seed=seed)
+        self.env.game.judge.scorer.get_payoff = self._get_payoff
+
     def _get_payoff(self, player: GinRummyPlayer, game) -> float:
         going_out_action = game.round.going_out_action
         going_out_player_id = game.round.going_out_player_id
--- a/pettingzoo/classic/rlcard_envs/test_gin_rummy.py
+++ b/pettingzoo/classic/rlcard_envs/test_gin_rummy.py
@@ -1,0 +1,66 @@
+"""Tests for the gin_rummy environment.
+
+Regression tests for https://github.com/Farama-Foundation/PettingZoo/issues/1312:
+calling ``reset(seed=...)`` rebuilt the underlying RLCard env and discarded the
+custom payoff scorer, silently reverting the configured ``knock_reward`` /
+``gin_reward`` to RLCard's defaults (0.2 / 1.0).
+"""
+
+from __future__ import annotations
+
+import random
+
+import numpy as np
+
+from pettingzoo.classic import gin_rummy_v4
+from pettingzoo.test import seed_test
+
+# Action ids that end a hand (gin, or any of the 52 knock actions).
+GIN_ACTION = 5
+KNOCK_ACTIONS = range(58, 110)
+
+
+def test_payoff_override_survives_seeded_reset():
+    """Ensure the custom payoff scorer survives a seeded reset."""
+    env = gin_rummy_v4.env(knock_reward=0.5, gin_reward=1.0)
+
+    env.reset()  # no seed: override installed in __init__
+    raw = env.unwrapped
+    assert raw.env.game.judge.scorer.get_payoff == raw._get_payoff
+
+    env.reset(seed=42)  # seeded: env is rebuilt, override must be re-applied
+    assert raw.env.game.judge.scorer.get_payoff == raw._get_payoff
+
+
+def test_configured_knock_reward_used_in_seeded_game():
+    """Ensure a seeded knock pays the configured reward, not RLCard's 0.2."""
+    knock_reward = 0.5
+
+    def finisher_policy(mask, rng):
+        legal = np.flatnonzero(mask).tolist()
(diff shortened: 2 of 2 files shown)


Nikelroid:
Heads up: this PR also includes one unrelated commit removing two unused imports
(`random`, `numpy as np`) in `pettingzoo/test/parallel_test.py`. Those are a
pre-existing flake8 `F401` introduced by #1313, and because the lint hook runs on
all files it was failing CI here even though it's unrelated to the gin_rummy fix.
Happy to split it into a separate PR if you'd prefer to keep this one focused.

jkterry1:
@Nikelroid I already merged those into main before I saw you're message, _however_ if you'd be kind enough to create a separate PR fixing the remaining CI issues on main I'd be very grateful

Nikelroid:
Sure, I will do that in a minute

pettingzoo/test/parallel_test.py
```diff
@@ -1,10 +1,7 @@
 from __future__ import annotations
```
jkterry1, line 2: The removes from this file are already in main now, they can be removed from this PR

Nikelroid:
Done — force-pushed; only the gin_rummy `_seed` commit remains on this branch.

The remaining failing checks (`tutorial-test (CustomEnvironment, …)` across all Python versions) are unrelated to this PR and pre-existing on master: gymnasium's `Discrete.sample(mask=...)` now requires an `np.ndarray`, but the tutorial's `reset()` returns Python lists. I've opened a separate PR for that as you suggested: #1339, which is merged and closed.

jkterry1:
Looking at this again, this would impact learning performance so needs to bump the environment version throughout the repo and docs as well. This was a good catch though, thank you

jkterry1:
Thanks so much @Nikelroid - would you be interested in taking on some more of the open issues for pettingzoo?

jkterry1:
@Nikelroid I accidentally merged this too early, could you please create a follow on Pr that bumps the environment version throughout?

Nikelroid:
> @Nikelroid I accidentally merged this too early, could you please create a follow on Pr that bumps the environment version throughout?

Thanks @jkterry1! I've opened the follow-on that bumps the environment version: gin_rummy_v4 → gin_rummy_v5 across the env module, registration, docstring/docs, and the tests, with a new v5 Version History entry (api_test, seed_test, and the gin_rummy tests all pass under v5). → #1367

Nikelroid:
> Thanks so much @Nikelroid - would you be interested in taking on some more of the open issues for pettingzoo?

Yes, I'd be glad to take on more open issues. If you could point me at a few you'd like prioritized, I'll start picking them up.

PR #948, opened 2023-04-13
Hanabi update

elliottower:
# Description

Splitting https://github.com/Farama-Foundation/PettingZoo/pull/933 into separate PRs for simplicity, see original for comments about changes.

Note: now that the PR has been split, this will fail until https://github.com/Farama-Foundation/PettingZoo/pull/946 is merged, as it fixes edge cases with testing that cause this to fail.

Fixes # (issue), Depends on # (pull request)

## Type of change

> Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

### Screenshots

> Please attach before and after screenshots of the change if applicable.
> To upload images to a PR -- simply drag and drop or copy paste.

# Checklist:

- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [ ] I have run `pytest -v` and no errors are present.
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I solved any possible warnings that `pytest -v` has generated that are related to my code to the best of my knowledge.
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #1355, opened 2026-06-24
Hanabi: finish OpenSpiel/Shimmy migration cleanup

### LINKED ISSUE #1333: Hanabi Upgrade
### Proposal

Right now, Hanabi in PettingZoo depends on https://github.com/Farama-Foundation/hanabi-learning-environment, which is a fork of an unmaintained repo. The hanabi learning environment was upstreamed into OpenSpiel, which in turn already has a generic pettingzoo wrapper via shimmy (https://github.com/Farama-Foundation/Shimmy). We aren't getting bug fixes, new python versions, etc this way.

If this is something a community member could create a PR for, I'd greatly appreciate it.

mfornet:
# Description

The Hanabi environment was migrated off the unmaintained [`hanabi-learning-environment`](https://github.com/Farama-Foundation/hanabi-learning-environment) fork onto OpenSpiel (via the Shimmy `OpenSpielCompatibilityV0` wrapper) in #948, which introduced `hanabi_v5`. That migration left behind dead code and docstrings that still describe the old backend's behavior. This PR finishes that cleanup — no change to game dynamics.

Changes in `pettingzoo/classic/hanabi/hanabi.py`:

- **`step()`**: removed the non-functional `observe=True, as_vector=True` parameters (silently ignored ever since the OpenSpiel switch) and the bogus `Optional[Union[np.ndarray, List[List[dict]]]]` return annotation. `step` now matches the AEC API — `step(self, action) -> None` — which also fixes an incompatible-override type mismatch against `AECEnv.step`.
- **`observe()`**: renamed `agent_name` → `agent` to match `AECEnv.observe(self, agent)` (was another override mismatch).
- Removed stale `# ToDo: Starts` and `# ToDo: Fix Return value` comments.
- Removed a duplicate `self.rewards = self.hanabi_env.rewards` assignment in `reset()`.
- Rewrote the `reset()`/`step()` docstrings, which described return values the methods no longer produce.
- Fixed the `Hanabi-Full` docstring config example, which had contradictory duplicate `hand_size` keys (`(4 if players >= 4 else 5)` plus a stray `"hand_size": 2`).
- Dropped the now-unused `Union` import.

Fixes #1333

## Type of change

- This change requires a documentation update (the env docs page is generated from the module docstring, so the docstring fixes propagate automatically)

> Note: `step()` and `observe()` signatures changed. The removed `step` kwargs were already inert (ignored since the v5 backend swap), so this is not a behavioral break, but a caller passing `step(action, observe=..., as_vector=...)` by keyword would now raise `TypeError`.

# Checklist:

- [x] I have run the `pre-commit` checks (black, isort, ruff, pydocstyle all pass on the changed file)
- [x] I have run `pytest -v` and no errors are present (verified via `api_test` and `seed_test` across default, 3/4/5-player, seer, minimal, and small-game configs)
- [x] I have made corresponding changes to the documentation (docstring is the docs source)
- [ ] I have added tests that prove my fix is effective or that my feature works (existing `hanabi_v5` tests cover the change; no new behavior added)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pettingzoo/classic/hanabi/hanabi.py
+++ b/pettingzoo/classic/hanabi/hanabi.py
@@ -161,7 +161,7 @@ If an illegal action is taken, the game terminates and the one player that took 
 
 """
 
-from typing import List, Optional, Union
+from typing import List, Optional
 
 import gymnasium
 import numpy as np
@@ -243,9 +243,8 @@ class raw_env(AECEnv, EzPickle):
                 "players": 2,
                 "max_information_tokens": 8,
                 "max_life_tokens": 3,
-                "hand_size": (4 if players >= 4 else 5)
+                "hand_size": 5,
                 "observation_type": "card_knowledge",
-                "hand_size": 2,
                 }
 
             Hanabi-Small : {
@@ -282,7 +281,6 @@ class raw_env(AECEnv, EzPickle):
             render_mode=render_mode,
         )
 
-        # ToDo: Starts
         # Check if all possible dictionary values are within a certain ranges.
         self._raise_error_if_config_values_out_of_range(
             colors,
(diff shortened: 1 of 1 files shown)


jkterry1:
Right so, in an internal issue tracking spreadsheet (which I'm in the process of moving entirely to github issues anyone can see), the fact that that happened 2 years ago was never done, and the hanabi learning environment readme wasn't updated like it should've been, so I was out outright confused and didn't realize that had been fixed, so that's highly clarifying thank you.

That being said on review of the changes to the PR file, they don't entirely make sense to me?

mfornet:
sorry, what doesn't make sense to you, I'm checking it now but can't figure out what do you mean. Also feel free to close it, as this is more a follow up, than really needed.

PR #1380, opened 2026-07-04
Add pygame rendering to Hanabi

### LINKED ISSUE #1379: Hanabi Rendering
If someone would be kind enough to create a PR that adds rendering using them, I commissioned a pixel artist to create all the cards needed for adding rendering to Hanabi.

[DELIVERY.zip](https://github.com/user-attachments/files/29581879/DELIVERY.zip)

RewardGuard:
Closes #1379.

Adds graphical rendering to Hanabi using the commissioned pixel-art card set, and removes the old console/`ansi` rendering it replaces. Previously the env only supported text output via the underlying OpenSpiel state.

The board is parsed from the OpenSpiel state string (`str(game_state)`) and drawn with pygame: the fireworks piles, every player's hand (the player to move is highlighted), the remaining info and life tokens, the deck size, and the discard pile.

**Render modes**
- `human` — draws the board to a pygame window
- `rgb_array` — returns the board as an `(H, W, 3)` array

The previous `ansi` render mode (which just returned the OpenSpiel text representation of the state) is removed, since the pygame board supersedes it. It is dropped from `metadata["render_modes"]` (now `["human", "rgb_array"]`), from `render()`, and from the docs; the `v5` version-history note is updated to say the pygame rendering replaces the `ansi` text rendering.

**Notes**
- State parsing lives in `rendering.py`, isolated from pygame so it can be unit-tested without a rendering backend. The parser targets the Hanabi Learning Environment `ToString` format (colours `RYGWB`, ranks `12345`).
- Layout was verified across 2–5 players by rendering synthetic states to PNG. I could not run the OpenSpiel backend locally (it needs Python ≥ 3.11, and the PyPI `open_spiel` wheel does not bundle the Hanabi game), so the state-string parser was validated against the documented `ToString` format and synthetic games; CI exercises the modes end-to-end. Happy to adjust if a reviewer with the backend spots a mismatch.


jkterry1:
@RewardGuard tests are failing

jkterry1:
Also @RewardGuard this needs to add a gif of the new rendering to docs

RewardGuard:
ok on it

jkterry1:
@RewardGuard can you please update this to also remove the old ansii rendering logic?

jkterry1:
@RewardGuard I apologize for the hassle, but could you please rebase your changes off main?

PR #1389, opened 2026-07-07
chore: update dependencies in pyproject.toml

Trenza1ore:
# Description

Same changes as `Gymnasium`:
- Use `box2d` on Python 3.10 - 3.13 (has prebuilt wheel, no source distribution); `box2d-py` on Python 3.14 (has source distribution).
- Use [self-reference](https://pip.pypa.io/en/latest/reference/requirement-specifiers/#self-referential-extras) for `all` extra.

## Type of change

# Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,14 @@ classic = [
     "open_spiel>=1.6.15; python_version >= '3.11'"
 ]
 butterfly = ["pygame-ce>=2.3.0", "pymunk>=6.2.0"]
-sisl = ["pygame-ce>=2.3.0", "pymunk>=6.2.0", "box2d-py>=2.3.5", "scipy>=1.4.1"]
+sisl = [
+    "pygame-ce>=2.3.0",
+    "pymunk>=6.2.0",
+    "scipy>=1.4.1",
+    "box2d==2.3.10; python_version < '3.14'",
+    "box2d-py==2.3.8; python_version >= '3.14'",
+    "swig==4.*; python_version >= '3.14'",
+]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -54,16 +61,7 @@ testing = [
...
-    "pygame-ce>=2.3.0",
-    "chess>=1.9.4",
-    "rlcard>=1.0.5",
-    "shimmy>=1.2.0",
-    "open_spiel>=1.6.15; python_version >= '3.11'",
-    "pymunk>=6.2.0",
-    "box2d-py>=2.3.5",
-    "scipy>=1.4.1",
-    "pillow>=8.0.1",
+    "pettingzoo[atari, classic, butterfly, sisl, other]"
```
jkterry1, line 64: @Trenza1ore are you sure this is valid syntax? If so I believe you, but it's newer than last time I deeply learned this syntax

Trenza1ore: Yeah, the same change has been applied when fixing Gymnasium's CI failure as well. It's not officially defined by any PEP (only ever mentioned in [PEP 771](https://peps.python.org/pep-0771/#supporting-extras-that-should-not-remove-default-extras) as a feature that has been supported by packaging tools)

It has been supported since pip 21.2 (2021-07-24):
- https://github.com/pypa/pip/issues/10393
- https://discuss.python.org/t/pyproject-toml-optional-dependencies-redundancy-aka-dry-extras/8428/6

This feature is still awaiting documentation:
- https://github.com/pypa/pip/issues/11296

PR #1390, opened 2026-07-08
fix(registry): normalize env id to use -v[version] & misc improvements to code quality

Trenza1ore:
# Description

PettingZoo registry implementation in #1388 did not normalize version parsing (since it seems that we only provide one version per environment within PettingZoo), so for example: `atari/basketball_pong_v3` is treated as an environment named `basketball_pong_v3` in the `atari` namespace with no versioning, instead of the `basketball_pong` environment's version `3`.

This PR does these:
- Proper version string normalization (`atari/basketball_pong_v3` to `atari/basketball_pong-v3`) so that the existing env id parsing logic can take advantage of this
- Made `EnvSpec` dataclass `frozen=True, slots=True`, as it should be immutable by design
- Changed a few variable names that are not informative enough
- Made the `@lru_cache` decorators untyped; no user should pass in some custom object mimicking a `str` anyway, and untyped is faster

## Type of change

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ or with the new Gymnasium-like registry:
 
 ```python
 from pettingzoo import make
-env = make("aec", "butterfly/pistonball_v6")
+env = make("aec", "butterfly/pistonball-v6")
 ```
 
 Environments can be interacted with in a manner very similar to Gymnasium:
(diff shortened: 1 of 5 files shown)


Trenza1ore:
I'm merging this so that documentation process for PZ registry may start. The changes here are either beneficial or benign, so should not cause any issue.

PR #1397, opened 2026-07-12
feat!: remove gin_rummy environment

### LINKED ISSUE #1383: Removing Unused Environments
Before we cut the next release, I think we should remove:

- TicTacToe (it's virtually unused - https://github.com/search?q=pettingzoo+tictactoe_v3&type=code - everyone seems to use rock paper scissors or connect four instead)
- Gin Rummy (based on the same source of metrics its essentially unused _and_ has no deep history with AI like some of the other classic games)

teddytennant:
Closes #1383 (gin rummy portion — tic-tac-toe is kept per the discussion above).

## Description

Removes the Gin Rummy environment: the env module, its `gin_rummy_v5` entry point, test, docs page/gif, and its entries in `all_modules`, the env registry, `api_test`'s env lists, the parameter-comb / pygame-init / pickle / unwrapped tests, and the SB3 action-mask tutorial.

The shared `rlcard_envs` card images and font stay — Leduc and both Texas Hold'em variants use them.

## One thing worth flagging

Deleting the module alone leaves anyone with `gin_rummy_v5` in their code hitting a **bare `AssertionError` with an empty message**. `deprecated_handler()` only has two paths for a missing module — "a newer version exists" and "not an env name" — and a removed env matches neither, so it falls out of the loop into `assert spec`:

```python
>>> from pettingzoo.classic import gin_rummy_v5
AssertionError:
```

So this PR adds a `REMOVED_ENVS` map, and removed envs now raise a `RemovedEnv(AttributeError)` pointing at RLCard, which still has the implementation:

```python
>>> from pettingzoo.classic import gin_rummy_v5
RemovedEnv: gin_rummy was removed from PettingZoo, see .../issues/1383. The underlying
implementation remains available in RLCard: https://github.com/datamllab/rlcard
```

It also makes the unknown-env fall-through raise a proper `AttributeError` rather than tripping the assert — that was reachable before this PR (e.g. `pettingzoo.classic.knights_archers_zombies_v11`). `RemovedEnv` subclasses `AttributeError` so `getattr(obj, name, default)` keeps working, per the existing comment in that file.

Happy to drop this and do a pure deletion if you'd rather keep the PR minimal.

## Testing

`test/deprecated_module_test.py` covers the removed-env message, the unknown-env `AttributeError`, the `getattr` default, and that superseded versions (`tictactoe_v2`) keep their existing deprecation behavior.

Full suite: no newly-failing tests vs. `main`. The only tests that disappear are the five gin_rummy parametrizations.

## Type of change

# Checklist:


jkterry1:
@teddytennant I apologize for the hassle, but could you please rebase your changes off main?

Issue #1332, opened 2026-05-24
[Bug Report] Possible Tic Tac Toe Logical Error


jkterry1:
### Describe the bug

See https://github.com/Farama-Foundation/PettingZoo/pull/1229

### Code example

```shell

```

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/PettingZoo/issues) in the repo

jkterry1:
@teddytennant do you feel like looking into this further?

teddytennant:
Had a look. Short version: #1229's bug doesn't exist, but there's a real one next to it.

**#1229 doesn't reproduce.** `reset()` restores `self.agents` from `possible_agents` *before* the reinit, so the selector always gets the full list back:

```python
self.agents = self.possible_agents[:]
...
self._agent_selector.reinit(self.agents)
```

That ordering is also present at `67ecdb2`, the commit #1229 was branched from, so it was never broken. I ran their exact scenario (illegal move terminates both, dead-step only `player_1` so `env.agents == ["player_2"]`, then `reset()`): both agents come back, `player_1` moves first, marks are correct. Matches what @benblack769 said on the PR in 2024. No env change, no version bump.

**The real issue:** `AgentSelector.reinit()` stores the order by reference (`self.agent_order = agent_order`), so `env._agent_selector.agent_order is env.agents` for tictactoe, chess, go, connect_four, rps, hanabi. `AECEnv._was_dead_step()` then does `self.agents.remove(agent)`, mutating the selector's cycle in place from the outside. Benign in the 2-player classics (game's already over), but the selector ends every episode with `agent_order == []` and `next()` raises `ZeroDivisionError`. It's a footgun for any env where agents die while play continues.

**The obvious fix breaks things.** Making `reinit()` copy takes the suite from 61 passed to 3 failed (`test_action_mask[env0]`, `test_generated_agents_aec`) — the `generated_agents` example envs *rely* on the alias: `add_agent()` appends to `self.agents` and never re-inits the selector.

So, two options, happy to PR either:

- **(a)** Document the live reference as intentional, add a regression test pinning the #1229 case. Zero risk.
- **(b)** Make `reinit()` copy and give `generated_agents` an explicit add path. Cleaner, but a breaking change for third-party env authors.

Leaning (b) with (a)'s test folded in, but (a) is the safe close. Either way I think #1229 was a false positive.

jkterry1:
Lets take a look at what a PR for B looks like

PR #1400, opened 2026-07-13
fix: give AgentSelector ownership of its agent order

teddytennant:
This is option **B** from #1332, as requested. It also folds in option A's regression test for the #1229 case.

## The problem

`AgentSelector.reinit()` stored the caller's list *by reference*:

```python
def reinit(self, agent_order):
    self.agent_order = agent_order   # not a copy
```

So for every env that passes `self.agents` — tictactoe, chess, go, connect_four, rps, hanabi, pistonball, pursuit, multiwalker — `env._agent_selector.agent_order` **is** `env.agents`, the same list object. Two pieces of core behavior were quietly riding on that aliasing:

- `AECEnv._was_dead_step()` shrinks the cycle via `self.agents.remove(agent)`
- the `generated_agents` example envs grow it via `self.agents.append(agent)` in `add_agent()`

Neither ever names the selector, so the coupling is invisible at both ends. It also leaves the selector with `agent_order == []` at the end of an episode, where `next()` raises `ZeroDivisionError: integer modulo by zero`.

## The change

`reinit()` now copies, and both mutations become explicit:

- `AgentSelector.add_agent()` / `AgentSelector.remove_agent()`
- `AECEnv._was_dead_step()` calls `remove_agent()` — this is where envs previously got removal for free
- the four `generated_agents` example envs call `add_agent()`

`remove_agent()` is a no-op for an agent that isn't in the cycle, so an env that already dropped the agent itself (KAZ maintains its own `_live_agents` copy) can call it unconditionally.

## It surfaced a real bug in cooperative_pong

`cooperative_pong` builds its selector once in `__init__` and never refreshes it on `reset()`. Under aliasing that was masked: `reset()` rebinds `self.agents` to a fresh list, so the selector's original list was never the one `_was_dead_step()` drained, and it happened to hold the right agents forever. With the selector owning its order, that stale list *does* get drained, and the env dies on the second episode. Fixed by re-initing against the fresh agent list on reset, which is what every other env already does.

I'd argue that's the point of the change — the aliasing was hiding it.

## Breaking change

A third-party AEC env with a **dynamic agent set** that adds agents by mutating `self.agents` and expects the selector to notice will need to call `add_agent()`. Removal is unaffected, since `_was_dead_step()` handles it. Envs with a fixed agent set (i.e. nearly all of them) need no changes.

## Testing

New `test/agent_selector_test.py`, 10 tests: that `reinit()` copies, `add_agent`/`remove_agent`, the no-op removal, that `_was_dead_step()` keeps the cycle in sync, that agents added mid-episode get selected, that the order doesn't leak across resets, and a full dynamic episode staying consistent.

It includes the **#1229 regression test**: reset while an agent is still missing from `.agents` restores the whole cycle. (As noted in #1332, that case already worked — this pins it so it doesn't get re-reported.)

Full suite: no newly-failing tests vs `main`.

## Type of change

# Checklist:


jkterry1:
@teddytennant your PR is failing CI

teddytennant:
Fixed the AgileRL CI failure.

**Root cause:** `_was_dead_step()` now correctly drops agents from the selector's cycle (the whole point of this PR). Envs like **mpe2** only call `_agent_selector.reset()` on env reset and never `reinit`. `reset()` was implemented as `reinit(self.agent_order)`, so after a full episode drained the cycle it rewound an empty list and the next `next()` raised `ZeroDivisionError: integer modulo by zero`. AgileRL's `AsyncPettingZooVecEnv` worker hit that on the second episode; the real traceback was then masked by a gymnasium logger format-string bug (`TypeError: not enough arguments for format string`).

**Fix:** `AgentSelector` now remembers the order last passed to `reinit()` / the constructor and restores it on `reset()`. `add_agent` / `remove_agent` stay episode-local. The `generated_agents` example envs start the cycle with `next()` after building via `add_agent`, so `reset()` no longer wipes their setup.

New unit tests cover restore-after-removals, drop-episode-local-additions, and reinit-updates-baseline. mpe2 multi-episode (AEC + parallel) and the existing agent_selector suite pass locally.

pettingzoo/utils/agent_selector.py
```diff
@@ -25,20 +36,55 @@ class AgentSelector:
...
+        Episode-local: :meth:`reset` restores the agent if it was in the order
+        last passed to :meth:`reinit`.
+        """
+        if agent in self.agent_order:
+            self.agent_order.remove(agent)
+
     def reset(self) -> Any:
-        """Reset to the original order."""
-        self.reinit(self.agent_order)
+        """Reset to the order last passed to :meth:`reinit`.
```
dm-ackerman, line 89: This should document the return value - it's (maybe) not obvious that reset steps through the agent list

pettingzoo/utils/agent_selector.py
```diff
@@ -25,20 +36,55 @@ class AgentSelector:
...
+
+        Mid-episode :meth:`add_agent` / :meth:`remove_agent` mutations are
+        undone, so a second env episode can start cleanly even if the previous
+        one dead-stepped every agent out of the cycle.
+        """
+        self.agent_order = list(self._original_agent_order)
+        self._current_agent = 0
+        self.selected_agent = 0
```
dm-ackerman: I believe this can be implemented as self.reinit(self._original_agent_order)  for less duplication at a slight performance cost

pettingzoo/utils/env.py
```diff
@@ -228,6 +228,12 @@ def step(self, action):
         del self.infos[agent]
         self.agents.remove(agent)
 
+        # Envs used to get this for free, because the selector aliased the very
+        # list we just mutated. It owns a copy now, so drop the agent explicitly.
+        agent_selector = getattr(self, "_agent_selector", None)
+        if agent_selector is not None:
+            agent_selector.remove_agent(agent)
+
```
dm-ackerman, line 269: This adds an undocumented expectation that environments have defined their agent selector as _agent_selector. If they haven't, this will potentially yield bugs. 
i think this needs to be handled somehow

pettingzoo/utils/agent_selector.py
```diff
@@ -25,20 +36,55 @@ class AgentSelector:
...
+        The order is copied, so later mutations of ``agent_order`` by the caller
+        do not silently change the cycle. The copy is also remembered as the
+        baseline that :meth:`reset` restores.
+        """
+        self.agent_order = list(agent_order)
+        self._original_agent_order = list(agent_order)
         self._current_agent = 0
         self.selected_agent = 0
 
+    def add_agent(self, agent: Any) -> None:
```
dm-ackerman, line 59: weird things can happen if there are duplicate agents added (due to removal and is first/last checks). There was never a check for that before, I don't know if it makes sense to add one now?

pettingzoo/utils/agent_selector.py
```diff
@@ -25,20 +36,55 @@ class AgentSelector:
...
+        Episode-local: :meth:`reset` will drop the agent again unless
+        :meth:`reinit` is called with it included.
+        """
+        self.agent_order.append(agent)
+
+    def remove_agent(self, agent: Any) -> None:
+        """Remove an agent from the cycle.
+
+        Does nothing if the agent is not in the cycle, so that an env which has
+        already dropped the agent itself can still call this unconditionally.
```
dm-ackerman: I'm not sure having remove_agent silently ignore the agent not being present is the best option. It's not in line with existing remove() behavior in the python library which raises errors. It's not clear to me that there is a need for supporting duplicate calls to remove_agent(). What is the use case for supporting that?

dm-ackerman (commented):
a few minor-ish things - a couple are just things to think about, but assuming the naming of the selector really ought to be addressed

dm-ackerman:
I made a couple comments on the code - mostly minor things and at least one more than minor thing

However, thinking more on this, I don't think this should be accepted. This is a breaking change that is likely to be overlooked by most users. We're probably not going to tag this version 2.0.0 so we're breaking other peoples code without telling them, which is not lining up with best practices or the project standards (https://farama.org/project_standards)

As for breaking changes, `agents` is tied to the agent selector (for good or bad) and the code assumes this. See for example AECIterator, agents and agent_selection are implicitly coupled. Breaking that coupling is going to enable subtle bugs that the AI missed, like the iterator breaking. This isn't just an AgentSelector design change, it is a codebase design change and needs more thought put into it to avoid breaking other stuff.

I'm not claiming this is a bad change, just premature. If this is the way to go, it makes more sense in PZ 2.0 than right now.

Is there an actual bug that this is fixing? If not, it seems questionable to knowingly break a bunch of existing environments and add potential bugs for something that is not needed.

teddytennant:
Thanks, both good points. I pushed fixes for the three code comments:

- `add_agent` now raises if the agent is already in the cycle. You were right that duplicates cause trouble: `remove_agent` only drops the first copy and `is_first`/`is_last` compare by value.
- `remove_agent` now raises like `list.remove` does. The only caller that can't know whether the agent is still there is `_was_dead_step`, so it checks membership itself now.
- For the `_agent_selector` naming: `api_test` now fails an env whose selector still lists an agent after that agent's final `step(None)`. It looks for selectors by type, not by name, so an env that stores one under a different name gets a clear failure instead of a subtle bug. I also documented the convention on `_was_dead_step`.

On whether this should land at all, to answer your question directly: yes, there are two real bugs.

`next()` raises `ZeroDivisionError: integer modulo by zero` once every agent has been dead stepped out, because the selector's order is the same list `_was_dead_step` drains. And `cooperative_pong` builds its selector once in `__init__` and never refreshes it on reset, which the aliasing was hiding. It dies on the second episode once the selector owns its order.

Worth noting this is option B from #1332, which was asked for there. If the breakage still feels too early for a 1.x release I'm happy to split it: the `ZeroDivisionError` fix, the `cooperative_pong` fix and the regression tests can go in on their own, and the ownership change can wait for 2.0. Just say which you'd prefer.

jkterry1:
@dm-ackerman further takes on this PR?

teddytennant:
rebased on main. kept both assertions after a dead step: the agent is gone from `env.agents` (the check that landed while this was open) and from any `AgentSelector` the env still holds.

PR #1414, opened 2026-08-07
docs: clarify when env.agents becomes empty after termination (#1244)

### LINKED ISSUE #1244: [Bug Report] Checking if the entire environment is done
### Describe the bug

According to the documentation:

"When an agent is terminated or truncated, it’s **removed** from `agents`, so when the environments done `agents` will be an **empty list**. This means not `env.agents` is a simple condition for the environment being done." - [PettingZoo Doc](https://pettingzoo.farama.org/content/basic_usage/)

I tested this on `connect_four_v3` and it does not work:

The `env.agents` will still have agents in the list even after all agents have terminated.  This can be checked by conditioning a break point on all terminated agents.

### Code example

```shell
import numpy as np
import pettingzoo
from pettingzoo.classic import connect_four_v3


def test_connect4():
    obs: dict[str, np.ndarray]
    termination: bool
    truncation: bool

    env: pettingzoo.AECEnv = connect_four_v3.env(render_mode="human")
    env.reset(seed=42)

        obs, _, termination, truncation, _ = env.last()  # type: ignore

        if termination or truncation:
            if len(env.agents) > 0:
                print(f"env.agents: {env.agents}")
                raise ValueError("env.agents should be empty!")

        mask = obs["action_mask"]
        action = env.action_space(agent).sample(mask)

        env.step(action)

    env.close()
```

### System info

Running on Intel based Macbook Pro 2019.

### Additional context

_No response_

### Checklist

teddytennant:
## Problem

Issue #1244 reports that docs say `not env.agents` means the environment is done, but on `connect_four_v3` (and other AEC envs) `env.agents` still contains agents when `last()` first returns `termination=True`.

## Root cause

This is not a connect_four bug. AEC environments are expected to give a dying agent one final vacuous `step(None)` before retiring it, so that `last()` can report its terminal observation/reward/flag. `_was_dead_step` performs the retirement, and only once every dead agent has been stepped off is `env.agents` empty.

This is a convention environments implement, not behaviour the base `AECEnv` provides — the base class never touches `agents`. The docs under "Checking if the entire environment is done" omitted the timing entirely, so users who checked `not env.agents` immediately on termination thought the env was broken.

## Fix

Docs:
- Clarify the removal timing in `docs/content/basic_usage.md`, with a connect_four example matching the recommended loop, and state that it is a convention environments follow rather than a base-class guarantee.
- Align the "Variable Numbers of Agents (Death)" section and the AEC API `agents` attribute docs.
- Document the dead step for environment authors: why the extra vacuous step exists, the recommended `step()` snippet, and what breaks if it is skipped.
- Render `_was_dead_step` on the AEC API page and expand its docstring with the rationale (the previous example also called it without the required `action`).
- Note the convention on `AECEnv.agents` in `pettingzoo/utils/env.py`.

Tests:
- `api_test` now checks both halves of the convention explicitly: a terminated/truncated agent is still in `env.agents` when `last()` reports it done, and is gone after its `step(None)`. The first half was already implied by the existing `live_agents` subset check.
- Add regression tests that pin the timing for connect_four, tictactoe, and rps, including the exact mistaken check from the issue.

No environment behavior change.

## Test plan

- [x] `pre-commit run --all-files`
- [x] `pytest test/agents_done_test.py test/doc_examples_test.py test/variable_env_test.py test/wrapper_test.py` (23 passed)
- [x] `pytest docs/content/basic_usage.md docs/api/aec.md docs/content/environment_creation.md --markdown-docs -m markdown-docs` (13 passed)
- [x] `sphinx-build -b dirhtml docs _build` clean for the changed files
- [x] `api_test` over classic + butterfly + the `generated_agents` example envs with the new assertions

Fixes #1244


dm-ackerman:
This is a good clarification of a confusing point. It should be noted that the description here is only true for environments that explicitly implement this behaviour. It is not universal AEC behaviour. That should be clarified as the comments to env.py are not correct in general - only when the env author implements it that way. 

I'm not sure if the dead step details are in the docs in enough detail for env designers to use but, if not, they should be and that would be a valuable addition. I think there is a recommendation to include it in step() but no rationale given. 

If this is the expected AEC behaviour, a test for this would be a good thing to add to the AEC api test (if it isn't there already)

Edit: I hope this comment didn't come across as negative. This change is tackling one of the weaker areas of AEC that has needed improvement for some time IMO. I just want to point out other areas connected to this that need work.

teddytennant:
Not negative at all, that was a fair catch. You're right that this isn't universal AEC behaviour, only what envs that implement it get.

Reworded `docs/api/aec.md`, `basic_usage.md` and the `env.py` comment to describe it as a convention envs are expected to follow rather than something the base class provides. I also filled in the dead step rationale for env authors, since you mentioned it wasn't documented well: why the extra vacuous step exists, what `_was_dead_step` does, and what breaks without it.

On the test, `play_test` already required a dead agent to still be listed via the `live_agents` subset check, so I made that explicit and added the missing other half, that the agent is gone after its `step(None)`.

jkterry1:
@dm-ackerman are you happy with the current phrasing?

PR #1415, opened 2026-08-08
Add FlattenObservation wrappers, ported from SuperSuit flatten_v0

Amala-Bharadwaj:
# Description

Ports SuperSuit's `flatten_v0` into PettingZoo natively, as part of #1365. Flattens each agent's observation into a 1D array and rewrites the advertised observation space to match.

Exposed as classes rather than a factory function, per @virgilt's note in #1365 about mirroring `gymnasium.wrappers`, and named after the Gymnasium equivalent. AEC and Parallel are implemented natively rather than going through SuperSuit's conversion machinery — so `FlattenObservationParallel` works on a Parallel env directly, without the convert-to-AEC-and-back dance the docs currently warn about.

I avoided `clip_actions_v0` (already covered by `ClipOutOfBoundsWrapper`) and `clip_reward_v0` / `color_reduction_v0` (covered by #1091).

**Question on placement:** I've put this in `pettingzoo/utils/wrappers/` alongside the existing native wrappers, and added both classes to `docs/api/wrappers/pz_wrappers.md`. #1365 says `pettingzoo.wrappers` and #1091 used `pettingzoo/utils/wrappers/supersuit/`, so there are three possible answers here — happy to move it, it's just a file move.

Fixes # (part of #1365)

## Type of change

- New feature (non-breaking change which adds functionality)

# Checklist:

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files`
- [x] I have run `pytest -v` and no errors are present — see note below
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I solved any possible warnings that `pytest -v` has generated that are related to my code to the best of my knowledge.
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

On `pytest -v`: I ran the suite excluding `pettingzoo/atari`, since `multi_agent_ale_py` doesn't build on Windows. 138 passed. The two `test/wrapper_record_video_test.py` failures reproduce identically on master with my changes reverted, so they're unrelated to this PR. `test/flatten_test.py` is 17/17.

Issue #1365, opened 2026-06-24
Migration to pettingzoo.wrappers


jkterry1:
Right now, pettingzoo relies on supersuit for environment wrappers. We want to refactor this into pettingzoo.wrappers, similar to gymnasium.wrappers, and kill off supersuit. Community PRs to begin porting the wrappers over (one PR per wrapper please) would be greatly appreciated.

jkterry1:
See https://github.com/Farama-Foundation/PettingZoo/pull/1091 and https://github.com/Farama-Foundation/PettingZoo/pull/1180

jkterry1:
@teddytennant do you want to start taking pot shots at these one wrapper at a time?

jkterry1:
It'd be a huge help :)

jkterry1:
To start could you figure out what wrappers we'd want to migrate over?

fraware:
I can take `clip_reward_v0` as one isolated migration, with native AEC and Parallel implementations and focused cumulative-reward tests. Is it still unclaimed, and should the public API be separate wrapper classes or retain the current factory function?

virgilt:
@fraware clip_reward_v0 is still unclaimed, so feel free to take it.

Since the goal is to mirror gymnasium.wrappers, we should follow the same pattern and expose wrapper classes.

akki-g:
Hi! I’d like to help with this isssue. Could I take `delay_observations_v0` as an isolated migration?

jkterry1:
Sure

ilovehhhyn:
Hi! Could I take frame_stack_v1 as an isolated migration? I’d implement native FrameStackObservation and FrameStackObservationParallel classes with independent per-agent histories, documentation, and focused AEC/Parallel tests. Before starting, should the new classes preserve SuperSuit’s stack_dim and zero-padding behavior, or follow Gymnasium’s newer FrameStackObservation shape and padding_type semantics?

quinnarnold:
I'd like to take `agent_indicator_v0` as one isolated migration, if it's still unclaimed.

I propose native `AgentIndicator` and `AgentIndicatorParallel` classes under `pettingzoo.utils.wrappers`, preserving SuperSuit's currently supported inputs and indicator semantics: homogeneous `Discrete` or 1D/2D/channels-last 3D `Box` observations, including `type_only=True`.

[...]

PR #1426, opened 2026-08-25
Add DtypeObservationV1 wrappers, ported from SuperSuit dtype_v0

teddytennant:
# Description

Ports SuperSuit's `dtype_v0` as `DtypeObservationV1` and `DtypeObservationParallelV1`, part of #1365. Recasts each agent's observation and rewrites the advertised observation space to the new dtype.

Classes rather than a factory function, per @virgilt's note in the issue about mirroring `gymnasium.wrappers`, and named after the Gymnasium equivalent. AEC and Parallel are both implemented directly on the PettingZoo wrapper bases, so the Parallel class works on a Parallel env without converting to AEC and back.

Same shape as #1415, which is the other wrapper port currently open.

## One thing worth knowing

The space conversion follows SuperSuit's `convert_box`: cast the bounds, then hand them to `Box`. `gymnasium.wrappers.DtypeObservationV1` hands the original bounds to `Box` and lets it validate. The two disagree on finite overflow, and I kept SuperSuit's behaviour so migrating code doesn't change under people:

    Box(0.0, 300.0, float32) -> uint8
      here and in SuperSuit:  Box(0, 44, uint8)      # 300 % 256
      gymnasium:              ValueError, high is out of bounds for the dtype

They agree on infinite bounds (both collapse to a degenerate box, neither raises), and both raise when a downcast flips low above high. `test_downcast_wraps_bounds_like_supersuit` pins this with literals so it's obvious if anyone changes it.

One related trap that has a test and a docstring line now: an unbounded Box cast to an integer dtype collapses to a single point, and `api_test` then fails with an out of bounds observation. That's true of SuperSuit too, but unbounded spaces are common in PettingZoo so it's worth saying out loud rather than leaving people to find it.

Box only, matching SuperSuit. Gymnasium's version also takes Discrete, MultiDiscrete and MultiBinary, so anyone coming from that one hits the assert.

## Verification

    $ pytest test/dtype_test.py -q
    23 passed, 5 warnings in 0.12s

    13 passed, 18 warnings in 22.93s

I checked the port against SuperSuit's actual `change_obs_space` and `change_observation` rather than against my reading of them, on uint8, float32, float64, int8, int32 and unbounded spaces. No divergences.

## A question

`observation_space(agent)` returns the cast space, but `observation_spaces[agent]` still returns the original, because neither base wrapper overrides the dict and `__getattr__` forwards it to the inner env. This PR and #1415 are the first PettingZoo wrappers that change an observation space, so it's the first time the two can disagree. Worth fixing once in `BaseWrapper` and `BaseParallelWrapper` rather than in each wrapper? Happy to do it in a separate PR if you want it, I just didn't want to change the base classes inside a wrapper PR.

## Type of change

# Checklist:

### DIFF
--- a/docs/api/wrappers/pz_wrappers.md
+++ b/docs/api/wrappers/pz_wrappers.md
@@ -129,5 +129,7 @@ while parallel_env.agents:
 .. autoclass:: AssertOutOfBoundsWrapper
 .. autoclass:: ClipOutOfBoundsWrapper
 .. autoclass:: OrderEnforcingWrapper
+.. autoclass:: DtypeObservation
+.. autoclass:: DtypeObservationParallel
 
 ```
--- a/docs/api/wrappers/pz_wrappers.md
+++ b/docs/api/wrappers/pz_wrappers.md
@@ -129,7 +129,7 @@ while parallel_env.agents:
 .. autoclass:: AssertOutOfBoundsWrapper
 .. autoclass:: ClipOutOfBoundsWrapper
 .. autoclass:: OrderEnforcingWrapper
-.. autoclass:: DtypeObservation
-.. autoclass:: DtypeObservationParallel
+.. autoclass:: DtypeObservationV1
+.. autoclass:: DtypeObservationParallelV1
 
 ```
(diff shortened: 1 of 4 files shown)


jkterry1:

teddytennant:
Renamed to DtypeObservationV1 and DtypeObservationParallelV1, SuperSuit's dtype_v0 plus one.

PR #1427, opened 2026-08-25
Add ReshapeObservationV1 wrappers, ported from SuperSuit reshape_v0

teddytennant:
# Description

Ports SuperSuit's `reshape_v0` as `ReshapeObservationV1` and `ReshapeObservationParallelV1`, part of #1365. Reshapes each agent's observation and reshapes the advertised Box the same way, so per-element `low` and `high` bounds survive instead of collapsing to a scalar range.

## Two things worth knowing

`-1` in the shape is rejected with its own message telling you to give the size of every axis. `gymnasium.wrappers.ReshapeObservationV1` nominally allows it, but its own `np.prod(shape) == np.prod(obs_space.shape)` assert runs first and a shape containing -1 has a negative product, so the allowance is dead code there. Rejecting it with a message that says what to do seemed better than letting it fail with a confusing element-count error. Say the word if you'd rather match Gymnasium's signature exactly.

Shapes are validated in `__init__` for every agent in `possible_agents`, and again on first use for anything that turns up later. The second check matters more than it looks: `generated_agents_env_v0` has no `possible_agents` before `reset()`, so a construction-time-only check silently passes and the user gets a bare numpy `ValueError` several steps in with no mention of the wrapper or the agent. That's a deviation from #1415, which is purely lazy.

Empty shape `()` is rejected. SuperSuit accepted it vacuously and gave you a 0-d observation.

## Verification

    $ pytest test/reshape_test.py -q
    33 passed, 3 warnings in 0.14s

    13 passed, 18 warnings in 23.58s

I compared the derived space against SuperSuit's `convert_box` on float32 with per-element bounds, float64, uint8, int8, and int64 and float32 with infinite bounds. Byte-identical low, high, shape and dtype in every case.

## A question

The wrapper takes one shape for every agent, so it only works when all agents' observations have the same element count. SuperSuit had the same limit. Would you rather it took an optional per-agent mapping, or is one shape the right scope?

## Type of change

# Checklist:

### DIFF
--- a/docs/api/wrappers/pz_wrappers.md
+++ b/docs/api/wrappers/pz_wrappers.md
@@ -140,5 +140,7 @@ while parallel_env.agents:
 .. autoclass:: DtypeObservationParallelV1
 .. autoclass:: PadObservationsV1
 .. autoclass:: PadObservationsParallelV1
+.. autoclass:: ReshapeObservation
+.. autoclass:: ReshapeObservationParallel
 
 ```
--- a/docs/api/wrappers/pz_wrappers.md
+++ b/docs/api/wrappers/pz_wrappers.md
@@ -140,7 +140,7 @@ while parallel_env.agents:
 .. autoclass:: DtypeObservationParallelV1
 .. autoclass:: PadObservationsV1
 .. autoclass:: PadObservationsParallelV1
-.. autoclass:: ReshapeObservation
-.. autoclass:: ReshapeObservationParallel
+.. autoclass:: ReshapeObservationV1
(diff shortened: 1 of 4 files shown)


jkterry1:

teddytennant:
Renamed to ReshapeObservationV1 and ReshapeObservationParallelV1, SuperSuit's reshape_v0 plus one.

teddytennant:
rebased on main; conflicts were __init__.py and the docs list after #1426.

jkterry1:
New merge conflict

jkterry1:
New merge conflicts

PR #1428, opened 2026-08-25
Add ColorReductionObservationV1 wrappers, ported from SuperSuit color_reduction_v0

teddytennant:
# Description

Ports SuperSuit's `color_reduction_v0` as `ColorReductionObservationV1` and `ColorReductionObservationParallelV1`, part of #1365. Drops the channel axis of an image observation, either by converting to grayscale or by taking one of R, G or B, and rewrites the advertised Box to match.

## On the name

I did not reuse `GrayscaleObservation`. Three of the four modes return a raw colour channel with no grayscale conversion in them at all, so the name would be wrong for most of what the wrapper does. Gymnasium's version also has a `keep_dim` argument this doesn't, so the mapping isn't clean in either direction. If you'd rather have a `GrayscaleObservation` that only does `"full"` plus a separate channel-picking wrapper, that's a reasonable split, but it's a different API from SuperSuit's rather than a rename.

## Two things worth knowing

`"full"` returns uint8 whatever the input dtype was, because that's what SuperSuit does. A float image in [0, 1] therefore comes out all zeros. It's a trap, it's now in the docstring, and there's a test pinning it as deliberate rather than accidental.

The R, G and B modes return a copy. SuperSuit returned a view into the environment's own buffer, so writing into the observation wrote into env state.

`state()` and `state_space` pass through unreduced, same as SuperSuit and #1415, so a wrapped image env's global state stays three-channel while observations go to one.

## Verification

    $ pytest test/color_reduction_test.py -q
    54 passed, 2 warnings in 0.14s

    13 passed, 18 warnings in 23.17s

I ran the transform and the derived space against SuperSuit's `change_observation` and `convert_box` over 800 random uint8 images in all four modes, plus float32, float64, int16 and uint16 inputs across six different observation spaces. No divergences in values or dtypes.

## A question

SuperSuit's `"full"` uses the BT.601 luminance weights `[0.299, 0.587, 0.114]`. `gymnasium.wrappers.GrayscaleObservation` uses BT.709, `[0.2125, 0.7154, 0.0721]`. This PR keeps SuperSuit's, so the same environment grayscaled through PettingZoo and through Gymnasium gives different pixel values. Do you want them to match inside Farama, at the cost of silently changing results for anyone migrating off SuperSuit?

## Type of change

# Checklist:

### DIFF
--- a/docs/api/wrappers/pz_wrappers.md
+++ b/docs/api/wrappers/pz_wrappers.md
@@ -129,6 +129,8 @@ while parallel_env.agents:
 .. autoclass:: AssertOutOfBoundsWrapper
 .. autoclass:: ClipOutOfBoundsWrapper
 .. autoclass:: OrderEnforcingWrapper
+.. autoclass:: ColorReductionObservation
+.. autoclass:: ColorReductionObservationParallel
 .. autoclass:: DtypeObservationV1
 .. autoclass:: DtypeObservationParallelV1
 
--- a/docs/api/wrappers/pz_wrappers.md
+++ b/docs/api/wrappers/pz_wrappers.md
@@ -129,8 +129,8 @@ while parallel_env.agents:
 .. autoclass:: AssertOutOfBoundsWrapper
 .. autoclass:: ClipOutOfBoundsWrapper
 .. autoclass:: OrderEnforcingWrapper
-.. autoclass:: ColorReductionObservation
-.. autoclass:: ColorReductionObservationParallel
+.. autoclass:: ColorReductionObservationV1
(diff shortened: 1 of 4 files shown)


jkterry1:

teddytennant:
Renamed to ColorReductionObservationV1 and ColorReductionObservationParallelV1, SuperSuit's color_reduction_v0 plus one.

teddytennant:
Rebased onto main. Conflicts in pettingzoo/utils/wrappers/__init__.py and docs/api/wrappers/pz_wrappers.md after #1426; kept both DtypeObservationV1 and ColorReductionObservationV1.

PR #1429, opened 2026-08-25
Add RescaleObservationV1 wrappers, ported from SuperSuit normalize_obs_v0

teddytennant:
# Description

Ports SuperSuit's `normalize_obs_v0` as `RescaleObservationV1` and `RescaleObservationParallelV1`, part of #1365. Linearly maps each element of a Box observation from the space's own `[low, high]` onto a target range, and advertises a Box over that range.

Named RescaleObservationV1 / RescaleObservationParallelV1: SuperSuit's normalize_obs_v0 plus one, per the review.

The arguments are `min_obs` and `max_obs`, matching Gymnasium. SuperSuit calls them `env_min` and `env_max`, which sounds like the environment's range when it's actually the output range. Defaults are unchanged at 0.0 and 1.0.

## The thing that took the longest

The naive version of this wrapper returns observations that fall outside the space it advertises, and it is not a rare corner. Doing the arithmetic in the source dtype and building the target `Box` by casting `min_obs`/`max_obs` separately gives you two different roundings:

    RescaleObservationV1(env, 0.1, 0.7)   # source Box(-1.0, 3.0, float32)
    observe   -> 0.70000005
    space.high ->  0.7, i.e. float32(0.7) == 0.69999999
    contains  -> False, and api_test fails

About 23% of random float32 endpoint cases escape, and float64 spaces are not safe either. So the rescale is computed in float64 and then clipped to the advertised bounds. `test_observation_stays_in_the_advertised_space` covers three target ranges, both classes, and fails without either half of that.

The clip also means an observation the environment emitted outside its own declared space gets clamped, where SuperSuit passed it straight through. Such an env is already violating the API, and without the clip the wrapper can't guarantee its own advertised space, but it is a behaviour change and I'd rather you knew.

Two checks are new relative to SuperSuit: the space has to be a Box, and every element needs `high > low`. The second one is why: SuperSuit divides by `high - low` unguarded, so a constant feature (`low == high` on one element, which is perfectly legal) silently produced NaN.

Also, `.astype(space.dtype)` on the way out. SuperSuit returns whatever numpy promotes to, usually float64 out of a float32 space, so its own advertised space is wrong about the dtype.

## Verification

    $ pytest test/rescale_observation_test.py -q
    43 passed, 8 warnings in 0.14s

    13 passed, 18 warnings in 25.39s

Beyond the test file I swept the transform over 40000 random spaces per dtype including both endpoints, float32 and float64: no observation outside the advertised space.

## A question

`min_obs` and `max_obs` are scalars here, matching SuperSuit. Gymnasium also accepts per-element arrays, and supports an infinite target bound where the source bound is infinite too. Grow the array form now, or leave it scalar until someone asks?

## Type of change

# Checklist:

### DIFF
--- a/docs/api/wrappers/pz_wrappers.md
+++ b/docs/api/wrappers/pz_wrappers.md
@@ -140,6 +140,8 @@ while parallel_env.agents:
 .. autoclass:: DtypeObservationParallelV1
 .. autoclass:: PadObservationsV1
 .. autoclass:: PadObservationsParallelV1
+.. autoclass:: RescaleObservationV1
+.. autoclass:: RescaleObservationParallelV1
 .. autoclass:: ReshapeObservationV1
 .. autoclass:: ReshapeObservationParallelV1
 .. autoclass:: ScaleActionV1
--- a/pettingzoo/utils/wrappers/__init__.py
+++ b/pettingzoo/utils/wrappers/__init__.py
@@ -24,6 +24,10 @@ from pettingzoo.utils.wrappers.pad_observations import (
 )
 from pettingzoo.utils.wrappers.record_video import RecordVideo
 from pettingzoo.utils.wrappers.record_video_parallel import RecordVideoParallel
+from pettingzoo.utils.wrappers.rescale_observation import (
+    RescaleObservationParallelV1,
+    RescaleObservationV1,
+)
 from pettingzoo.utils.wrappers.reshape import (
     ReshapeObservationParallelV1,
     ReshapeObservationV1,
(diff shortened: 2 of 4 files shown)


jkterry1:

teddytennant:
Renamed to RescaleObservationV1 and RescaleObservationParallelV1, SuperSuit's normalize_obs_v0 plus one.

teddytennant:
Rebased on main. Conflicts were in pettingzoo/utils/wrappers/__init__.py and the docs list after #1426; kept DtypeObservationV1/DtypeObservationParallelV1 and RescaleObservationV1/RescaleObservationParallelV1.

jkterry1:
New merge conflict

jkterry1:
New merge conflicts

teddytennant:

jkterry1:

teddytennant:
rebased on #1432. the conflict is only the alphabetical wrapper list in `__init__.py` and the docs autoclass block, so these three will keep re-conflicting one at a time. happy to keep rebasing, just flagging it.

PR #1430, opened 2026-08-25
Add MaxObservationV1 wrappers, ported from SuperSuit max_observation_v0

teddytennant:
# Description

Named MaxObservationV1 / MaxObservationParallelV1: SuperSuit's max_observation_v0 plus one, per the review.

Ports SuperSuit's `max_observation_v0` as `MaxObservationV1` and `MaxObservationParallelV1`, part of #1365. Each agent's observation becomes the elementwise max over its last `memory` frames, with independent per-agent history. The observation space is unchanged, since the max of points in a Box is still in that Box.

## Where the history advances

This is the whole design of the wrapper, so it's worth being explicit. The AEC class records in `reset()` and `step()`, for `agent_selection` only, which is where SuperSuit records. `observe()` is pure and just reduces what's there.

I originally had it recording inside `observe()` with a per-step guard, and that is wrong in a way that's easy to miss. In any `parallel_to_aec` environment, which is all of the Atari envs this wrapper exists for, `observe()` returns a cached frame that only changes when the whole cycle completes. So anything that reads observations out of turn, a centralised critic or an observation logger, fills the window with copies of the same frame and the max quietly becomes a no-op. There's a test with a two-agent flicker env that asserts the result is identical with and without an extra `{a: env.observe(a) for a in env.agents}` per turn.

An agent that joins mid-episode and has not had a turn yet falls through to the inner `observe()`.

## Two deviations from SuperSuit

Observation spaces are restricted to Box, MultiBinary and MultiDiscrete, checked at construction and again per agent when recording. SuperSuit accepted anything and then broke oddly: on a `Dict` space, `np.copy` gives a 0-d object array, so the first frame silently returns something that fails `contains`, and frame `memory` dies inside numpy with an error that names neither the wrapper nor the space. `api_test` on `generated_agents_env_action_mask_obs_v0` reproduces it.

`memory` must be a positive int. SuperSuit called `int(memory)` and discarded the result, so `0` and `1.5` were accepted and failed later or not at all.

## Verification

    $ pytest test/max_observation_test.py -q
    47 passed, 9 warnings in 0.15s

    13 passed, 18 warnings in 22.99s

Includes `api_test` on `generated_agents_env_v0`, so agents appearing and disappearing mid-episode are covered.

## A question

Rejecting `Dict` observations is the conservative call, but every `classic` env uses `Dict({"observation", "action_mask"})`, so the wrapper is unusable there. Would you rather it recursed into Dict and Tuple and maxed the leaves, which would max the action mask too and is probably wrong, or maxed only a named key like `"observation"`? I didn't want to invent an interface SuperSuit never had without asking.

## Type of change

# Checklist:

### DIFF
--- a/docs/api/wrappers/pz_wrappers.md
+++ b/docs/api/wrappers/pz_wrappers.md
@@ -138,6 +138,8 @@ while parallel_env.agents:
 .. autoclass:: ColorReductionObservationParallelV1
 .. autoclass:: DtypeObservationV1
 .. autoclass:: DtypeObservationParallelV1
+.. autoclass:: MaxObservationV1
+.. autoclass:: MaxObservationParallelV1
 .. autoclass:: PadObservationsV1
 .. autoclass:: PadObservationsParallelV1
 .. autoclass:: ReshapeObservationV1
--- a/pettingzoo/utils/wrappers/__init__.py
+++ b/pettingzoo/utils/wrappers/__init__.py
@@ -15,6 +15,10 @@ from pettingzoo.utils.wrappers.dtype import (
     DtypeObservationParallelV1,
     DtypeObservationV1,
 )
+from pettingzoo.utils.wrappers.max_observation import (
+    MaxObservationParallelV1,
+    MaxObservationV1,
+)
 from pettingzoo.utils.wrappers.multi_episode_env import MultiEpisodeEnv
 from pettingzoo.utils.wrappers.multi_episode_parallel_env import MultiEpisodeParallelEnv
 from pettingzoo.utils.wrappers.order_enforcing import OrderEnforcingWrapper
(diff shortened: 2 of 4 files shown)


jkterry1:

teddytennant:
Renamed to MaxObservationV1 and MaxObservationParallelV1, SuperSuit's max_observation_v0 plus one.

teddytennant:
Rebased on main. Conflicts were in pettingzoo/utils/wrappers/__init__.py and docs/api/wrappers/pz_wrappers.md after #1426; kept both DtypeObservationV1 and MaxObservationV1.

jkterry1:
New merge conflict

jkterry1:
New merge conflicts

teddytennant:

jkterry1:

teddytennant:
rebased on #1432.

PR #1431, opened 2026-08-25
Add StickyActionV1 wrappers, ported from SuperSuit sticky_actions_v0

teddytennant:
# Description

Ports SuperSuit's `sticky_actions_v0` as `StickyActionV1` and `StickyActionParallelV1`, part of #1365. With probability `repeat_action_probability` the agent's previous action is passed to the environment instead of the one just submitted. Previous actions are tracked per agent.

## Three things this fixes relative to a direct port

The previous action is deep-copied, both when stored and when replayed. Storing the caller's array by reference means that a policy loop reusing one action buffer, which is the normal way to write one, gets no stickiness at all: the "previous" action is the same object as the current one. And an env that normalises or clips its action in place corrupts the wrapper's memory.

The wrapper's RNG is spawned from a child `SeedSequence` rather than seeded with the same integer the inner env gets. Almost every PettingZoo env does `seeding.np_random(seed)` in its own `reset`, so seeding the wrapper the same way produces a byte-identical generator. Against a one-agent env drawing one uniform per step I got perfect lockstep: actions stuck exactly when the env's own transition noise was below the threshold. It's still fully reproducible from the seed, just an independent stream. `SeedSequence.spawn` rather than `Generator.spawn`, since the latter needs numpy >= 1.25 and `pyproject.toml` allows >= 1.21.

`step()` after every agent is done no longer raises. `_was_dead_step` empties `terminations` while `agent_selection` still names the removed agent, so reading `terminations[agent]` before the inner guard runs gave a `KeyError` where `OrderEnforcingWrapper` would have logged its warning and returned. Since every `env()` factory returns an `OrderEnforcingWrapper`, `StickyActionV1(rps_v2.env())` is the normal composition and hit this. Same guard also stops `step()` before `reset()` from degrading the documented error into an `AttributeError`.

## Two things documented rather than fixed

`reset(seed=None)` starts a fresh unseeded stream, matching SuperSuit and this repo's own example envs, but the opposite of `gymnasium.Env.reset`, which reseeds only when a seed is given. So you need a seed at every reset for a whole run to be reproducible. That's in the docstring now.

Only the wrapper's own `reset` clears the memory. Under `MultiEpisodeParallelEnv` the inner env auto-resets without going through it, so episode 2's first action can be episode 1's replayed one. Also in the docstring.

## Verification

    $ pytest test/sticky_action_test.py -q
    42 passed, 2 warnings in 0.14s

    13 passed, 18 warnings in 23.02s

Covers dead agents stepped with `None`, step after done, step before reset, seed forwarding, per-agent isolation, buffer reuse, in-place env mutation, and that the sticky decisions don't track the env's own noise.

## A question

`repeat_action_probability` is checked with `assert`, matching SuperSuit and the `isinstance` assert in #1415. It disappears under `python -O`, and `gymnasium.wrappers.StickyActionV1` raises `gymnasium.error.InvalidProbability` for exactly this. Should these wrappers raise rather than assert on user-supplied values? It's a one-line change and it probably wants to be consistent across the whole migration.

## Type of change

# Checklist:

### DIFF
--- a/docs/api/wrappers/pz_wrappers.md
+++ b/docs/api/wrappers/pz_wrappers.md
@@ -144,5 +144,7 @@ while parallel_env.agents:
 .. autoclass:: ReshapeObservationParallelV1
 .. autoclass:: ScaleActionV1
 .. autoclass:: ScaleActionParallelV1
+.. autoclass:: StickyActionV1
+.. autoclass:: StickyActionParallelV1
 
 ```
--- a/pettingzoo/utils/wrappers/__init__.py
+++ b/pettingzoo/utils/wrappers/__init__.py
@@ -32,6 +32,10 @@ from pettingzoo.utils.wrappers.scale_action import (
     ScaleActionParallelV1,
     ScaleActionV1,
 )
+from pettingzoo.utils.wrappers.sticky_action import (
+    StickyActionParallelV1,
+    StickyActionV1,
+)
 from pettingzoo.utils.wrappers.terminate_illegal import TerminateIllegalWrapper
 
 __all__ = [
@@ -57,5 +61,7 @@ __all__ = [
     "ReshapeObservationV1",
     "ScaleActionParallelV1",
     "ScaleActionV1",
+    "StickyActionParallelV1",
+    "StickyActionV1",
     "TerminateIllegalWrapper",
 ]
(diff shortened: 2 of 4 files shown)


jkterry1:

teddytennant:
Renamed to StickyActionV1 and StickyActionParallelV1, SuperSuit's sticky_actions_v0 plus one.

teddytennant:
Rebased on main. Conflicts were in pettingzoo/utils/wrappers/__init__.py and the docs list after #1426; both DtypeObservationV1 and StickyActionV1 are kept.

jkterry1:
New merge conflicts

teddytennant:

jkterry1:

teddytennant:
rebased on #1432 too.

PR #1432, opened 2026-08-25
Add ScaleActionV1 wrappers, ported from SuperSuit scale_actions_v0

teddytennant:
# Description

Ports SuperSuit's `scale_actions_v0` as `ScaleActionV1` and `ScaleActionParallelV1`, part of #1365. The advertised Box action space has its `low` and `high` multiplied by `scale`, and an action submitted against that space is mapped back into the environment's own space.

## This one changes behaviour, on purpose

`scale_actions_v0` scales the bounds by `scale` and then also multiplies the incoming action by `scale`. So an action sampled from the space the wrapper advertises reaches the environment `scale**2` times too large. For `scale > 1` that is out of range: on `Box(-1, 1)` at `scale=2`, 1975 of 2000 sampled actions land outside the inner space, and at `scale=3.7` all 2000 do. For `scale < 1` it stays in range but can never reach the full space.

There's no reading of `scale` under which both halves are right, so I treated it as a bug rather than a convention. This PR keeps the advertised space as SuperSuit had it and divides the action instead of multiplying it, so `env.action_space(agent).sample()` is always valid for the wrapped env. That is the direction `gymnasium.wrappers.RescaleAction` goes: advertise the new interval, map back into the inner one.

I'd rather you knew about this than discovered it, so: **anyone porting from `scale_actions_v0` will see different behaviour.** If you'd rather the port reproduce SuperSuit exactly, bug included, say so and I'll change it.

## The endpoints need a clip, and it is not unconditional

`low * scale` is computed in the space's dtype and the inverse is not exact, so the endpoint of the advertised space maps an ULP outside the inner space. On `Box(-1.7, 1.7, float32)` at `scale=3.0` the env receives `1.7000001668930054` against a bound of `1.7000000476837158`, and with `AssertOutOfBoundsWrapper` underneath that's a crash. It fails for every scale that isn't a power of two, and float64 is no safer, since `x * s / s` is not exact in any finite precision.

The clip is elementwise and conditional on the submitted action having been inside the advertised space. An unconditional clip would silently sanitise genuinely out-of-range actions, which hides caller bugs and duplicates `ClipOutOfBoundsWrapper`. So valid in gives valid out, and out of range stays out of range for `AssertOutOfBoundsWrapper` to catch.

## Three smaller deviations

A negative `scale` swaps the bounds and reflects, rather than letting `Box` raise `ValueError` on `low > high` the way SuperSuit does. `scale=0` is rejected, where SuperSuit gave you a degenerate `Box(0, 0)`. And the inner dtype and shape are preserved, where SuperSuit's `Box(low=..., high=...)` dropped both and silently made everything float32.

## Verification

    $ pytest test/scale_action_test.py -q
    46 passed, 2 warnings in 0.22s

    13 passed, 18 warnings in 23.24s

`test_action_from_advertised_space_is_valid_inside` steps the env across seven scales and three bounds, including both endpoints, and asserts both that the inner space contains what it received and that the value is the expected inverse. The value assertion matters: once the clip is in, containment alone passes for almost any transformation.

## A question

Once the direction is fixed, `ScaleActionV1(env, s)` is exactly `RescaleAction(env, low*s, high*s)`, the proportional special case of Gymnasium's wrapper. Would you rather have a full multi-agent `RescaleAction(env, min_action, max_action)` matching Gymnasium's signature, with the scalar case falling out of it, instead of a separate scalar-only wrapper?

## Type of change

# Checklist:

### DIFF
--- a/docs/api/wrappers/pz_wrappers.md
+++ b/docs/api/wrappers/pz_wrappers.md
@@ -142,5 +142,7 @@ while parallel_env.agents:
 .. autoclass:: PadObservationsParallelV1
 .. autoclass:: ReshapeObservationV1
 .. autoclass:: ReshapeObservationParallelV1
+.. autoclass:: ScaleActionV1
+.. autoclass:: ScaleActionParallelV1
 
 ```
--- a/pettingzoo/utils/wrappers/__init__.py
+++ b/pettingzoo/utils/wrappers/__init__.py
@@ -28,6 +28,10 @@ from pettingzoo.utils.wrappers.reshape import (
     ReshapeObservationParallelV1,
     ReshapeObservationV1,
 )
+from pettingzoo.utils.wrappers.scale_action import (
+    ScaleActionParallelV1,
+    ScaleActionV1,
+)
 from pettingzoo.utils.wrappers.terminate_illegal import TerminateIllegalWrapper
 
 __all__ = [
@@ -51,5 +55,7 @@ __all__ = [
     "RecordVideoParallel",
     "ReshapeObservationParallelV1",
     "ReshapeObservationV1",
+    "ScaleActionParallelV1",
+    "ScaleActionV1",
     "TerminateIllegalWrapper",
 ]
--- a/pettingzoo/utils/wrappers/scale_action.py
+++ b/pettingzoo/utils/wrappers/scale_action.py
@@ -1,0 +1,172 @@
+"""Wrappers that scale the bounds of a Box action space."""
+
+from __future__ import annotations
+
+from typing import Any, cast
+
+import gymnasium.spaces
+import numpy as np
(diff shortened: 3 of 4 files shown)


jkterry1:

teddytennant:
Renamed to ScaleActionV1 and ScaleActionParallelV1, SuperSuit's scale_actions_v0 plus one.

teddytennant:
Rebased on main. Conflicts were in pettingzoo/utils/wrappers/__init__.py and docs/api/wrappers/pz_wrappers.md after #1426; kept DtypeObservationV1 / DtypeObservationParallelV1 and ScaleActionV1 / ScaleActionParallelV1.

jkterry1:
New merge conflicts

teddytennant:

PR #1433, opened 2026-08-25
Add PadObservationsV1 wrappers, ported from SuperSuit pad_observations_v0

teddytennant:
# Description

Ports SuperSuit's `pad_observations_v0` as `PadObservationsV1` and `PadObservationsParallelV1`, part of #1365. Pads every agent's observation up to the largest observation space across `possible_agents`, so all agents advertise the same space. Multi-agent only, so there's no Gymnasium equivalent to name it after.

Padding goes at the end of every axis, so the real values sit at the leading corner, and the padded Box bounds are computed the way `homogenize_spaces` computes them, including the `max(1e-5, ...)` clamp on the high side. That clamp looks like a typo and isn't: it's what guarantees the zero fill is inside the advertised bounds when every real bound is negative. There are two tests on it now, including the integer case where `1e-5` truncates to 0.

## Two bugs fixed rather than ported

`homogenize_observations` returns the space object instead of the observation in its Discrete branch, so `pad_observations_v0` on an env with Discrete observations hands the agent a `Discrete(4)` object as its observation. This returns the observation.

Fixing that surfaced a second one. Taking `Discrete(max(n))` and discarding `start` gives a space that doesn't contain the observation: `Discrete(3, start=5)` alongside `Discrete(5)` advertised `Discrete(5)` for an observation of `5`, and `api_test` fails on it. The padded space is now `Discrete(end - start, start=start)` over the min start and max end, which reduces to `Discrete(max(n))` when every start is 0, so the common case is unchanged.

## One thing kept

An agent whose observation is already the full shape gets its array back without a copy, same as SuperSuit, so writing into it writes into the environment. That's in both docstrings now rather than being left as a surprise. Everyone else gets a fresh array from `np.pad`.

## Verification

    $ pytest test/pad_observations_test.py -q
    34 passed, 4 warnings in 0.13s

    13 passed, 18 warnings in 23.51s

I diffed the port against SuperSuit's real `homogenize_ops` on a 4000-case fuzz over five dtypes, ranks 1 to 3, two and three agents, and random per-axis shapes and bounds, plus a 2000-case Discrete fuzz. No mismatches on the padded space or the padded observation for the Box path, and `padded_space.contains(padded_obs)` held throughout.

## A question

`observation_space(agent)` returns the same object for every agent, since the padded space is identical for all of them. That means `observation_space("agent_1").seed(0)` reseeds the generator `observation_space("agent_0").sample()` draws from. SuperSuit did the same. Want a per-agent copy instead?

Also: should a `PadActions` port of `pad_action_space_v0` follow? Padding observations alone still leaves action spaces heterogeneous, and the two are normally used together. Happy to take it if nobody else has.

## Type of change

# Checklist:

### DIFF
--- a/docs/api/wrappers/pz_wrappers.md
+++ b/docs/api/wrappers/pz_wrappers.md
@@ -131,5 +131,7 @@ while parallel_env.agents:
 .. autoclass:: OrderEnforcingWrapper
 .. autoclass:: DtypeObservationV1
 .. autoclass:: DtypeObservationParallelV1
+.. autoclass:: PadObservations
+.. autoclass:: PadObservationsParallel
 
 ```
--- a/docs/api/wrappers/pz_wrappers.md
+++ b/docs/api/wrappers/pz_wrappers.md
@@ -131,7 +131,7 @@ while parallel_env.agents:
 .. autoclass:: OrderEnforcingWrapper
 .. autoclass:: DtypeObservationV1
 .. autoclass:: DtypeObservationParallelV1
-.. autoclass:: PadObservations
-.. autoclass:: PadObservationsParallel
+.. autoclass:: PadObservationsV1
+.. autoclass:: PadObservationsParallelV1
 
 ```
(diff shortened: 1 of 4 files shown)


jkterry1:

jkterry1:

teddytennant:
Renamed to PadObservationsV1 and PadObservationsParallelV1, SuperSuit's pad_observations_v0 plus one.

jkterry1:
@teddytennant this PR and all your other wrapper PRs have merge conflicts?

teddytennant:
yeah, #1426 landed in __init__.py and the docs list. rebased this one and the other wrapper PRs onto main.

Issue #1364, opened 2026-06-24
Port multiwalker from box2d to pymunk following gymnasium port


jkterry1:
See https://github.com/Farama-Foundation/Gymnasium/issues/1597

PR #1435, opened 2026-08-26
Prototype Multiwalker physics with Pymunk

tzh476:
# Description

Publishes the requested Multiwalker Pymunk prototype so maintainers and future contributors can
review and continue the migration.

Related to #1364 and Farama-Foundation/Gymnasium#1597.

The existing Box2D-backed `multiwalker_v9`, registrations, documentation, and public API are
unchanged. Nothing here is wired into a released environment version.

## In place

- three articulated walkers (hulls, four leg bodies, pivot joints, rotary limits, motors), a shared
  package, density-derived mass/center-of-gravity/moment, Box2D-style 180 solver iterations
- procedural and hardcore terrain, generated from the same `np_random` draw sequence as
  `multiwalker_base._generate_terrain`
- lower-leg contact tracking, hull/package failure detection, terrain LIDAR, the existing 24-value
  walker state plus neighbor/package observations, and the reward/termination skeleton

## Parity, and what it is measured against

**Mass/inertia.** Pymunk hull reproduces the Box2D reference mass; its center-of-gravity moment plus
the parallel-axis term reproduces the Box2D body-origin inertia.

**Terrain equivalence is the RNG draw sequence, not the geometry** — a port consuming the stream in a
different order gets different ground for the same seed, and nothing errors. 5 seeds x hardcore
on/off match draw for draw, heights identical at `atol=0`. Obstacles are matched on centroid and
extent, not exact equality, because Box2D stores vertices as float32.

**Motor torque.** Box2D `maxMotorTorque` and Pymunk `SimpleMotor.max_force` are different
quantities; the mapping is pinned by six stalling cases, and a `dt`-scaled mapping I proposed is
refuted by its own test.

**Fixed-action trajectory parity has a hard limit.** Perturbing one motor command by `1e-6` amplifies
to `0.1` by step 82, so any two solvers diverge past that horizon regardless of correctness. The test
bounds a short horizon and records that number instead of asserting long-run equality.

## One correction worth carrying to the other ports

`pymunk_terrain.py` had complete generation with 26 passing tests while `reset()` still called
`_create_flat_terrain()` — for five days. Its only importer was its own test file: every test pointed
at the module, none at the environment, and a component suite cannot see that nothing imports it.
Fixed in `67e0a50a`, which also consumes the cloud RNG draws — `_generate_clouds` is decorative and
this prototype does not render, but it draws 11 times per cloud and shares the stream with
observation noise, so skipping it aligns the terrain and silently shifts every draw after it.

## A smaller follow-up in `ec4dd353`, and the claim I had to retract

Both new test files opened with a module-level `pytest.importorskip("Box2D")`, which skips *every*
test in the file — including the twenty that never touch Box2D. `ec4dd353` scopes the skip to the
class and the two functions that actually compare the two engines.

**This does not affect CI, and I initially wrote that it did.** I claimed `linux-test (3.14)` was
green while collecting none of these tests. Then I read the job log for the previous head, and it
shows the opposite: `box2d-py==2.3.8` builds from source there via the pinned `swig`, and the
Box2D-only tests ran — 25 PASSED in `TestAgainstTheRealBox2DGenerator`, 19 PASSED in the motor-parity
file. Nothing was being skipped on CI. Correcting it here rather than leaving a tidier but false
statement in the PR.

What the change is actually worth is narrower. `box2d-py` on 3.14 has no wheel and is built from the
sdist, so it needs a working SWIG; on a machine without one the install fails, and *before* this
commit that turned into a silent loss of the twenty pymunk-only tests rather than a visible skip of
the three cross-engine ones:

| `test/sisl` | Box2D importable | Box2D not importable |
|---|---:|---:|
| before `ec4dd353` | 102 passed | 45 passed, 2 skipped |
| after | 102 passed | 65 passed, 37 skipped |

The twenty are load-bearing, not filler. With Box2D absent, reverting `reset()` to flat ground fails
12 of the 13 recovered checks in `test_..._env_terrain.py`, dropping the cloud-draw call fails 2, and
pinning `hardcore` to `False` fails 1 — before the change all three mutants passed. The seven
recovered motor-parity tests are the ones that establish that experiment is not blind
(`test_cases_are_distinct_operating_points`, pure arithmetic, and `test_the_budget_actually_binds`,
six parametrised pymunk-only cases).

So: no CI behaviour changes, 102 stays 102 wherever Box2D imports, and a contributor without SWIG
now keeps the tests that need only pymunk. Worth a glance for the other ports in
Farama-Foundation/Gymnasium#1597, which all face the same shape — a reference engine that is
awkward to install on the newest Python, and pymunk-side tests that should not disappear with it.

## Still open

rendering; fallen-agent removal (`remove_on_fall`); learning-curve parity; not registered as a new
environment version.

## Type of change

## Testing

CI was 48/48 green on the previous head, and that run did execute the full `test/sisl` suite on
3.10–3.14 (410 passed in the 3.14 job).

Locally, 102 passed in three configurations: 3.12 with `box2d==2.3.10`, 3.14 with `box2d-py==2.3.8`
built against a system SWIG, and 3.14 without Box2D gives 65 passed / 37 skipped as above. All with
pymunk 7.3.0.

Note for local runs on 3.14: the repo's `addopts` includes `-n=auto` and `pytest-xdist` crashes there
(`KeyError: WorkerController gw2`), so serial needs `-o addopts=""`. The pinned `pyupgrade` hook also
fails on 3.14 — it calls `ast.Str`, removed in that version — identically on files this PR does not
touch. CI is unaffected by both; happy to file separately.

# Checklist

- [x] I have run the pre-commit checks on the changed files
- [x] I have run the full pytest suite
- [x] I have commented the prototype boundaries and non-obvious physics mapping
- [ ] I have made corresponding documentation changes
- [x] I resolved warnings generated by the focused tests
- [x] I added focused tests for the prototype
- [x] Focused tests pass locally


tzh476:
Pushed the procedural terrain, so this now generates the actual task rather than
flat ground: `pymunk_terrain.py` plus tests, `6d39a84`.

The thing I want to flag, because it would be easy to get wrong in the other
environments too: **equivalence here is the RNG draw sequence, not the resulting
geometry.** The terrain is fully determined by draws from `np_random`, so a port
that consumes that stream in a different order produces different ground for the
same seed. Nothing errors — reproducibility just silently stops holding, and
learning-curve comparisons against the Box2D version stop meaning anything. So
the state machine mirrors `multiwalker_base._generate_terrain` line for line and
only body construction is replaced.

The tests transcribe the Box2D state machine into a reference function (deleting
only body construction, keeping every draw in place) and require the two draw
sequences to be exactly equal, over 5 seeds × hardcore on/off. I checked the
assertions aren't vacuous by injecting two faults:

| injected fault | result |
|---|---|
| swap `random()` / `integers(4,5)` in STAIRS | 5 failed / 21 passed |
| drop the GRASS `uniform(-1,1)` draw | 20 failed / 6 passed |
| correct implementation | 26 passed |

31/31 pass locally on the branch (5 prototype + 26 terrain).

One Pymunk-specific gotcha worth carrying to the other ports: concave terrain has
to be a `Segment` chain, because Pymunk polygons must be convex. Building it as
polygons looks fine until something falls through the ground — there's a 600-step
simulation asserting a ball doesn't.

Still explicitly a prototype and still not wired into `multiwalker_v9`: LIDAR,
rewards, `scroll_subroutine`, and rendering are not done. Happy to keep going on
those, or to split this differently if you'd rather review it in smaller pieces.

A process question, since Python 3.14 is the reason this port matters: the repo's
`addopts` includes `-n=auto`, and `pytest-xdist` crashes on 3.14 here
(`KeyError: WorkerController gw2`), so I ran serial with `-o addopts=""`. Worth a
separate issue?

tzh476:
The red `docs-test (10)` on this branch is an Atari ROM fetch, not this PR. Evidence, in case it saves anyone a look:

**What failed**

```
ERROR docs/code_examples/usage_aec.py - OSError: rom space_invaders is not installed.
Please install roms using AutoROM tool ...
8 passed, 318 warnings, 4 errors
```

The other seven `docs-test` shards show `CANCELLED` because `fail-fast` killed them once shard 10 errored, not because they found anything.

**Why it isn't this branch**

- `docs-test.yml` runs `pytest docs --markdown-docs -m markdown-docs --ignore=tutorials --ignore=docs/tutorials`. It never collects `test/`, so nothing this PR adds is in scope.
- This PR touches 0 files under `docs/`, `code_examples/`, or anything Atari.
- Same workflow, same branch, earlier today: `81006e5c` → success, `27e81c19` → failure. The only difference between those two commits is one added file, `test/sisl/test_multiwalker_pymunk_trajectory_parity.py` (+203/−0) — which that workflow does not run. The 08-29 run also passed.

So the delta is `AutoROM -v` in the setup step, not the diff.

I don't have rerun rights here, so I can't retry it myself — a re-run of that job should clear it. Flagging rather than pushing an empty commit to trigger it, since that would just add noise to the history.

*Disclosure: this analysis and this PR's code were produced with AI assistance (Claude). The CI logs, workflow definition, and commit comparison quoted above were read directly from this repository's Actions output and API.*

tzh476:
Pushed 59668bf. While checking this prototype against the merge standard @pseudo-rnd-thoughts set out on Farama-Foundation/Gymnasium#1602 — (1) same goal and reward functions, (2) similar training curves, (3) higher implementation standards — I found that criterion 1 was not met, and the miss was large.

`multiwalker_base.scroll_subroutine` charges the fall penalty per walker, and when falling does not end the episode for everyone it also charges that walker `terminate_reward` and ends it alone:

```python
if fallen:
    rewards[i] += self.fall_reward
    if self.remove_on_fall:
        walker._destroy()
    if not self.terminate_on_fall:
        rewards[i] += self.terminate_reward   # per walker
    done[i] = True                            # per walker
```

This prototype had collapsed that into a single global branch. Enumerating the reward algebra over 48 states (`terminate_on_fall` × fallen-pattern × `game_over` × `package_x`) gave **18/48 divergent rewards and 3/48 divergent dones**, the reward error being a flat **−100 per fallen walker** — two orders of magnitude above a typical per-step shaping delta, so large enough to change what a policy learns rather than merely shift returns. With `terminate_on_fall=False`, no walker was ever marked done.

Fixed, with six tests in `test/sisl/test_multiwalker_pymunk_reward_parity.py` that drive the real `step` with the fallen mask forced, so they test the reward algebra rather than solver dynamics. Reverting the fix fails two of them. Existing suite: 71 passed, 37 skipped.

Worth flagging for the LunarLander port too, @Lonny154 — the per-walker clause is easy to miss because the collapsed version looks right and produces plausible returns.

On criterion 2, one question before I spend the compute. Training curves need a policy on both backends, and the Box2D side needs `box2d-py`, which has no cp314 wheel — so a comparison has to pin an older Python. Would you rather I:

1. run PPO on Box2D `multiwalker_v9` vs the pymunk prototype at matched hyperparameters on Python 3.12, and post both curves, or
2. leave the curve comparison until the remaining physics gaps (rendering, fallen-agent removal, motor-force calibration) are closed, since a curve mismatch now would not tell us which gap caused it?

I lean toward 2 — a curve difference today is not attributable — but I will run 1 if you would rather see numbers early. I am not asking anyone to review the physics yet; this is only about which of the two produces something useful for you.

Lonny154:
Thanks for the heads up @tzh476 I checked my LunarLander prototype and found a similar mismatch. I had added a `hard_leg_Impact` heuristic that terminated `-100` on sufficiently hard leg-only contact, even though the Box2D environment only marks a crash when the lander hull itself makes contact.

I added a direct Box2D/Pymunk parity test to isolate that case, confirmed Box2D continued while Pymunk terminated, then removed the extra heuristic. The focused LunarLander suites are still green at 71 passed / 6 skipped. Pushed the fix in `51ed030fb`.

PR #1440, opened 2026-09-04
Fix MultiEpisode wrappers dropping the seed after reset(seed=0)

yurekami:
# Description

`MultiEpisodeEnv` and `MultiEpisodeParallelEnv` advance the seed for every internal reset with

    self._seed = self._seed + 1 if self._seed else None

That truthiness check treats the valid seed `0` like "no seed", so after `reset(seed=0)` every following episode is reset with `seed=None`. The comment right above the line says the seed is incremented per episode, and that is what happens for every other seed:

| first reset | seeds handed to the inner env (3 episodes) |
|---|---|
| `reset(seed=42)` | `[42, 43, 44]` |
| `reset(seed=0)` | `[0, None, None]` |

So episode k of a seed-0 evaluation no longer reproduces a standalone `env.reset(seed=k-1)`, while it does for any other seed (checked with `pistonball_v6`: the wrapper's second episode equals `parallel_env().reset(seed=43)` for seed 42, but not `reset(seed=1)` for seed 0).

The fix compares against `None` instead, so seed 0 is incremented like any other seed. Both wrappers are changed in the same way.

Tests: `test/wrapper_test.py` gains `test_multi_episode_env_wrapper_seeds_every_episode` and `test_multi_episode_parallel_env_wrapper_seeds_every_episode`, parametrised over seeds `0` and `42`. Each wraps the inner env in a small recording wrapper and asserts the seed sequence `[seed, seed + 1, seed + 2]`. The seed-0 cases fail on `main` (`assert [0, None, None] == [0, 1, 2]`) and pass with the fix. `test/wrapper_test.py`: 17 passed.

## Type of change

# Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011vZKTbRaDh8FuXSbVpQomD


Trenza1ore (approved):

virgilt:
I reviewed the seed propagation tests for both `0` and a nonzero seed. Both are preserved across the internal environment resets, and the full CI matrix passes.

PR #1449, opened 2026-09-05
Apply ruff autofixes to the ClipReward/wrapper test files

tzh476:
`pre-commit` is failing on main. `ruff check` reports three I001 (unsorted import
block) errors and `ruff format` wants three files reformatted. Running the repo's
own pinned hooks (ruff 0.14.10) fixes all of them mechanically:

```
before: Found 3 errors. / 3 files would be reformatted
after:  All checks passed! / 265 files already formatted
```

Import ordering and line wrapping only — no logic touched. Verified
behaviour-neutral: the three affected test files give 4 failed / 106 passed both
before and after, the same four failures. Those four also fail on unmodified
main, so they are pre-existing and out of scope here.

This also clears the red `pre-commit` check on #1435, which picked the failure up
by merging main.


jkterry1:
Hey thanks a ton :)

PR #1297, opened 2025-04-30
Update knights_archers_zombies to v11

dm-ackerman:
# Description

Behavior changes:
* replaced `vector_state`, `use_typemasks`, and `sequence_space`
  parameters with `obs_method`
  The old options had multiple invalid combinations. The obs_method
  options are listed in the documentation
* fixed bugs related to initial spawning
* fixed state, obs, typemasks to match docs
* arrows persist even if the archer that fires them is removed
* 'pad_observation' parameter was removed (it did nothing)
* 'spawn_rate' argument is now 'spawn_delay'. Same Behavior.
  'Rate' implied higher numbers were faster which was the
  opposite of the behavior.
* fixed incorrect bounds on agent movement

Other changes:
* Documentation was updated and corrected
* doc strings added
* Type hints were added
* Fixed issue with FPS scaling movement
* Agent actions were explicitly defined
* Removed redundant/dead code
* Made various classes into subclasses as needed
* Moved agent logic into agent classes
* Removed extraneous data structures
* Adjusted rendering/draw/screen code
* Rotated player icons to correct orientation
* Added env specific tests
* Combined all zombie actions into single function
* Combined weapon actions into single function
* Add check that render_mode is valid
* Move space creations to own functions
* Fixed some int/float confusion
* Renamed items to avoid name shadowing/collision
* Did misc code cleanup

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

# Checklist:

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [x] I have run `pytest -v` and no errors are present.
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I solved any possible warnings that `pytest -v` has generated that are related to my code to the best of my knowledge.
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1398, opened 2026-07-13
Fix cooperative_pong ball bounce physics (v7) — supersedes #1336

RewardGuard:
# Description

Supersedes #1336 (which had merge conflicts and a failing CI) and fixes #1289.

Cooperative pong resolved collisions by snapping the ball's position exactly onto the colliding surface, throwing away the distance the ball had already travelled *past* the wall/paddle during that step. The correct behaviour is to reflect that overshoot back out of the surface, so the bounce conserves the ball's motion within the step.

This PR:

- **`ball.py`** — reflects the overshoot on top/bottom wall bounces instead of clamping to the wall.
- **`paddle.py`** — reflects the overshoot on all four paddle-collision cases (left/right/top/bottom) instead of clamping to the paddle edge.
- Bumps `cooperative_pong_v6` → `v7` (the physics change alters environment dynamics) and updates every reference: the version module, `all_modules.py`, the env registry, docs, and tests.
- Adds a v7 entry to the Version History.

Rebased cleanly on current `main`, so there are no merge conflicts (the `pyproject.toml`/`pygame_init_test` fixes that #1336 carried are already present on `main`).

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- This change requires a documentation update

## Checklist:

- [x] I have run `pytest` for the affected env (`api_test`, `parallel_api_test`, `seed_test` all pass for `cooperative_pong_v7`).
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] New and existing unit tests pass locally with my changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1453, opened 2026-09-07
Reflect the cooperative pong ball off horizontal walls instead of clamping it

### LINKED ISSUE #1289: [Bug Report] Cooperative pong wall bounce physics is wrong
### Describe the bug

When the ball passes through either top or bottom wall, the (psuedo-)code for the ball is:

```
x += vx
y += vy

# note: in pygame, larger values are *below* smaller ones so this is
# saying if the bottom of the ball is bellow the bottom of the screen
if ball.bottom > area.bottom:
    ball.bottom = area.bottom  # set the ball on the bottom wall
    vy = -vy  # reverse the vertical velocity 
```

Instead of bouncing off the wall, it goes through the wall and then is teleported back to the wall. The collision is elastic, but the position of the ball is incorrect. See image below. This error gets worse with speed. (note the same error may exist with bounces off of paddles. I didn't check.)

Image

### Code example

```shell

```

### System info

Installed via pip
```
# python -c "import pettingzoo; print(pettingzoo.__version__)"
1.24.3
# uname -a
Linux pz 5.15.0-136-generic #147-Ubuntu SMP Sat Mar 15 15:53:30 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
# python --version
Python 3.11.5
```

### Additional context

I'm not currently working on a fix for this. The question for someone else to answer is whether to fix this or not. One can say this is just the physics of the game and leave it. In that case it is correct by definition. I do not know how important it is to people that the bounce happens the way someone might expect from comparison to billiards, for example.

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/Farama-Foundation/PettingZoo/issues) in the repo

tzh476:
### Description

Closes #1289.

`Ball.update2` moved the ball, then set `rect.bottom = area.bottom` (or `rect.top = area.top`) when it had crossed a wall. That discards the distance travelled after the crossing, leaving the ball resting against the wall instead of where it physically belongs. The discarded distance scales with speed, which is why the artifact is invisible in a slow game and obvious in a fast one — exactly as the reporter's screenshot shows.

### Measured, before the fix

400×300 area, ball one pixel from the bottom wall, moving straight down. Comparing the resulting edge against the mirrored position:

| ball speed | resulting edge | mirrored (physical) | error |
|---:|---:|---:|---:|
| 5 | 300 | 297 | **3 px** |
| 10 | 300 | 292 | **8 px** |
| 20 | 300 | 282 | **18 px** |
| 40 | 300 | 262 | **38 px** |

Reflecting about the wall (`2 * wall - edge`) gives **0 px error at every speed above, on both walls**, and a ball that never reaches a wall is left untouched.

### Version bump

This changes trajectories, so per @benblack769's and @jkterry1's comments on #1289 the environment is bumped **v6 → v7** across the repo — module file, `all_modules`, `env_registry`, docs and tests — following the pattern of #1297.

### Tests

`test_wall_bounce_is_reflected_not_clamped` is parametrized over both walls × four speeds.

| check | result |
|---|---|
| `test_ball.py` with the fix | **10 passed** |
| the same tests against the **unmodified** `ball.py` | **8 failed** (all new cases), 2 pre-existing pass |
| env loads and steps as `cooperative_pong_v7` | 81 steps, registry key `butterfly/cooperative_pong_v7` |
| `test/pygame_init_test.py -k pong` | 2 passed |
| `ruff format --check` | clean |

The mutation result is the one worth noting: the new tests fail without the source change, so they are load-bearing rather than decorative.

`ruff check` reports 4 findings in this tree (SIM102, PLR1722, and two SIM113). All four are present identically on the unmodified files, so they predate this change and I have left them alone rather than widening the diff.

### AI use

This contribution was assisted by an AI tool (`Assisted-by: Claude (Anthropic)` in the commit). I verified every claim above by running it — the before/after numbers and the mutation check are measurements, not estimates — and I'm happy to answer questions about any part during review.


tzh476:
Closing this in favour of #1398, which is older, broader and already green.

I opened this without noticing #1398 was live — my mistake. Comparing them:

- The wall fix is the same algebra. Their `area.bottom - overshoot` and my `2*area.bottom - rect.bottom` are identical expressions.
- **#1398 covers strictly more**: it also reflects the overshoot in `paddle.py` for all four paddle-collision cases. This PR only touched `ball.py`.
- Both bump `cooperative_pong_v6` → `v7` across the version module, `all_modules.py`, the registry, docs and tests.

The one thing here that #1398 does not have is a regression test. `test_ball.py` adds a parametrized test over ball speeds × both walls, asserting the ball lands at the reflected position rather than resting on the wall, and it fails on `main` before the fix. @RewardGuard — you are welcome to lift it into #1398 verbatim if it is useful; it needs no changes beyond the import. No attribution needed.

Sorry for the extra notification, @jkterry1.

PR #1446, opened 2026-09-04
pursuit: render agents as squares and update the docs gif

teddytennant:
# Description

Based on #1399. Fixes #1360.

Pursuit agents (red pursuers, blue evaders) are drawn as filled squares at 80% of a grid cell and centered in it, which is the pre-pygame look the issue asked for.

#1399 already had this render change, but Jordan asked for a new PR with a regenerated gif. I reimplemented `draw_agent_square` on current main and recorded `docs/environments/sisl/sisl_pursuit.gif` from `render_mode="rgb_array"` instead of copying the blob from #1399.

Commands actually run:

```
python3 -m venv .venv
.venv/bin/pip install pygame-ce pillow numpy gymnasium scipy pymunk
.venv/bin/pip install -e ".[sisl]"
# then a short random episode (seed=42, 60 frames) via pursuit_v6.env(render_mode="rgb_array")
# frames written with Pillow, encoded with:
ffmpeg -y -framerate 5 -i /tmp/pz-pursuit-gif-frames/frame_%04d.png -filter_complex "split[s0][s1];[s0]palettegen=stats_mode=full[p];[s1][p]paletteuse=dither=bayer:bayer_scale=5" docs/environments/sisl/sisl_pursuit.gif
```

`env.render()` returned a `(480, 480, 3)` uint8 ndarray. A sampled frame had a 24x24 red blob with filled corners (fill ratio 0.969), so the agents are squares, not circles.

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- This change requires a documentation update

# Checklist:

- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `pre-commit run --all-files` (see `CONTRIBUTING.md` instructions to set it up)
- [ ] I have run `pytest -v` and no errors are present.
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I solved any possible warnings that `pytest -v` has generated that are related to my code to the best of my knowledge.
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

Issue #1385, opened 2026-07-06
Better Documentation Gifs for Multiwalker, Cooperative Pong and Pursuit


jkterry1:
The rendering for the environments in classic is planned to be overhauled, the gif of Pistonball is already good, and PR just went up to give KAZ (the most difficult environment to deal with in some regards) a gif showing a good policy.

Ignoring the multi-agent atari games, which are kind of their own thing, this leaves needing documentation gifs showing good learned policies for Multiwalker, Cooperative Pong (it has a policy it's just not very good), and Pursuit (once https://github.com/Farama-Foundation/PettingZoo/issues/1360 is resolved first since the rendered appearance will change).

jkterry1:
Learning the cooperative pong policy _may_ want to also wait on https://github.com/Farama-Foundation/PettingZoo/pull/1336 first

PR #1456, opened 2026-09-08
multiwalker: replace the docs gif with a trained walking policy

tzh476:
Covers the Multiwalker half of #1385, following #1446's pattern for Pursuit: the environment's docs gif replaced in place.

PPO on the **default reward**, 6M timesteps, ~20 min on CPU. The policy is an MLP over the 31-dimensional observation, so no GPU is involved.

**Displacement is the metric, not reward.** With `terminate_reward = -100` against `forward_reward = 1.0`, a policy that stands still and never drops the package already scores positively — a shorter run of this same setup reached **+6.30 reward with 20/20 full-length episodes while the package moved 0.95 units and then stopped dead**. I nearly shipped that. Reward cannot separate those two behaviours, so both scripts report displacement first.

Two disjoint seed blocks, 25 episodes each, `max_cycles=500`:

| arm | seeds | displacement | reward | steps |
| --- | --- | --- | --- | --- |
| PPO | 0–24 | **4.97 ± 1.74** | 8.27 | 492.1 |
| PPO | 1000–1024 | **4.75 ± 1.51** | 7.50 | 492.8 |
| do nothing | 0–24 | 1.98 ± 0.21 | −96.84 | 111.0 |
| do nothing | 1000–1024 | 1.81 ± 0.36 | −97.39 | 112.9 |
| random | 0–24 | 0.13 ± 0.57 | −100.76 | 83.5 |
| random | 1000–1024 | 0.29 ± 0.51 | −104.38 | 72.5 |

The do-nothing arm still shows ~1.9 displacement because the walkers topple forward before terminating, so **the honest margin over it is about 2.5×**, not the 12× the reward column suggests.

The gif is seed 1, displacement 6.02, and reproduces byte-for-byte from the documented command. It is 311 KB against the 323 KB it replaces. No checkpoint is committed — the training command regenerates it.

`ruff check` and `ruff format --check` pass on both added files.

I'm aware 4.97 is a modest fraction of the 200-unit terrain: the walkers stride and carry the package for the full episode but do not cross the map. If you want a stronger policy before the gif changes, say so and I'll train longer rather than have this land as-is.


tzh476:
Correcting an overstatement in my own description, before anyone spends review time on it.

I wrote that the gif "reproduces byte-for-byte from the documented command". That was true only on my machine, because a trained checkpoint was already sitting there. The training run was **unseeded**, so a reader following the documented command would have landed on a different policy and a different gif. That is not reproducible, and I should not have claimed it was.

Fixed in the latest commit: training now seeds torch and numpy directly (PPO's own `seed=` forwards to `env.seed()`, which SuperSuit's `ConcatVecEnv` does not implement and which raises `AttributeError`).

The honest scope of the claim is now:

- **Reproducible:** the evaluation numbers and the reported displacement, from a seeded run.
- **Not promised:** a matching file hash. PyTorch does not guarantee bit-identical results across versions or hardware, so a gif regenerated elsewhere will look equivalent rather than hash-equal.

The committed gif is still the one measured above — seed 1, displacement 6.02, verified again after this change.

